PR #33, opened 2017-09-08
Add RouteWrapper higher order component to add `fetch` and `to` attributes

brettjonesdev:
## Feature
In beacon embed, we had a problem where we wanted to fetch some data from the server before navigating to a new <Route>. To make this easier, this Pull Request creates a Higher Order Component, `RouteWrapper`, which wraps around components and adds `fetch` and `to` properties. The `to` property works like the `to` on a React Router `<Link>`, but the `fetch` property lets you specify a thunk-returning function to fetch data before triggering the route specified in `to`. 

The `<Link>` and `<Button>` components have been wrapped with `RouteWrapper`, which means we can now do things like:
```javascript
<Link fetch="fetchArticle" to={`docs/articles/${id}`}>
<Button fetch="fetchArticles" to="docs/articles">
<Button to="synchronous/route/path">
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/blue",
-  "version": "0.0.14",
+  "version": "0.0.15",
   "private": false,
   "main": "dist/index.js",
   "scripts": {
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -2,10 +2,23 @@
 
 Actionable HTML buttons with basic styling
 
-## Example
+## RouteWrapper
+
+This component is extended by the `RouteWrapper` higher order component, and can be used similarly to a [`react-router` `<Link>`](https://reacttraining.com/react-router/web/api/Link) component by using the `to` prop. It also has an optional `fetch` property which can specify a promise-returning function which will be invoked before the `to` route is navigated
+
+
+## Examples
 
 ```html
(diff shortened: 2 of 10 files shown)


brettjonesdev:
@Charca I'm 90% sure that the context comes from using a component inside a `<Router>`, which we will always be doing in Beacon. In `hs-app`, we may not use this HOC in the short term, since we're still tied to Backbone Router/history. 

I can add a warning about the missing router history context, is there a particular way to do that in React? Or just like a `console.warn()`?

ItsJonQ:
Hmm! Now that you bring this up @Charca … would this break if you tried to use it without `react-router`?

brettjonesdev:
@ItsJonQ You wouldn't be able to use the `to` property is all. Which is actually the same as it was before, since you originally created a `<RouterLink>` if the `to` field was defined.

brettjonesdev:
(You also wouldn't be able to use `fetch`, FWIW but I think that's probably fine)

ItsJonQ:
Oh true!

Charca:
@brettjonesdev Yes! You're right that the `<Route>` component defines the context, thanks for the reminder! And just a plain `console.error` would do I think :)

brettjonesdev:
@Charca done!

PR #48, opened 2017-09-18
Fix `initials` attribute being passed down to `<Avatar>` <div>

brettjonesdev:
Bug fix - `initials` was being passed down into the `Avatar`'s `div`.


brettjonesdev:
Merging up so I can build a new version and use it in beacon2 - @ItsJonQ please do take a look and update as you see fit!

PR #59, opened 2017-10-02
Add hint text to input component

knicklabs:
It may be desirable to have text appear above and/or below the input but currently it is only possible to configure the input to have text appear below it via the `helpText` prop.

This PR adds a `hintText` prop that can be used to configure the input to have text appear above it, as shown below.

Screenshot

### DIFF
--- a/src/components/Input/README.md
+++ b/src/components/Input/README.md
@@ -29,6 +29,7 @@ An Input component is an enhanced version of the default HTML `<input>`. Input c
 | className | string | Custom class names to be added to the component. |
 | disabled | boolean | Disable the input. |
 | helpText | string | Displays text underneath input. |
+| hintText | string | Displays text above input. |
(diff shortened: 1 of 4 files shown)


ItsJonQ:
@knicklabs Cool! My only thought is that the hintText appears where the label should be. Is there a chance we'd ever use both?

If not, then maybe we can do something like…

`label ? label : hintText ? hintText : null`

Which prioritizes label > hintText

What you thinky? :D

knicklabs:
@ItsJonQ It actually is below the label. The thought would be that they could be used together whereas the hint text may expand upon the label. See: http://c.hlp.sc/1L2Z0c3g1G2l

Here's a real world use case: http://c.hlp.sc/19461w0c1S0V

ItsJonQ:
Oh I see! That looks great :D! Merging

Thanks!!!! <3

PR #64, opened 2017-10-10
Prevent default on SidebarCollapsibleCard link click

lgariepy:
Currently, when I click on the toggle on the `SidebarCollapsibleCard` component, the browser navigates to "#". This is a quick fix to prevent that behavior.

I also added a new line to .gitignore, for those of us who are using PHPStorm (me 😄 )


ItsJonQ:
Thanks for the fix @lgariepy !! Will merge once Travis is green :)

PR #90, opened 2017-10-18
Added onClick handler back to the Card component

Charca:
After upgrading Blue to the latest version in Beacon, I found out that the `Card` component no longer had an `onClick` handler when they were given an `href` or `to` attribute. I've added it back to fix its behavior.


ItsJonQ:
@Charca Thanks for catching this! Gonna add a couple of tests 🙌

ItsJonQ:
Will merge as soon as Travis tests pass ❤️

PR #95, opened 2017-10-21
Add ScrollLock component, extend Scrollable

### LINKED ISSUE #56: Component: ScrollLock

alisdair:
Demo

Fixes #56

The ScrollLock component wraps any scrollable element, and prevents users from scrolling past its top or bottom extent. This is especially important for scrollable modals or dropdowns, as scrolling past the end will otherwise cause the page to scroll.

Implemented by cloning the only child element and catching the wheel event, which is cancelled if the scroll would move past the end of the container.

Since this is the expected user behaviour, also extend Scrollable to use ScrollLock by default. This can be disabled with a prop.


ItsJonQ:
@alisdair Amazing!! Thank you so much :D. As a bonus, it works beautifully in components like Modals that already `<Scrollable>`., which get scrolllock for free with this update 😊 

Super approved!

PR #120, opened 2017-11-22
Add ChatSidebar component

ItsJonQ:
## Add ChatSidebar component

screen recording 2017-11-22 at 01 29 pm

This (pretty big) update adds (and enhances) a variety of components to create a new ChatSidebar component.

The ChatSidebar is a larger component that helps manage the presentation, interactions, and state for the new ChatInbox and ChatList components. Ideally, state for ChatSidebar should be managed externally (via a parent component), and passed in as props.

### Example

```jsx
class MyComponent extends Component {
  ...
  render() {
    const {
      handleOnHideStatusBar,
      isShowStatusBar,
      newMessageCount,
    } = this.state

    return (
      <ChatSidebar
        newMessageCount={newMessageCount}
        onHideStatusBar={handleOnHideStatusBar}
        isShowStatusBar={isShowStatusBar}
      >
        <ChatInbox>
          ...
        </ChatInbox>
      </ChatSidebar>
    );
  }
}
```

An example of this (and all the new components) can be found in Blue's storybook:
http://localhost:8900/?selectedKind=ChatSidebar&selectedStory=default&full=0&down=1&left=1&panelRight=0

### List of new components

* ChatSidebar
* ChatInbox
  * ChatInbox
  * Header
  * Content
* ChatList
  * ChatList
  * BlankSlate
  * Item
* Tag
* Illo
* Overflow
* StatusBar

### List of existing components that were enhanced

* AnimateGroup
* Scrollable
* ScrollLock
* Text

@knicklabs Ready for you to review!!!!! 🤗 

There's a TON of changes, mostly with tests + documentation. The best course of action to review is to check out the Storybook examples for all the components I mentioned above. Need your 👍 for component implementation and state/prop management.

I wold focus on `ChatInbox`, `ChatList`, and `ChatSidebar`, but feel free to check out the others (like `Overflow`).

Resolves: https://github.com/helpscout/blue/issues/115


src/components/AnimateGroup/index.js
```diff
@@ -1 +1,66 @@
...
+
+      const key = child.props.id || index
+
+      if (
+        child.type === Animate ||
+        child.type === Transition ||
+        child.type === CSSTransition
+      ) {
+        return React.cloneElement(child, {
+          wait: (child.props.wait + ((index + 1) * staggerDelay)),
```
knicklabs: This is clever :)

src/components/ScrollLock/index.js
```diff
@@ -1,35 +1,82 @@
...
     if (!children) {
       return null
     }
 
     const child = React.Children.only(children)
     const events = {
-      onWheel: handleWheelEvent
+      onWheel: event => {
+        handleWheelEvent(event, direction, stopPropagation)
+        onWheel(event)
```
knicklabs: Do you think these calls should be throttled?

src/components/AnimateGroup/index.js
ItsJonQ: Thanks!

src/components/ScrollLock/index.js
ItsJonQ: 🤔 I think in the case of ScrollLock (and Overflow), we **can't** throttle the scroll event, as it would result in a scroll-jacking vibe. The calculations need to fire at every tick. But I totally know where your concern is coming from.

I've done extensive testing to check for re-draw performance and fps drops. So far, these components hold up very well, and fps stays at 60 with low CPU/GPU load.

I think we'll only need to consider throttling if we notice dropped frames or jankiness.

Hope that makes sense!!

ItsJonQ:
@knicklabs Thanks a bunch! Will merge this in :D

src/components/ScrollLock/index.js
knicklabs: Sounds good!

PR #163, opened 2017-12-13
Allow input state to be updated by passing new value prop

knicklabs:
## Input : Allow state to be updated by new `value` prop

When the input is initialized, the value property of its state is set to the passed value prop. The value prop on the internal state is then only updated via the `handleOnChange`. This is fine when the user is in control of the input. There may be times however, where we want to update the value on the internal state programatically, such as allowing a picker to inject content, as shown below. 

This PR updates the Input component such that the value on the internal state can be updated by passing a new value prop to the component.


knicklabs:
Thanks @ItsJonQ!

PR #166, opened 2017-12-13
Modal: Provide delay for CloseButton reposition

ItsJonQ:
## Modal: Provide delay for CloseButton reposition

This update adds a new prop to `Modal`, which provides a delayed time before
the CloseButton is repositiioned (if the CloseButton is enabled).

This is necessary for certain situations where content that's injected into
the `Modal.Body` changes size, from a no-scrollbar state to a scrollbar state.
An example would be media (like images) unfurling during the mount state,
which is hidden via the Animation transition.

This is very edge-case, but it's something that I saw during Beacon
integration.


ItsJonQ:
Need to fix timings for Travis…

PR #168, opened 2017-12-14
Utilities: Expose all of Blue's utils

ItsJonQ:
## Utilities: Expose all of Blue's utils

This update exposes all of Blue's utils, making them available to be
imported and used.

All of the utilities come bundled within a `utilities` object.

**Example**
```
import { smoothScroll } from '@helpscout/blue/utilities'

// Now you can do things with smoothScroll.smoothScrollTo()!
```

Alternatively, you can do this (which might not be recommended, as it imports all the utils):
```js
import { utilities } from '@helpscout/blue'

const {
  smoothScroll
} = utilities

// Now you can do things with smoothScroll.smoothScrollTo()!
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,6 +165,7 @@
       "src/**/*.{js,jsx}",
       "!src/tests/helpers/**/*.{js,jsx}",
       "!src/utilities/smoothScroll.{js,jsx}",
+      "!src/utilities/index.{js,jsx}",
       "!src/utilities/log.{js,jsx}",
       "!src/vendors/**/*.{js,jsx}",
       "!src/components/index.js",
--- a/src/index.js
+++ b/src/index.js
@@ -1,10 +1,2 @@
-import { classNames, variantClassNames } from './utilities/classNames'
-import { createUniqueIDFactory } from './utilities/id'
-
+export {default as utilities} from './utilities'
 export * from './components'
-
-export const utilities = {
-  classNames,
-  createUniqueIDFactory,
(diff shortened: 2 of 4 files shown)


brettjonesdev:
@knicklabs @Charca @lgariepy was just discussing with Q - I'm partial to `export * from './utilities'` and then `import { smoothScroll } from '@helpscout.blue'`. But Q's concern was that names are too similar to the component, with the only difference being UpperCase vs lowerCase. My concern was that this way ^^ I think forces one to import _all_ of the utilities even if you only want one. 

I can see both sides, wonder what you all think?

knicklabs:
I too prefer the `export * from` approach. I also don't think it's an issue with similar names as there is a definite convention between upper- and lower-case. I personally don't think that is confusing. 

As an alternative, what if we use `@helpscout/blue/utilities/index.js` as our entry point for those exports. That would give us a situation like this: `import {smoothScroll} from '@helpscout/blue/utilities'` and `import {SmoothScroll} from '@helpscout/blue'`, so that we don't have to differentiate between components utils based on case along, where those names are the same. We can always look to the import statements for clarity.

brettjonesdev:
I like the `import { smoothScroll } from '@helpscout/blue/utilities'` the most! Is that possible with the current implementation?

ItsJonQ:
@brettjonesdev Yup! That should work 👍 . Just did a little test, and it looks like it's importing correctly with that structure.

PR #181, opened 2018-01-25
Avatar Feature: Fallback to initials on image error

knicklabs:
This PR introduces a new Avatar feature whereby if the image fails to load, the Avatar will fallback to rendering the initials.

This feature makes use of the HTMLImageElement `onError` callback to detect errors and update the `imageLoaded` property on the component's state. This causes the component to re-render with the initials instead of the image.

Because we cannot programmatically detect errors on background images, we use a hidden image element to detect and act on errors.

### DIFF
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -30,6 +30,22 @@ const defaultProps = {
 }
 
 class Avatar extends Component {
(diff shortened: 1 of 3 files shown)


ItsJonQ:
Will merge as soon as Travis is 💚

PR #182, opened 2018-01-25
Apply transparent background when Avatar has image

knicklabs:
Currently, the Avatar component has a background color even when an image is used instead of initials. This results a flash of color before the image loads that is particularly pronounced when the image is not cached in the browser.

This PR introduces a change to use a transparent background when an image is used. The effect of this change is that the space for the avatar is rendered when waiting for an image to load with a transparent background, making for a more seamless loading effect.

Note that in the render function we are building up a style object and that object is reassigned to null if no styles are added. This is to make it easier for testing to simply verify that styles are falsy.

### DIFF
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -63,10 +63,11 @@ class Avatar extends Component {
     } = this.props
 
     const { imageLoaded } = this.state
+    const hasImage = image && imageLoaded
 
     const componentClassName = classNames(
       'c-Avatar',
-      image && imageLoaded && 'has-image',
+      hasImage && 'has-image',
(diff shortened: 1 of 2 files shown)


src/components/Avatar/index.js
```diff
@@ -93,10 +94,16 @@ class Avatar extends Component {
         {text}
       </div>
 
-    const styles = borderColor ? {
+    let styles = borderColor ? {
       border: '2px solid',
       borderColor
-    } : null
+    } : {}
```
ItsJonQ: Ooh! Awesome fix! Keeping the return the same type.

PR #237, opened 2018-03-19
Make LineItems support dangerously-set HTML `body`s

brettjonesdev:
ChatTranscript `LineItem` bodies contain HTML. This PR makes the `LineItem` support dangerously-set HTML in the `body`.

### DIFF
--- a/src/components/ChatTranscript/LineItem.js
+++ b/src/components/ChatTranscript/LineItem.js
@@ -43,7 +43,7 @@ const LineItem = props => {
     </span>
   ) : null
 
-  const contentMarkup = body || children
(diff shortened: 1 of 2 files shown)


ItsJonQ:
@brettjonesdev Oo, nice!! Would you be able to write a quick test for this? With that, it should be 👍 from me!!

brettjonesdev:
@ItsJonQ done!

ItsJonQ:
@brettjonesdev Awesome! Thanks 🍡 .

I'm going to be submitting an update for the `<Inline>` component. Will make a new release with those updates plus this PR

PR #250, opened 2018-04-20
Guard against invalid values in the nameToInitials strings util

tompedals:
We had some bad data turn up in prod, which was causing Agent Chat to throw a JS error and stop working. I haven't been able to fully track down the data, but the error was within the `nameToInitials` util. I could reproduce the error by passing a string with additional whitespace.

Needed for https://github.com/helpscout/hs-app/pull/4885

### DIFF
--- a/src/utilities/strings.js
+++ b/src/utilities/strings.js
@@ -3,7 +3,16 @@ export const isString = (string) => {
 }
 
 export const nameToInitials = (name = '') => {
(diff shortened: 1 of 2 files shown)


tompedals:
[Avatar component JS error causing chat to completely stop working](https://trello.com/c/GrCgMJpo/344-avatar-component-js-error-causing-chat-to-completely-stop-working)

PR #251, opened 2018-04-23
Timestamp: Live update

tompedals:
This adds support for live updating the timestamp. The existing behaviour of outputting the given `timestamp` prop is still valid, however we can now also specify a `formatter` and `live` prop to have the timestamp over the time period. The interval depends on the difference in time between the current timestamp and the given timestamp. This is to reduce the number of updates when we are formatting the date as a time/date string rather than `X minutes ago`.

I have attempted to keep test coverage at 100% but struggled to test the lifecycle events properly. I still need to figure out how to get the `componentDidUpdate` lifecycle function to fire in the test.

For Chat we will be utilizing `formatDate` from https://github.com/helpscout/js-utils for timestamp formatter.

This is needed for https://trello.com/c/BIcQTV9V/247-chat-initiated-time-on-preview-cards-doesnt-update-in-realtime. I have pulled in these changes and tested within hs-app. Once we're happy with this and it's in a Blue release I'll create the hs-app PR also.


tompedals:
[Chat initiated time on preview cards doesn't update in realtime](https://trello.com/c/BIcQTV9V/247-chat-initiated-time-on-preview-cards-doesnt-update-in-realtime)

ItsJonQ:
@tompedals Awesome 😻 ! Reviewing now

ItsJonQ:
@tompedals Neato features!!! The formatter is really cool, but the auto self-updating timestamp is reeeeeally cool!

I looked over the code, and I kinda get it at a high level, but I'm not 100% sure I know how the timing mechanism works. It usually clicks when I can play around with it in Storybook.

I've pushed an update that sets up the Storybook for Timestamp (there wasn't one before).
You'll be able to access it in this URL:
http://localhost:8900/?selectedKind=Timestamp&selectedStory=default&full=0&down=1&left=1&panelRight=0

(After running `npm start`)

The file is located at:
```
stories/Timestamp.js
```

I gave it a shot, but I didn't see anything happen :O.
If you could add working examples, that would be awesome!

It also looks like there's some functionality not yet tested in your PR:

screen shot 2018-04-23 at 11 41 56 am

You can check this locally by running `npm run test`, which runs a full test + generates a coverage report, which you can access at:

```
/coverage/lcov-report/index.html
```

Or specifically:

```
/coverage/lcov-report/components/Timestamp/index.html
```

You can open this in your browser locally. My URL is:
file:///Users/Q/Sites/helpscout/blue/coverage/lcov-report/index.html

Lastly, would you be able to update the `README.md` file in the `Timestamp` component directory to describe the new props you added?

Thanks a bunch!!

tompedals:
@ItsJonQ I've updated the story to have a live example and also added more test cases to cover the lifecycle events.

The way the timing mechanism works is that we want to force an update to render once every so often. The less the difference between the timestamp and now the more frequent we update.

* < 1 minute: update every second
* < 1 hour: update every minute
* < 1 day: update every hour

There are some assumptions being made that the formatter would become less specific the more time passes. This is what happens with our formatter in `js-utils`.

timestamp-story

ItsJonQ:
@knicklabs I'm gonna be merging this one to help speed things along! Hope you don't mind 😘

PR #266, opened 2018-05-31
Add `SleepDetector` component to track browser/computer sleep

brettjonesdev:
## Feature
This PR adds a new `<SleepDetector onWake={cb}/>` component to help us detect when the browser or computer goes to sleep. This will allow us to do things like re-fetch the latest data when the user re-opens their laptop or unsuspends a tab.


ItsJonQ:
Made the change. Did it so that I can get this merged up with my other Blue update and make a release 🙌

Charca:
Nice job with this component, looks great! 👏

PR #277, opened 2018-06-07
Split Time component from Timestamp to allow just rendering time without wrapper

tompedals:
I needed a live updating `time` element that wasn't wrapped in the usual way that `Timestamp` does. It needed to be inline and inherit the styles from the container. To achieve this I split out the logic for the live updating timestamp into a `Time` component that could be used independently of `Timestamp` as `Timestamp.Time`.

The `calculateTimeoutPeriod` function is used to determine the interval to use for updating the live timestamp. When the difference is more than one minute we only updated every minute. This was causing some display issues when we had multiple components that were rendered at different times. One might say "3 min ago" and the other "4 min ago" as the intervals are out of sync. To alleviate some of this I updated the interval to be every 15 seconds rather than every 1 minute.

### DIFF
--- a/src/components/Timestamp/Time.js
+++ b/src/components/Timestamp/Time.js
@@ -1,0 +1,90 @@
+import React, { Component } from 'react'
+import { calculateTimeoutPeriod } from '../../utilities/timestamp'
+import PropTypes from 'prop-types'
+
+export const propTypes = {
+  formatter: PropTypes.func,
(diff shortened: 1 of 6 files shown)


ItsJonQ:
@tompedals Ah!!! Gotcha. Niceeeee! Micro functionality and composition :D. 

Very cool!! 🚀

PR #281, opened 2018-06-11
Transform newlines to break elements in chat content

tompedals:


tompedals:
[Chat transcript: Linebreaks are ignored](https://trello.com/c/HBqD6W7A/307-chat-transcript-linebreaks-are-ignored)

PR #291, opened 2018-06-28
Illo: Search: Fix color rendering

ItsJonQ:
## Illo: Search: Fix color rendering

This update resolves the rendering issue with the Illo:Search component.


ItsJonQ:
(Intentionally killed the test)

PR #297, opened 2018-07-16
Is typing events for inputs

tjbo:
This PR adds some callbacks to inputs for when the user starts or stops typing in an input.


tjbo:
Lodash's debounce and throttle have some a lead and trail call, but when both set to true and there is only one invocation of the debounce or throttle, it doesn't fire the trailing call. I though this would be better done in blue anyways; as then we have less places to update if we want to change.

tjbo:
pfffffttttt.... just fyi, it's common for me to push `test.only`... I do it like 30% of the time.

PR #318, opened 2018-07-30
Fix documentation for input

tjbo:
This PR makes small changes to `inputs` README.md.


tjbo:
@ItsJonQ , just a README.md change, not at all pressing and won't affect anything; other than the documentation.

PR #321, opened 2018-07-31
Animate: CSS - Refactor with Fancy

ItsJonQ:
## Animate: CSS - Refactor with Fancy

screen recording 2018-07-31 at 02 22 pm

This update refactors the `Animate` component's CSS with Fancy.
This is especially important for Portal'ed components like Tooltips and
Modals to animate correctly, as the previous SCSS styles may / may not be
available/loaded correctly in the Portal'ed environments.

Bonus: `Animate` has also been refactored with Flow


ItsJonQ:
Gonna merge this one up!

PR #324, opened 2018-08-01
PropProvider: A configurable Blue. Unlocked!

ItsJonQ:
## PropProvider: A configurable Blue. Unlocked!

This update adds a brand new component, the `PropProvider`. Powered by
React's new `context` API (polyfilled for React 15), it enables users to
specify configs to customize Blue components within an app.

The first component to use this is `Tooltip`, starting with just the `zIndex`
prop. I've wanted to implement this system for a while. The `zIndex`
update was the perfect excuse, as it is necessary for `Tooltip` components
to render correctly within nested `Modal`. (An edge-case... but one I experienced
in Beacon)

The `zIndex` can be customized at the App, or even component scope level
as required.

## Example

```jsx
import React from 'react'
import PropProvider from '@helpscout/blue/components/PropProvider'
import Tooltip from '@helpscout/blue/components/Tooltip'

const customConfigs = {
  Tooltip: {
    zIndex: 123,
  },
}

const App = () => {
  return (
    <Provider value={customConfig}>
      ...
      <Tooltip>...</Tooltip>
      ...
    </Provider>
  )
}
```

In the example (above), the `Tooltip` will use the `zIndex` config as a prop.

## How it works

Under the hood, components are rigged up, or "connected" with a special HOC
called `configConnect`, which works very similarly to the one found in Redux.

`configConnect` attempts to find the appropriate configs (`Object`) from the
`PropProvider`. If it's available, the configs will be passed to the component
(as a `...spread`).

---

Edit: Changed name from `ConfigProvider` to `PropProvider`, as it better describes what it does and follows the React/Prop naming patterns.


src/components/Tooltip/index.js
```diff
@@ -57,7 +62,7 @@ class Tooltip extends Component<Props> {
 
     if (!this.shouldRenderPopper()) {
       return children ? (
-        <span className={componentClassName} {...rest}>
+        <span className={componentClassName} {...getValidProps(rest)}>
```
ItsJonQ: Filters out non valid (default) React props 🎉 

https://github.com/helpscout/react-utils/blob/master/docs/getValidProps.md

knicklabs:
Blue is really maturing nicely. This is a great addition @ItsJonQ!

ItsJonQ:
Thanks all!! Merging :D

PR #326, opened 2018-08-03
Input: onResize callback + Enhance propConnect/PropProvider

ItsJonQ:
## Input: onResize callback + Enhance `propConnect`/`PropProvider`

screen recording 2018-08-03 at 01 39 pm

This update adds `onResize` to the `Input` component, which fires whenever
the `multiline` textarea resizes.

`Input` has been "connected" with the new `PropProvider`.

`PropProvider` was also refactored and improved. It follows HOC conventions
established by libraries like `redux`, such as component wrapping and
hoisting statics.

`configConnect` was also renamed to `propConnect`, to better reflect
`PropProvider`. The version of `react-utils` was bumped,
and used dependencies from that library instead (such as `createContext` and
`hoistNonReactStatics`.


ItsJonQ:
@Charca Thanks for 🍐 on `propConnect`! I've enhanced it to be able to grab multiple props.

Here are the updated docs:


## Connecting multiple props

You can connect your component to **multiple** config props using a variety of methods.

### Array

```js
const ConnectedDerekComponent = propConnect([
  'BlueSteel',
  'Karate',
  'Modelling',
  'Magnum'
])(DerekComponent)
```

### Function: `remapConfigToProps()`

This works similarly to Redux's [`mapStateToProps`](https://github.com/reduxjs/react-redux/blob/master/docs/api.md)


```js
const remapConfigToProps = config => {
  const = {
    BlueSteel,
    Karate,
    Modelling,
    Magnum
  } = config

  return {
    BlueSteel,
    Karate,
    Modelling,
    Magnum
  }
}

const ConnectedDerekComponent = propConnect(remapConfigToProps)(DerekComponent)
```


### Object

If an `Object` is provided, `propConnect` will evaluate keys with **truthy** values.

```js
const remapConfigToProps = {
  BlueSteel: true,
  Hansel: false,
  Karate: true,
  Modelling: true,
  Magnum: true
}

const ConnectedDerekComponent = propConnect(remapConfigToProps)(DerekComponent)
```

ItsJonQ:
Gonna merge this sucker!

CC'ing @tjbo and @knicklabs so you guys are aware of the new `PropProvider` features (above)

🤗

PR #327, opened 2018-08-08
Message: Timestamp interaction refactor

ItsJonQ:
## Message: Timestamp interaction refactor

screen recording 2018-08-08 at 12 41 pm

This update refactors how a `Timestamp` is rendered into a `Message`,
or a `Timeline` component.

It now always renders into the DOM, but is hidden with CSS `opacity`.
Previously, it would be mounted/unmounted and revealed with `HoverWrapper`
combined with `Animate`. This worked, but it may have resulted in some
scroll jittering as the layout re-adjusts.

Opting for this new CSS reveal method, means that I was able to remove
`HoverWrapper`.

Other enhancements include adjusting how the `Meta` markup is structured
within `ChatBlock`. It now no longer affects the vertical positioning of
a Chat `Timestamp`.

Lastly, `Timeline` was refactored with Flow, and several components were
"prop connected".


ItsJonQ:
Gonna merge this one up!!

PR #328, opened 2018-08-08
Fancy: Hoist statics for styled components

ItsJonQ:
This update bumps Fancy to the latest v2.0.5, which hoists non React
statics to the wrapped Styled component.

This is expected behaviour for HOC wrappers, but is missing from (stock)
Emotion.

Certain Components were refactored to better leverage the static hoists.

Reference: React static hoisting:
https://github.com/mridgway/hoist-non-react-statics

It's a thing that libraries like Redux use.

---

Fancy hoist tests:

https://github.com/helpscout/fancy/blob/master/src/styled/__tests__/styled.test.js#L110


src/components/Link/index.js
```diff
@@ -1,6 +1,5 @@
 // @flow
 import React, { PureComponent as Component } from 'react'
-import PropTypes from 'prop-types'
```
ItsJonQ: No longer used! Yay

ItsJonQ:
Merging this!

PR #329, opened 2018-08-09
IE: Improve rendering stability

ItsJonQ:
## IE: Improve rendering stability

This update adjusts the (Flex-based) styles of an inner component
of `Message.Bubble` to improve responsive rendering for IE 11.

The level of nested Flexbox-based components led to unpredictable
width calculations.

Other updates include adding a guard for any node element using
`scrollTo`. IE11's console produced an error for this.

---

CC'ing @Charca , @tjbo , @knicklabs


ItsJonQ:
Merging!

PR #330, opened 2018-08-10
Message: Support Embedded HTML Content

knicklabs:
## Message: Support Embedded HTML Content

This update adds a new `Message.Embed` component to Blue. It accepts an `html` prop and renders the HTML inside the component with style treatments similar to `Message.Media`. Unlike `Message.Media`, it does not open the content in a modal.

Screenshot

### DIFF
--- a/src/components/Message/styles/Embed.css.js
+++ b/src/components/Message/styles/Embed.css.js
@@ -1,0 +1,45 @@
+// @flow
+import baseStyles from '../../../styles/resets/baseStyles.css.js'
+import { BEM } from '../../../utilities/classNames'
+import { getColor } from '../../../styles/utilities/color.js'
+import { noteBoxShadow } from '../../../styles/mixins/noteStyles.css'
+
+const config = {
+  embedBorderRadius: '3px',
+  embedWidth: '9999px',
+  embedMaxWidth: '300px',
+}
+
+const bem = BEM('.c-MessageEmbed')
+
+export const css = `
+  ${baseStyles}
+
+  ${bem.element('html')} {
+    border-radius: ${config.embedBorderRadius};
+    max-width: ${config.embedMaxWidth};
+    overflow: hidden;
+    width: ${config.embedWidth};
+  }
+
+  ${bem.element('bubble')}.c-MessageBubble {
+    background-color: white;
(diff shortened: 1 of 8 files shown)


src/components/Message/styles/Embed.css.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import baseStyles from '../../../styles/resets/baseStyles.css.js'
+import { BEM } from '../../../utilities/classNames'
+import { getColor } from '../../../styles/utilities/color.js'
+import { noteBoxShadow } from '../../../styles/mixins/noteStyles.css'
+
+const config = {
+  embedBorderRadius: '3px',
+  embedWidth: '9999px',
+  embedMaxWidth: '300px',
```
ItsJonQ: I noticed a responsive issue for smaller viewports (mobile).
If you adjust to `100%`, then it (and the contents within) should be responsive:

screen recording 2018-08-10 at 05 40 pm

src/components/Message/styles/Embed.css.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import baseStyles from '../../../styles/resets/baseStyles.css.js'
+import { BEM } from '../../../utilities/classNames'
+import { getColor } from '../../../styles/utilities/color.js'
+import { noteBoxShadow } from '../../../styles/mixins/noteStyles.css'
```
ItsJonQ: Flow complains for this import path. Needs `.css.js`.
It still works.. just Flow yelling.

For some context, I previously didn't use `Flow` for the `.css.js`. I figured... these shouldn't have too much going on.

I was wrong.

I forgot how complex CSS could be (it's been a while since I wrote Seed-like functional Sass).

So there's inconsistencies between the `css.js` files.

I'm gonna do a cleanup to add Flow and fix import paths (in another PR)

ItsJonQ:
screen shot 2018-08-10 at 5 49 22 pm

Leaving a note here as well! This is the only thing not covered :).
Should be easy peasy to write a quick test to check for the theme className.

After this, and the 2 things I pointed out (`maxWidth` and `.css.js` import), this is 🚀 !!

PR #334, opened 2018-08-16
ButtonV2!

ItsJonQ:
## ButtonV2!

button states

This update completes the initial work for ButtonV2. V1 is still the
default version for the `Button` component. The component API is
slightly different compared to ButtonV1. Please refer to the docs when
migrating from one version to the other.

In additional to updated styles, ButtonV2 also supports `<Icon />` usage
within, and applies the appropriate margin offset for the `Icon` + text.

Most of the work in this update is within the `Button.css.js` file.

Buttons are very complicated components. And the CSS styles reflect this, as
Buttons can have a multiple of various combinations and states.

I tried to keep things as organized as possible.


ItsJonQ:
Thanks all!! Gonna merge this up :)

Charca:
Nice work, Q! What are your thoughts on replacing v1 with v2? Do you know where are we using the v1 Button?

ItsJonQ:
@Charca We use ButtonV1 in Embed. I'm nervous about replacing it there, because we wrote a bunch of stuff to support "theming" for Embed buttons.

It's safer in Builder/HS-App, as there were minimal adjustments made on the base `Button` component

Charca:
Got it, thanks! If embed is the only project using v1, I think it makes sense to move the core logic of the v1 Button there and only support v2 in Blue, right? It'd be nice to have just one button to mantain in Blue. Something we can CANI eventually!

ItsJonQ:
@Charca Soudns like a plan!!! I'd be down moving ButtonV1 out of Blue once we can confirm that ButtonV2 works 💯 in HS-App :D

PR #337, opened 2018-08-24
Remove tabindex on close button for modals

tjbo:
This PR removes the `tabindex` attribute from modal close buttons. This fixes some 'focus tab' issues where the user can't tab to the close button.

more info: https://trello.com/c/CsX0PDSX/552-accessibility-issues-tabbing-escing-highlighting-within-embed

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -227,7 +227,7 @@ class Modal extends Component {
           this.closeNode = node
         }}
       >
-        <CloseButton onClick={closePortal} tabIndex="0" />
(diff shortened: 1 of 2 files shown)


tjbo:
more info here:
https://trello.com/c/CsX0PDSX/552-accessibility-issues-tabbing-escing-highlighting-within-embed

ItsJonQ:
@Charca The close button was intentionally set to not be tabbable. The idea behind that was that the user is able to close the modal with another keyboard command (`ESC`).

I'm good with this change. We can always revert if we want to modify the behaviour 👍

PR #338, opened 2018-09-01
Do not enlarge images

knicklabs:
**[Trello Card](https://trello.com/c/F42QuDSu)**

When the Image component is provided with a maxWidth and maxHeight, we would calculate an aspect ratio for the image that would fit within those bounds. This would create values that we could multiple the width and height by to get new widths and heights. The problem was in doing this when the width and height were less than the maxWidth and maxHeight. In such a case we would multiple the width and height resulting in larger numbers. In these cases, we should not apply a multiplier, setting the width and height to their original values.


ItsJonQ:
@knicklabs Amazing!! Thank you for this 🙏

PR #344, opened 2018-09-10
Add styles for disabled state on the Switch component

tompedals:
## Switch: Add styles for disabled state

Within Beacon Builder we have a number of switches that are sometimes in a disabled state. This functions as expected, however it's not visually clear that the switch is disabled. We have had a number of false bug reports due to this.

This change adds some styling for the disabled state. I have attempted to make the switches slightly opaque when disabled to differentiate to the enabled switches. It doesn't quite look right to me, but code wise it's working.

screen shot 2018-09-10 at 18 18 45

@digitaldesigner Do you have any input on this? I couldn't see this state in the designs https://github.com/helpscout/library-hsapp/wiki/Example:-Settings-Refresh

### DIFF
--- a/src/components/Switch/Switch.js
+++ b/src/components/Switch/Switch.js
@@ -192,6 +192,7 @@ class Switch extends Component<Props, State> {
 
     const componentClassName = classNames(
       'c-Switch',
+      disabled && 'is-disabled',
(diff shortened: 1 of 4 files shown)


ItsJonQ:
@tompedals Amaze! 😍 Thank youuuuu!

PR #346, opened 2018-09-12
Fix Codesandbox Integration

ItsJonQ:
## Fix Codesandbox Integration

screen shot 2018-09-12 at 5 01 13 pm

https://codesandbox.io/s/ww2kyjp5vw

Originally brought on my radar by @knicklabs and @ShayCichocki (❤️)

I got it working with Codesandbox. Stackblitz is.. another thing. Not sure how to resolve it on there. The issue is that Stackblitz is trying to compile/parse SCSS->CSS things. I don't know how to make it NOT do that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/blue",
-  "version": "1.2.8",
+  "version": "1.2.9-0",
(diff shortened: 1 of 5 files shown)


knicklabs:
❤️

PR #352, opened 2018-09-19
Accordion Component

knicklabs:
## Accordion Component
This update creates an accessible accordion component. By default it enforces only one open section at a time. It can be configured to allow multiple open sections at a time.

Here is a demo of the default display and operation:

Demo 1

Here is a demo of the default display with the allow multiple option enabled. 

Demo 2

Here is a demo of the seamless display. Boring yes. But also more easily customizable.

Demo 3

And finally, here is a demo of the accordion magically enhanced inside a page component:

Demo 4


ItsJonQ:
🎉 This is an amazing start!! @knicklabs Thank you so much for making these 🙏

ItsJonQ:
@knicklabs and I 🍐 on this sucker!! Just pushed my refactors + enhancements.

@knicklabs back to you for testing + wrap up!!

PR #359, opened 2018-09-25
PropProvider: Enhance to extends prop configs

ItsJonQ:
## PropProvider: Enhance to extends prop configs

This update improves `PropProvider` to allow nested instances to **extend**
configs. This makes usage of PropProvider more versitile. Allowing it to be
used, not just at the root App level, but in views, sub-views, and component
levels as well.

This Provider component is constructed using patterns adapted from libraries
like Emotion and Styled Components. The have an internal Pub/Sub system that
allows them nested Providers to communicate to pass along configs internally.

The propConnect() component then subscribes to the closest PropProvider,
where the extended props are then merged and rendered.

An example of this can be found here:
https://github.com/emotion-js/emotion/blob/master/packages/emotion-theming/src/theme-provider.js

### Example

It is possible to nest `PropProvider` components. `PropProvider` will automatically **extend** config defined in any parent instance.

```jsx
<PropProvider value={{ Button: { kind: 'secondary' } }}>
  ...
  <PropProvider value={{ Button: { block: true }, Tooltip: { zIndex: 123 } }}>
    ...
    <Tooltip>...</Tooltip>
    <Button>...</Button>
    ...
  </PropProvider>
  ...
</PropProvider>
```

In the above example, at the level of `Button` and `Tooltip`, the `PropProvider` config will be:

```js
{
  Button: {
    block: true,
    kind: 'secondary'
  },
  Tooltip: {
    zIndex: 123
  }
}
```

Redux implements a similar, but more complicated, pattern as well for it's `connect`/`Provider`.


ItsJonQ:
OoooOoO boy! I came up with this solution last night:
https://codesandbox.io/s/k5oj5rj9nv

It leverages default `React.createContext` in a.. recursive, self-aware way. 
It accomplishes the same thing, but with way less code :D.

(Note: Used functional components in codesandbox cause FASTER to prototype)

PR #367, opened 2018-10-09
Only call highlightBlock if the ref element is not null

tompedals:
When unmounting the Highlight component the `ref` was being set to null, but this was not guarded against so it was attempting to highlight the `null` element. This change prevents the error by checking element is not `null` before calling `highlightBlock`.


ItsJonQ:
@tompedals Great fix! I made a tiny refactor to use the `methodName = () =>` convention, which is babel transformed in Blue. This ensures proper binding of `this` (the class instance) by default 👍

ItsJonQ:
After trying to implement Blue v1.7.0 into the app. I gotta say, this fix is really needed!
Currently unable to switch views because Highlight.js is failing to do it's thing on an unmounted `div`.

PR #383, opened 2018-10-17
Choice: Add a stacked option

tompedals:
## Choice: Add a stacked option

Required for https://trello.com/c/mmLEsjjF/64-1-add-mode-field-to-the-new-beacon-page

Adds a new `stacked` option to the `Choice` component which stacks the input above the label.

screen shot 2018-10-17 at 10 49 10

The desired result from the designs:

screen shot 2018-10-17 at 10 50 28

The horizontal positioning of the choices will be done within HS app as the layout of the choices will depend on page width.


ItsJonQ:
@tompedals Thanks!!! This looks great! Upon inspection.. I realized there may be some things that I'll need to adjust for `ChoiceGroup` and `Choice` . Specifically, to leverage `context` instead of relying on React.Children.map clone (which is kinda brittle).

I realized this when trying to setup the UI to look like what we need to for Beacon Builder.

I was able to get this to work:

screen shot 2018-10-18 at 1 02 51 pm

Updating/writing tests now

PR #391, opened 2018-10-23
AvatarStack + Avatar - Version 2!

ItsJonQ:
## AvatarStack + Avatar - Version 2!

screen recording 2018-10-23 at 05 02 pm

This update adds a new version of `AvatarStack` to Blue - `AvatarStackV2`. This component can be enabled by adding the `version={2}` to **any** `AvatarStack`. The component was designed with backwards compatibility.

### `propConnect` all the things!

Quite a bit of refactor work was done with Blue's internals, specifically the `Animate` and `AnimateGroup` component. The previous `React.cloneElement` based pattern has been replaced with `PropProvider` to pass props down. This is a much better pattern as it doesn't force the target child to be a direct child. There can be many components in between because the passing of props operates under React `context`.

This pattern was also used for `AvatarStack` passing props down to `Avatar`.

### Avatar enhancements

The `Avatar` component has been improved to better handle the image load transition. Previously, it was jump between a transparent circle -> image. Now, it shows the specified brand color as the background, and the image is transitioned in (when ready).

### Layer stacking / Auto sizing

screen shot 2018-10-23 at 5 10 26 pm

In addition to some minor UI adjustments for `AvatarV2`, `AvatarStackV2` presents the Avatars slightly differently. The Avatars are layered differently (as seen in the screenshot), and resize based on the number of avatars within the collection.


ItsJonQ:
@Charca Not that I know of. The component API and things should remain the same. The difference is how `AnimateGroup` communicates with `Animate`. Previously, it relied on `React.cloneElement`. It now uses `context`.

ItsJonQ:
Awesome! Thanks for the review all! Will get this merged up :)

PR #400, opened 2018-11-12
Dropdown V2!

ItsJonQ:
## Dropdown V2

screen recording 2018-11-16 at 03 50 pm

**[Video overview](https://soapbox.wistia.com/videos/5QrPA33LvU)**

Mega feature!~ This update adds a brand new component.. `DropdownV2`.

It's better in every way compared to the original implementation of Dropdown. However, it is **not** backwards compatible.

It must be imported like so:

```jsx
import Dropdown from '@helpscout/hsds-react/components/Dropdown/DropdownV2'
```

## Example

```jsx
<Dropdown items={[...]} />
```

### Custom Trigger

To customize the dropdown trigger, pass any component into the `renderTrigger` prop.

```jsx
<Dropdown items={[...]} renderTrigger={<div>My Trigger!</div>} />
```

### Customizing the Menu/Items

The `children` prop renders with an `Object` containing an **enhanced** listing of the original `item`. These enhanced items have special click handlers and other props that allow for the dropdowns to work.

```jsx
<Dropdown items={...}>
  {({ items }) => (
    <Dropdown.Menu>
      <CustomComponent />
      {items.map(item => <MyCustomItem {...item} />)}
    </Dropdown.Menu>
  )}
</Dropdown>
```

### Accessibility

screen recording 2018-11-16 at 01 38 pm

DropdownV2 has accessibility in mind, and has `aria` propped all the things!

🙌 

Storybook preview:
https://deploy-preview-400--hsds-react.netlify.com/?selectedKind=DropdownV2&selectedStory=Menu%2FDefault&full=0&addons=1&stories=1&panelRight=0


ItsJonQ:
I know coveralls!! I'm working on it! Calm down!!

😂

ItsJonQ:
@tinkertrain Tagging you just for 👀 and awareness! 🤗

src/components/Dropdown/V2/Dropdown.Container.tsx
```diff
@@ -0,0 +1,96 @@
...
+    ...initialState,
+    onBlur: noop,
+    onOpen: noop,
+    onClose: noop,
+    onFocus: noop,
+    isOpen: false,
+    innerRef: noop,
+    onSelect: noop,
+    subscribe: noop,
+    trigger: 'Dropdown',
```
knicklabs: I wonder if `Dropdown` is the most useful default Trigger value, or if it should be `Open` or `Close` depending on the state.

src/components/Dropdown/V2/Dropdown.Item.tsx
```diff
@@ -0,0 +1,225 @@
...
+  index: string
+  innerRef: (node: HTMLElement) => void
+  isHover: boolean
+  items: Array<any>
+  onMouseEnter: (event: Event) => void
+  onMouseMove: (event: Event) => void
+  onBlur: (event: Event) => void
+  onClick: (event: Event, props: any) => void
+  onFocus: (event: Event) => void
+  renderItem?: (props: any) => void
```
knicklabs: I know having strong types is good, but I think noop functions are one case where typed arguments are not important, what about using something like `(...args: any[]) => void` and creating a `noop` util to clean this up a bit, since now writing one each time is a bit more verbose?

src/components/Dropdown/V2/Dropdown.Container.tsx
ItsJonQ: Gave this some thought! I think `Dropdown` is the easiest. Showing labels like `Open`/`Close` would mean they would need to specify 2 props (for open/close).

Ideally.. you'd always change what the Trigger is 😅 . (Be it the label or markup)
Hope that makes sense!

src/components/Dropdown/V2/Dropdown.Item.tsx
ItsJonQ: @knicklabs TOTALLY did not know you can do that. Just gave it a shot, and it worked! (At least in VS Code) 😍

ItsJonQ:
@knicklabs Thank you!!!! Will push the type updates after running some tests.
Will merge this in and create a minor release!

src/components/Dropdown/V2/Dropdown.Container.tsx
knicklabs: Yeah, that's true. I would expect people to supply their own too. 👍 on your decision.

PR #405, opened 2018-11-26
Update Page component

tinkertrain:
## Page Component
- Updates styles to match latest designs
- Improves html semantics and accessibility:
  a. Page header using `<header>`
  b. Headings Using `<h1>` and `<h2>`
  c. Cards use `<section>`
- Moves to TypeScript
- Fixes import in `<Accordion>`

## <Page.Heading> Component
Pages generally have a main heading and multiple secondary ones, this component adds the ability to use <Page.Heading> with the correct styles for pages

`<Page.Header>` leverages this component in `primary` mode

Updated Page.Card.stories to see the changes

## Adds fsevents to optional dependencies

Webpack has an issue where on a mac, the CPU usage spikes to more than a 100%

Reference:
- https://github.com/storybooks/storybook/issues/752
- https://github.com/webpack/webpack/issues/701

Adding fsevents as a dependency fixes this issue, putting as optional means it won't break non-mac installations

## Tests

- Adds tests for Page.Heading
- Adds test to Page.Header making sure it renders an `<h1>`


package.json
```diff
@@ -186,5 +186,8 @@
       "stories/**/*.js"
     ],
     "cache": false
+  },
+  "optionalDependencies": {
+    "fsevents": "^2.0.1"
```
ItsJonQ: Oh! Was this due to Node 10? (Or something)
I experienced this randomly on one of my personal machines.

tinkertrain: TBH I'm not sure what the root cause is, I'm running 11, I had storybook running (and hsapp) and then the laptop's fans went crazy after a little while even when I wasn't doing anything, peaked at the activity monitor and both node processes were above 100%, after a little search, tried the `fsevents` fix and haven't have the issue again.

src/components/Page/utils.ts
```diff
@@ -0,0 +1,15 @@
+export const COMPONENT_KEY: {
+  Page: string
+  Actions: string
+  Card: string
+  Content: string
+  Header: string
+  Heading: string
+} = {
+  Page: 'Page',
+  Actions: 'Page.Actions',
```
ItsJonQ: Totally not your fault here... Moving forward, we can't use namespaces with `.` in them.
Thank Enzyme 3 for that 🤦‍♀️ .

Would you be able to change the names from `Page.Actions` -> `PageActions`?

Explanation 👇 

In Enzyme 2.x, doing finds for `Component.SubComponent` would find the components.
However, in Enzyme 3, doing a find for that query... Enzyme tries to find a `Component` with a **className** of `SubComponent`.

I discovered this in my upgrade all the things to React 16 PR:
https://github.com/helpscout/hsds-react/pull/403

Overtime, I'll have to go back to update the namespaces for other connected components.

stories/Page/Page.Card.stories.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
ItsJonQ: Not a huge deal.. but we probably should do this for `.js` files. It throws errors in console (cause it tries to import `PropTypes` incorrectly or something).

This is totally the thing to do for `.tsx` files though. Confusing I know :(

src/components/Page/utils.ts
```diff
@@ -0,0 +1,15 @@
+export const COMPONENT_KEY: {
+  Page: string
```
ItsJonQ: I love that you typed these 😍

src/components/Page/styles/Page.css.ts
```diff
@@ -1,8 +1,7 @@
-// @flow
 import baseStyles from '../../../styles/resets/baseStyles.css.js'
 import { breakpoint } from '../../../styles/mixins/breakpoints.css.js'
 import styled from '../../styled'
-import PageConfig from './Page.config.css.js'
+import PageConfig from './Page.config.css'
```
ItsJonQ: @tinkertrain I deliberately kept `.css.js` as an extension for these imports to imply that the imported code was coming from a CSS-in-JS based file... rather than a standard `.css` file.

Unsure if that's necessary... Wanted to get your thoughts on this!! <3

src/components/Page/styles/Card.css.ts
```diff
@@ -45,19 +44,19 @@ export const CardUI = styled('div')`
   }
 
   ${breakpoint(
-    PageConfig.breakpoint.widescreen,
```
ItsJonQ: Looks like some formatting went wonky here. No big deal at all! (For real, haha).

If prettier can't catch/autofix it, then it doesn't matter (IMO). Not worth debating these things 🤗

src/components/Page/styles/Header.css.ts
```diff
@@ -13,7 +12,7 @@ export const config = {
   },
 }
 
-export const HeaderUI = styled('div')`
+export const HeaderUI = styled('header')`
```
ItsJonQ: Awesome! Thank you 🙏

src/components/Page/styles/Page.css.ts
tinkertrain: I actually do like a bit better to leave it at .css.js, the issue is that I get an error building because they are .css.ts files and it can find the .css.js (makes sense) but also you can't make the import with the .ts extension, so our options are:

a. Using TypeScript on css-in-js files: `import PageConfig from './Page.config.css'`
b. Don't use TypeScript on css-in-js files: `import PageConfig from './Page.config.css.js'`

My hunch is that we would want to have all in TS, so option **a**?

src/components/Page/styles/Card.css.ts
tinkertrain: Ahhh, not sure why my editor didn't catch some of these, just ran prettier and it did find some other places and fixed them, I remember adding a few easy steps so prettier runs always on precommit, which was very handy in case people didn't have their editor to do it automatically on save :)

src/components/Page/utils.ts
tinkertrain: First time using TypeScript actually! 👀

src/components/Page/styles/Page.css.ts
ItsJonQ: Hmm!! Good point! Let's go with **a** 🙌

stories/Page/Page.Card.stories.js
tinkertrain: Ahh yeah, it was throwing errors on tsx files so I changed that one too, changed it back!

src/components/Page/utils.ts
tinkertrain: Makes sense! Silly that Enzyme should dictate this kind of thing  ¯\_(ツ)_/¯

PR #407, opened 2018-11-28
FormLabel: Adds a new mode (inline)

tinkertrain:
## Inline Form Labels

This PR proposes the addition of an "inline" mode on `FormLabel`, where the label (including help text) sits on the right and the content on the left. This is specially designed for when the content is a checkbox or a `<Switch>` as most recent designs lays them out.

The alignment of items is done using flexbox

This PR:

- Adds the prop `inline` to enable this mode
- Adds tests
- Adds stories

#### Default mode
```jsx
<FormLabel
  label="This is the label for the Input down there"
  helpText="Ignore this message, says nothing interesting."
>
  <Input />
</FormLabel>
```
default

#### Inline without help text present and with (one line, multi-line)
```jsx
<FormLabel
  label="Label for the Switch"
  inline
>
  <Switch />
</FormLabel>
```

```jsx
<FormLabel
  label="Label for the Switch"
  helpText="Enable this feature or you might regret it later."
  inline
>
  <Switch />
</FormLabel>
```

```jsx
<FormLabel
  label="Label for the Switch"
  helpText="Enable this feature or you might regret it later. Enable this feature or you might regret it later. Enable this feature or you might regret it later."
  inline
>
  <Switch />
</FormLabel>
```

inline labels

#### Example inside a page

inpage

### DIFF
--- a/src/components/FormLabel/FormLabel.tsx
+++ b/src/components/FormLabel/FormLabel.tsx
@@ -16,7 +16,7 @@ export interface Props {
   id?: string
   label?: any
   helpText?: any
-  inline?: boolean
+  isInline: boolean
 }
 
 export interface State {
@@ -31,6 +31,10 @@ class FormLabel extends React.Component<Props, State> {
     this.state = {
       id: props.for || props.id || uniqueID(),
     }
+  }
+
+  static defaultProps = {
+    isInline: false,
   }
 
   getContextProps = () => {
@@ -69,7 +73,7 @@ class FormLabel extends React.Component<Props, State> {
     const {
       className,
       children,
-      inline,
+      isInline,
       for: htmlFor,
       id: idProp,
       ...rest
@@ -77,7 +81,7 @@ class FormLabel extends React.Component<Props, State> {
 
     const componentClassName = classNames(
       'c-FormLabel',
-      inline && 'c-FormLabel--inline',
+      isInline && 'is-inline',
       className
     )
     const labelMarkup = this.getLabelMarkup()
@@ -89,7 +93,7 @@ class FormLabel extends React.Component<Props, State> {
           {...getValidProps(rest)}
           className={componentClassName}
           isHelpTextPresent={!!helpTextMarkup}
-          inline={inline}
+          isInline={isInline}
         >
           <div className="c-FormLabel__label">
             {labelMarkup}
--- a/src/components/FormLabel/FormLabel.tsx
+++ b/src/components/FormLabel/FormLabel.tsx
@@ -88,6 +88,7 @@ class FormLabel extends React.Component<Props, State> {
         <FormLabelUI
           {...getValidProps(rest)}
           className={componentClassName}
+          isHelpTextPresent={!!helpTextMarkup}
           inline={inline}
         >
           <div className="c-FormLabel__label">
--- a/src/components/FormLabel/FormLabel.tsx
+++ b/src/components/FormLabel/FormLabel.tsx
@@ -16,6 +16,7 @@ export interface Props {
   id?: string
   label?: any
   helpText?: any
+  inline?: boolean
 }
 
 export interface State {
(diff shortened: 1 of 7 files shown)


src/components/FormLabel/FormLabel.tsx
```diff
@@ -16,6 +16,7 @@ export interface Props {
   id?: string
   label?: any
   helpText?: any
+  inline?: boolean
```
ItsJonQ: I think we can probably make this always present, but have a `defaultProps: { inline: false }` :)

@tinkertrain What you thinky?

src/components/FormLabel/FormLabel.tsx
```diff
@@ -68,21 +69,33 @@ class FormLabel extends React.Component<Props, State> {
...
+      inline,
       for: htmlFor,
       id: idProp,
       ...rest
     } = this.props
 
-    const componentClassName = classNames('c-FormLabel', className)
+    const componentClassName = classNames(
+      'c-FormLabel',
+      inline && 'c-FormLabel--inline',
```
ItsJonQ: I've actually started to drop the BEM convention, and stuck with prefix based modifier names like `is-inline`.
May reasons is because it's simpler (less verbose). And since these classNames are controlled and scoped to the component, the chances if it conflicting its probably the same as the BEM version.

@tinkertrain What you think ? :D

src/components/FormLabel/__tests__/FormLabel.test.js
```diff
@@ -111,3 +112,50 @@ describe('Context', () => {
...
+    const wrapper = mount(
+      <FormLabel>
+        <Input />
+      </FormLabel>
+    )
+
+    expect(wrapper.find('.c-FormLabel--inline').length).toBe(0)
+  })
+
+  test('calculateContentRules generates the correct rules to align content in inline mode', () => {
```
ItsJonQ: Hah! Nice!! You added tests for the CSS-in-JS function 😍

src/components/FormLabel/styles/FormLabel.css.ts
```diff
@@ -3,7 +3,43 @@ import styled from '../../styled'
 
 export const FormLabelUI = styled('div')`
   ${baseStyles};
+  display: flex;
+  flex-flow: ${props => (props.inline ? 'row' : 'column')};
+  justify-content: space-between;
+  align-items: ${props => (props.inline ? 'flex-end' : 'stretch')};
```
ItsJonQ: @tinkertrain Although Emotion does support prop interpolation.. I actually prefer just using CSS className.modifiers for the rules, example:

```
flex-flow: column;

  &.is-inline {
    flex-flow: row;
  }
```

I personally find it easier to parse 😊 (and it's usually less verbose. Usually).

However, for trickier things that involve dynamic calculation, like your `calculateContentRules` implementation, it's okay.

What you think? :D

src/components/FormLabel/FormLabel.tsx
ItsJonQ: Update! Another convention we've used for a while now is to prefix boolean based props with `is`. So it would be `isInline`. I think it's a little clearer to the component consumer, as a boolean value is implied.

Unfortunately, since it's a newer(ish) convention, it's consistent through the codebase :(. Something I'm working on cleaning up overtime.

src/components/FormLabel/FormLabel.tsx
tinkertrain: I think is great, never been a big fan of BEM's super long class names. And I think think they are as useful in component and css-in-js world :)

src/components/FormLabel/FormLabel.tsx
tinkertrain: Makes sense to me! I like the convention

src/components/FormLabel/styles/FormLabel.css.ts
tinkertrain: I like it! In my first revision I didn't have the `is-inline` class added so I couldn't do that, but now I do so I will update this :)

PR #408, opened 2018-11-28
DropdownV2 Performance Boost + ComboBox

ItsJonQ:
## DropdownV2 Performance Boost + ComboBox 🚀 

Wowowow! Another big Dropdown (V2) update! This one brings a ton of performance enhancements to Dropdown (V2). Not only that, but it's now even flexible-lier.

We also get a brand new component called `ComboBox`, which is essentially a filterable Dropdown.

### ComboBox

screen recording 2018-11-28 at 06 02 pm

`ComboBox` is built on top of `DropdownV2`. It works very similar to other select/filterable Dropdown libraries like Chosen or React-Select.

With the help of @knicklabs , the component API is extremely flexible! Here's an example of how you can customize the filter/results.

### Customizing the filter

This component uses a very simple filter algorithm to match items. You can use your own custom method for finer-grain control.

#### `customFilter({ hasGroups, items, inputValue }, defaultFilter)`

| Prop          | Type         | Description                                   |
| ------------- | ------------ | --------------------------------------------- |
| hasGroups     | `boolean`    | Determines if the provided items are grouped. |
| items         | `Array<any>` | Collection of initial (unfiltered) items.     |
| inputValue    | `string`     | The search query from the `Input`.            |
| defaultFilter | `Function`   | The default filter algorithm.                 |

In the example below, we'll return an empty collection (`Array`) if there are no `inputValue`. Otherwise, we'll just use the default filter algorithm.

```jsx
const customFilter = ({ hasGroups, items, inputValue }, defaultFilter) => {
  if (!inputValue) return []

  return defaultFilter({hasGroups, items, inputValue})
}

...

<ComboBox items={[...]} customFilter={customFilter} />
```

### StateReducer

screen recording 2018-11-28 at 06 02 pm_1

`DropdownV2`'s internal state management system (Unistore... aka. super light Redux) has been enhanced to use a ["state reducer pattern"](https://blog.kentcdodds.com/the-state-reducer-pattern-%EF%B8%8F-b40316cfac57). This exposes the state changes externally, allowing the user to **change** the state when actions are performed, if they want to.

This can also be used like middleware. The above example showcases `console.log`'ing the state/action changes.

### Groups/Dividers

screen recording 2018-11-28 at 06 03 pm

This update adds grouping + divider support to Dropdown. Need groups with nested results? Sure! Check out the demo above ☝️ .

### Performance!! 🚀 

screen recording 2018-11-28 at 06 03 pm_1

This update **dramatically** improves the performance of interactions with the Dropdown, especially fast actions like mouse hover or key presses. The above GIF showcases how little CPU is being used as I hold down the `down` arrow key to page through all the dropdown items.

This is achieved by a special component called `Dropdown.Renderer`. It handles the rendering for interactions that occur when the Dropdown is open. It does this by manipulating the DOM directly. It's a bit of an.. React anti-pattern perhaps... However, the performance gains is worth it. Until React can natively support fast state manipulation + UI changes in a declarative way (which they're planning on doing with React Suspense), this is probably the best alternative we have.

Fun fact. This has been stress tested for 7000+ items. Note: Those items rendering into the DOM... that's another story. Interacting with them is fine tho!

screen recording 2018-11-28 at 06 17 pm

### Review

The best way to review is to open up the Storybook and poke around :)

Netlify linky:

https://deploy-preview-408--hsds-react.netlify.com/?selectedKind=DropdownV2&selectedStory=Dropdown%2FDefault&full=0&addons=1&stories=1&panelRight=0


ItsJonQ:
> Overall coverage decreased (-0.2%) to 99.834%

NOOOOO!!! Scuba Steve! Damn you!!!

ItsJonQ:
Hmm! Not sure what's up with the Netlify URL

src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,304 @@
...
+  renderFooter = () => {
+    const { renderFooter } = this.props
+    if (!renderFooter) return
+
+    return (
+      <Dropdown.Block className="c-ComboBoxFooter">
+        {renderRenderPropComponent(renderFooter)}
+      </Dropdown.Block>
+    )
+  }
```
knicklabs: @ItsJonQ Tiny nitpick, but given what we now know about how these static prop arrow function are being transpiled, such that they are defined in the constructor and so each instance cannot share an underlying implementation on the prototype, should we replace all of these (except where they are serving the purpose of binding) with instance methods?

ItsJonQ: @knicklabs You got it!!! :D

PR #411, opened 2018-11-30
Move help text inside the label when Choice is stacked

tompedals:
In a PR for Beacon Builder @ItsJonQ suggested making the help text part of the label so it's clickable when using the stacked style: https://github.com/helpscout/hs-app/pull/5934#issuecomment-442864744

This PR moves the help text inside the label only when it is stacked.

choice-stacked-helptext-label


tompedals:
I'm gunning for that 🥈 

screen shot 2018-11-30 at 17 30 40

PR #414, opened 2018-12-02
Improve build size

ItsJonQ:
## Improve build size

This update improves the build size of dist files by adjusting the Babel env config,
and replacing lodash.get with [dash-get](https://github.com/ItsJonQ/dash-get) (200b in size!)

Using some things we learned from the React conference on Dec 1 😊


ItsJonQ:
@knicklabs Thanks!!!

PR #416, opened 2018-12-03
Fix HelpText state and add stories

tompedals:
## Fix HelpText state and add stories

The `state` prop of `HelpText` was not being respected and it was always using the `faint` text style. This PR moves the state styling to the child `Text` component rather than adding styles to the `HelpText` component as they were are not specific enough.

screen shot 2018-12-03 at 19 05 36
screen shot 2018-12-03 at 19 05 31
screen shot 2018-12-03 at 19 05 27

I added a story to test these changes.

### DIFF
--- a/src/components/HelpText/HelpText.js
+++ b/src/components/HelpText/HelpText.js
@@ -26,6 +26,7 @@ const HelpText = (props: Props) => {
     'c-HelpText',
     isCompact && `is-compact`,
     shade && `is-${shade}`,
+    state && `is-${state}`,
     className
   )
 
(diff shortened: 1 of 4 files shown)


ItsJonQ:
@tompedals Gonna help fix tests

tompedals:
@ItsJonQ I'm just looking at the tests now

ItsJonQ:
@tompedals Looks like we just need to add `state && `is-${state}`,  back into the `classNames()` :).
Even thought it's not used to render styles, its' handy to have it in the classNames for `querySelector` or inspect element purposes

tompedals:
I need to rework this solution slightly, it doesn't work when passing in child components and not just a string.

tompedals:
@ItsJonQ I've updated the fix now so we keep the state styles on the `HelpText` component too for when we don't have string children. It wouldn't quite work if we have `<HelpText state="error"><Text>Something</Text></HelpText>` as the `Text` would still override the error state styles, but I think that is acceptable as we would just pass the string value usually instead.

ItsJonQ:
@tompedals Sounds good to me! Thank you!!!

PR #418, opened 2018-12-03
Add new heading size and story

knicklabs:
This is a small PR to:

1. Add a new heading size for Skeletons. The new small-sized heading should be the same height as text but with extra bottom margin to distinguish it from other lines of text.
2. Make a slight adjustment to the heigh and bottom margin of text. 

Example of small heading with text:

Demo


ItsJonQ:
@knicklabs Awesome! This looks great :D. Will merge when gren!

PR #422, opened 2018-12-05
Remove polished. Write own rgba method

ItsJonQ:
## Remove polished. Write own rgba method

This update removes the `polished` library, which was taking up quite a bit of
space when bundled.

The `rgba` method was replaced by custom method, composed of other color based
methods under `utilities/color`.


src/styles/utilities/__tests__/color.test.js
```diff
@@ -1,5 +1,5 @@
 import colors from '../../configs/colors'
-import { getColor, getThemeBrandProp } from '../color'
+import { getColor, getThemeBrandProp, rgba } from '../color'
```
Charca: Not that it really matters, but doesn't look like we're using the `rgba` method in this test, right?

ItsJonQ: Oops! You're right. I originally had the tests here. Will remove!! Thanks 😍

PR #437, opened 2019-01-03
Revamps Page component and subcomponents

tinkertrain:
# <Page>

## Problem
While working on settings pages and while trying to follow the most recent responsive designs I found that the structure in how the current `<Page>` components (and subcomponents like `<Card>` or  `<Header>`) are laid out offer little flexibility for trying to get those designs right.

Specifically, currently moving secondary headings to the side in large screens is not possible (without hacks):
hsds-product-ui wiki 2019-01-03 12-59-01

## Proposal
This PR proposes a revamp of the Page components.

Starting with the ideal markup (using pseudo classnames):

```html
<div class="Page">
  <div class="Card">
    <section class="Section">
      <header class="Header">
        <h1 class="Heading Title"></h1>
        <div class="Subtitle"></div>
      </header>
      <div class="Content"></div>
    </section>

    <section class="Section">
      <header class="Header">
        <h2 class="Secondary Heading"></h2>
        <div class="Subtitle"></div>
      </header>
      <div class="Content"></div>
    </section>
  </div>

  <div class="Card"><!-- ...content... --></div>
</div>
```

Notice:

- One Page can contain multiple Cards
- One Card can contain multiple Sections
- Each Section would normally be comprised of Header and Content
- Only one Card should have an `h1` (usually the one at the top), subsequent Sections and Cards should be `h2`

### 📄 Page
Usage remains identical, basically just a wrapper with a flag to indicate if it should be responsive or not

```jsx
<Page isResponsive></Page>
```

### ♠️ Page.Card
Usage remains almost the same, this component has been "dumbed" down and now is just a wrapper with styles, no responsive responsibilities for it.

```jsx
<Page isResponsive>
  <Page.Card></Page.Card>
  <Page.Card></Page.Card>
  <Page.Card></Page.Card>
</Page>
```

### 🗂 Page.Section (new 🌟)
This component divides each Card into sections and it handles responsiveness

```jsx
<Page isResponsive>
  <Page.Card>
    <Page.Section></Page.Section>
    <Page.Section></Page.Section>
    <Page.Section></Page.Section>
  </Page.Card>
</Page>
```

### 🗣 Page.Header (revamped 🚀)
This is the most changed component. The title and subtitle strings props, remain for backwards compatibility but as they were inflexible: unable to add anything but text (icons, links, etc), unable to change the heading level (always `h1`) is best to avoid them, I added a warning on docs stating their drawbacks.

In exchange, we get a `render` prop, which is a function with an object argument with 2 keys: Title and Subtitle.

Those 2 are React Components with all the pretty styles ready, use like this:

```jsx
<Page isResponsive>
  <Page.Card>
    <Page.Section>
      <Page.Header
        render={
          ({Title, Subtitle}) => (
            <div>
              <Title level="h1">Headers!</Title>
              <Subtitle>Oh so <strong>nice</strong></Subtitle>
            </div>
          )
        } />
    </Page.Section>
  </Page.Card>
</Page>
```

`Title` takes two props: `headingLevel`, with values of `"h1"` or `"h2"` and `isSecondary` (boolean)

* `headingLevel`: The html tag to use "h1" or "h2", has no effect on looks, just semantics
* `isSecondary`: This determines if it should look slightly smaller when true. Use when you need a secondary heading in the same Card.

### 🖼 Page.Content
Unchanged

```jsx
<Page isResponsive>
  <Page.Card>
    <Page.Section>
      <Page.Header
        render={
          ({Title, Subtitle}) => (
            <div>
              <Title level="h1">Headers!</Title>
              <Subtitle>Oh so <strong>nice</strong></Subtitle>
            </div>
          )
        } />
      <Page.Content>Some surely fascinating content would live here.</Page.Content>
    </Page.Section>
  </Page.Card>
</Page>
```

## Docs & Stories (revamped 🚀)
Updated with examples of everything

## Tests
Unit tests added for everything


ItsJonQ:
@tinkertrain This is amazing! Thank you for this proposal 😍 !

I only have a few thoughts!

### Fragments

For your examples, can we swap `React.Fragment` -> `div`? (We chatted about this briefly on Slack). Although the library + Storybook totally supports `Fragments`, I think it would be more helpful to have markup that library consumers can use (aka. copy/paste). Since the other React projects are still on 15, they can't use Fragments. I want to avoid the idea that Fragments may be necessary for the component to work

### render(Title, Subtitle)

I really like the concept of render props! 
I would vote for the `render` arguments to be an Object rather than an array of args. Something like:

```
render({ Title, Subtitle })
```

This feels easier to use IMO :).

Speaking of the `render` prop.. I haven't encountered an example of it providing Components as arguments before. Typically it provides functions or (enhanced) data.

I'm not opposed to the idea though!

### heading/subtitle -> render() change

This is defo a big change. With the update, it breaks previous implementations of `Page.Header`

Which is okay, assuming that after the changes are merged/released, we update all the uses of `Page.Header` to match the new implementation.

An alternative idea... is we keep the `heading/subtitle` props, and add the new `render` as an enhancement (rather than a replacement).. Sort of like providing the ability to do finer grain/custom rendering, if you choose.

Would love to hear your thoughts!

tinkertrain:
Thanks Q! Great feedback, I have made some updates:

1. React.Fragment in examples -> removed!
2. `render` prop signature, agreed! I have changed it to an object. As for passing components back, I wanted to avoid people having to remember the right classnames or importing yet more components and since these are only pertinent to Page.Header, I did it like this, I think is nice that you can do this kind of composition!
3. I added back the props, I thought about initially and didn't because as it stands you can't choose to make the headings secondary or h2's without having to add yet more props to Header. Because this is just for backwards compatibility I didn't add those abilities if using these props, and wrote a warning on the docs to try to avoid using them.

Cheers!

ItsJonQ:
@tinkertrain Haiii!! Just a heads up, I made a release for `2.5.7`

https://github.com/helpscout/hsds-react/releases/tag/v2.5.7

If you need to make more beta releases, you can run this:

```
npm run release -- --yolo --any-branch
```

You'll see the `np` CLI select screen:

```
❯ patch 	2.5.8
  minor 	2.6.0
  major 	3.0.0
  prepatch 	2.5.8-0
  preminor 	2.6.0-0
  premajor 	3.0.0-0
  prerelease 	2.5.8-0
  ──────────────
  Other (specify)
```

Go down to `Other`

And you can add a custom name.. like `2.5.8-page-beta-0` or whatever you see fit :).

I would merge in the latest from `master` branch though!

Hope this helps!

ItsJonQ:
@tinkertrain Awesome! Gonna do a final check in local storybook. Then will merge + release!

PR #454, opened 2019-01-25
PropProvider Context + Portal Fixes for React 16

ItsJonQ:
## PropProvider Context + Portal Fixes for React 16

screen recording 2019-01-25 at 04 50 pm

(GIF: Demo shows a iFrame rendered, Portal'ed component with buttons correctly rendering values defined by a parent `PropProvider`. In other words.. super hard mode!)

This update fixes the transfer for `PropProvider`'s context to content rendered
with `Portal`.

### contextConnect

During a React 16 update, we discovered that `PropProvider` values defined
outside of the portal'ed content were not being transferred. This is
due to how `ReactDOM.unstable_renderSubtreeIntoContainer` renders the content
vs. React 16's new context system.

The solution was to create a new HOC (called `contextConnect`), which provides
a component directly with the values from the closest `PropProvider`. Enhancing
`Portal` with this new HOC, I'm able to pass the values through the children
elements. Things work as expected!

### ReactDOM.createPortal?

I've experimented with this solution. It kinda works with HSDS's Portal mechanism. However, the
requirement to support React 15, iFrames, and programatic open/close makes this (native) solution
very unstable.

We should consider exploring this once we're fully migrated to React 16 :)

### Refactors

I made some refactors to the `Portal` file structure during this update. I decided against converting
things to Typescript (for now).


src/components/Portal/Portal.js
```diff
@@ -0,0 +1,196 @@
...
+    return mountSelector || this.document.body // fallback
+  }
+
+  renderPortalContent = props => {
+    const { propProviderContextApp, propProviderContextValue, children } = props
+    if (!children || !React.isValidElement(children)) return
+
+    this.portal = ReactDOM.unstable_renderSubtreeIntoContainer(
+      this,
+      <PropProvider
```
ItsJonQ: This is the magical bridge right here that transfers the PropProvider context props!

PR #456, opened 2019-01-28
Fix carriage returns

tjbo:
## Input: Fix carriage returns

This update fixes allows a flag to be set on an input 'hasCarriageReturns', which will allow when pressing `option`, `ctrl` or `meta` with enter for a carriage return to be inserted into the input.


tjbo:
[Move key event listeners for extra returns to HSDS](https://trello.com/c/e838Xj8F/870-move-key-event-listeners-for-extra-returns-to-hsds)

tjbo:
event.preventDefault works, thanks!

tjbo:
[HSDS: Fix empty return if no value](https://trello.com/c/PF8qpM9J/914-hsds-fix-empty-return-if-no-value)

PR #457, opened 2019-01-28
Fix carriage returns (Tests)

tjbo:
## Fix carriage returns (Tests)

In a former PR:
https://github.com/helpscout/hsds-react/pull/456

- We updated the value of an Input internally in HSDS, but it does not propagate up to the implementer, which is needed so that the implementer is aware of the value change.

- Fixed for 100% test coverage


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
```
tjbo: @ItsJonQ, should this be 2.6.2, now?

ItsJonQ:
@tjbo Hmm! This PR seems to undo an update I did on master:

https://github.com/helpscout/hsds-react/pull/457/files#diff-295fe7f8838adf9ff25c431419d04f41R8

Has this been merged/rebased with latest `master`?

Thanks!~

tjbo:
[Change default input placeholder text to be "Send a message" instead of "How can we help"](https://trello.com/c/Xw4apqY1/835-change-default-input-placeholder-text-to-be-send-a-message-instead-of-how-can-we-help)

PR #458, opened 2019-01-29
Enable Programmatic Control of Accordion Sections

knicklabs:
This update adds support for programatically controlling the sections of an Accordion. The sections can be open and closed by setting/updating the `openSectionIds` property on the `Accordion` component. For each id in the array, the corresponding section will be open. All others, will be closed. 

To achieve this, the `Section` component needed to be updated to allow the library user to set an `id` which supersedes the `uuid` generated by the `Section`. Should an id be provided, that id will be used to track the open/closed state of `Section` in the `Accordion`.

This feature supports programmatically opening open or more Sections at a time. 

Below is an example of one Section being controlled at a time.

Demo 1

Below is an example of multiple Sections being controlled at a time.

Demo 2

### DIFF
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -48,6 +48,11 @@ const buildOpenSections = sectionIds =>
     {}
   )
 
+const stringifyArray = arr => arr.sort().toString()
+
+const didOpenSectionIdsChange = (prevSectionIds, nextSectionIds) =>
+  stringifyArray(prevSectionIds) !== stringifyArray(nextSectionIds)
+
 class Accordion extends PureComponent<AccordionProps, AccordionState> {
   static Body = Body
   static Section = Section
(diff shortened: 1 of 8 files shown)


ItsJonQ:
@knicklabs Uh oh! looks like we might be missing code coverage on this:
https://coveralls.io/builds/21344182/source?filename=src%2Fcomponents%2FAccordion%2FAccordion.js#L86

😅

knicklabs:
@ItsJonQ I covered it with https://github.com/helpscout/hsds-react/pull/458/commits/f8352c45a7472c930a1a76d16632d7e3db1d99be#diff-9f4bfd720351bcd173b6b39d4988d0f3R134

Waiting on Travis to report back.

src/components/Accordion/Accordion.js
```diff
@@ -39,19 +39,60 @@ const getComponentClassName = ({
...
+    openSectionIds: [],
     size: 'md',
   }
 
-  state = {
-    sections: {},
+  constructor(props) {
+    super(props)
+
+    this.state = {
```
ItsJonQ: @knicklabs I believe you can just write:

```js
state = {
  sections: buildOpenSections(this.props.openSectionIds)
}
```

No constructor needed. The babel transformer should take care of it :)
I only learned about this like 2 months ago

knicklabs: I originally had something like that, but changed it when I started using the props to the set the initial state. Wasn't aware we could use `this.props` there.

knicklabs:
@ItsJonQ I added (and utilized) this little function here for cheap comparison of flat arrays.

https://github.com/helpscout/hsds-react/pull/458/files#diff-47616ab991b48c665258097c8664dea1R54

ItsJonQ:
@knicklabs Brilliant!!!!!

PR #460, opened 2019-01-30
Fix carriage returns: Modifier keys

tjbo:
## Fix carriage returns: Modifier keys

Adds a utility to check if modifier key is pressed.


tjbo:
@ItsJonQ , great catch. Fixed!

PR #464, opened 2019-01-31
Fixes inserting empty return if there is no value in field

tjbo:
## Fixes inserting empty return if there is no value in field

This update fixes empty carriage return insertions when there is no pre-existing value in the field.


tjbo:
[HSDS: Fix empty return if no value](https://trello.com/c/PF8qpM9J/914-hsds-fix-empty-return-if-no-value)

Issue #483, opened 2019-02-13
Add ability to "float" a button over an Input


brettjonesdev:
We need to be able to create components where a Button or other element is floated over an Input. 

![image](https://user-images.githubusercontent.com/2163045/52680303-0c90a680-2edc-11e9-9a22-e4fc9e9daa79.png)

PR #489, opened 2019-02-19
Input: Add embedded Action

ItsJonQ:
## Input: Add embedded Action

screen recording 2019-02-19 at 02 30 pm

This update adds the ability for `Input` to render an embedded `Action` within
the Component UI.

Actions (typically [Buttons](../Button)) can be embedded within the `Input` using the `action` prop:

```jsx
<Input
  action={
    <Button version={2} size="sm">
      Submit
    </Button>
  }
/>
```

The `Button` should be one step smaller compared to the `Input`. For example, if the `Input` has a size of `md` (default), then the `Button` should be `sm`. If the `Input` is `sm`, then the `Button` should be `xs`.

Note: Be sure the version for `Button` is at least version 2.

### Overflow

Since the new `action` UI uses the existing Input `Prefix`/`Suffix` system, overflow is automatically handled.

screen recording 2019-02-19 at 02 50 pm

Bonus: Multiple actions can also be rendered:

screen shot 2019-02-19 at 2 36 40 pm

Resolves https://github.com/helpscout/hsds-react/issues/483


src/components/Input/Input.js
```diff
@@ -501,13 +502,7 @@ export class Input extends Component<Props, State> {
...
     const { prefix, seamless } = this.props
 
-    return (
-      prefix && (
-        <Prefix className="c-Input__item c-Input__prefix" isSeamless={seamless}>
-          {prefix}
-        </Prefix>
-      )
-    )
+    return prefix && <Prefix isSeamless={seamless}>{prefix}</Prefix>
```
ItsJonQ: Refactored. `Prefix` renders the classNames by itself.

src/components/Input/Input.js
```diff
@@ -527,13 +522,13 @@ export class Input extends Component<Props, State> {
...
     const { suffix, seamless } = this.props
 
-    return (
-      suffix && (
-        <Suffix className="c-Input__item c-Input__suffix" isSeamless={seamless}>
-          {suffix}
-        </Suffix>
-      )
-    )
+    return suffix && <Suffix isSeamless={seamless}>{suffix}</Suffix>
```
ItsJonQ: Refactored. `Suffix` renders the classNames by itself.

tinkertrain:
🚀 Pretty cool!

PR #498, opened 2019-02-25
Add NavLink component

ItsJonQ:
## Add NavLink component

screen recording 2019-02-25 at 11 07 am

This update adds a new `NavLink` component, which works with `react-router` and our Application's Marionette set up.

`NavLink` requires `Route` from `react-router-dom` in order to determine the active state. Because of this, `react-router-dom` has been added as a peer dependency. The cut off is at `4.3.1`, which is the version before the library switches to the new React 16 `createContext` internal.

Note: `NavLink` is styleless, and works just like `react-router-dom`'s NavLink out of the box.


ItsJonQ:
### Demo

https://deploy-preview-498--hsds-react.netlify.com/?selectedKind=NavLink&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

src/components/NavLink/NavLink.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+NavLink.className = 'c-NavLink'
+NavLink.defaultProps = {
+  'aria-current': 'page',
+  activeClassName: 'active is-active',
+  activeStyle: {},
+  style: {},
+}
+
+const PropConnectedComponent = propConnect(COMPONENT_KEY, { pure: false })(
```
ItsJonQ: `{ pure: false }` is an internal option for `propConnect` to render a `React.Component` vs a `React.PureComponent`.

For this case, we cannot use a `PureComponent` as react-router does not play nicely with those

src/components/NavLink/NavLink.tsx
```diff
@@ -0,0 +1,71 @@
...
+  location,
+  activeClassName,
+  className,
+  activeStyle,
+  style,
+  isActive: getIsActive,
+  'aria-current': ariaCurrent,
+  ...rest
+}) => {
+  const path = typeof to === 'object' ? to.pathname : to
```
ItsJonQ: Note: 99% of the code here is copied/pasted from `react-router-dom@4.3.1`

src/components/NavLink/NavLink.tsx
```diff
@@ -0,0 +1,71 @@
...
+        )
+      }}
+    />
+  )
+}
+
+NavLink.className = 'c-NavLink'
+NavLink.defaultProps = {
+  'aria-current': 'page',
+  activeClassName: 'active is-active',
```
ItsJonQ: Added `is-active` to follow our conventions.

ItsJonQ:
@tinkertrain Thanks!! The story could be better. I just copied/pasted the BasicExample from `react-router`'s docs 😂

PR #502, opened 2019-02-26
Fix: Opening Autodropdown causes scroll to bottom

knicklabs:
This update addresses an issue where opening an autodropdown would cause the page to scroll to the bottom.

This issue impacted the autodropdown when the number of items exceeded the threshold that has us render a filterable dropdown menu. That is 15 items by default. It also was only a noticeable issue on pages with a scrollbar. However, when those conditions were met, the dropdown was unusable.

Here is a demo of the problem:

Problem

Here is a demo of how things are looking after the fix:

Fix

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -665,7 +665,8 @@ export class Input extends Component<Props, State> {
 
     const componentProps = {
       ...getValidProps(rest),
-      autoFocus,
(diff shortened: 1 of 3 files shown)


ItsJonQ:
@knicklabs lolol.. mind updating this test?

https://travis-ci.org/helpscout/hsds-react/builds/498916679#L532

:cry-through-pain:

PR #504, opened 2019-02-27
Fix appearance of disabled SplitButtons and SplitButtons with state

knicklabs:
Nice job on the SplitButton component @TerrenceLJones! Soon as it went out I scooped it up to use in app :)

This update improves the styles for the primary button with success + error state as well as the disabled states.

Here are some screenshots of before:

Screenshot 1: Before

Screenshot 2: Before

Here are some screenshots of after:

Screenshot 3: After

Screenshot 4: After

### DIFF
--- a/src/components/SplitButton/SplitButton.css.ts
+++ b/src/components/SplitButton/SplitButton.css.ts
@@ -1,6 +1,5 @@
 import styled from '../styled'
 import Button from '../Button'
-import { config } from '../Button/Button.css.js'
 import { getColor } from '../../styles/utilities/color'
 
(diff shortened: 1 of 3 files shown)


knicklabs:
@ItsJonQ Good call. I did not change the border color because we want that to stay the same, just the shadow should be impacted, hence the CSS changes you see remove the Button `config` and use a custom color specific to the `SplitButton` component.

Screenshot

I thought 600 looked most suitable.

ItsJonQ:
Will merge on green

TerrenceLJones:
@knicklabs Thanks for these refinements! 😍

PR #510, opened 2019-03-01
Update chat components to use raw body property

tompedals:
https://trello.com/c/zowspdvk/889-use-rawbody-rather-than-body-when-outputting-chat-messages

We are making changes to Chat API so that we do not strip HTML tags from messages and the autolinking of URLs is being moved to the front-end. This will allow us to escape the message body for output.

This PR assumes HS App and Beacon are passing the `rawBody` property to the HSDS components rather than the `body` property. `rawBody` will be the exact message the user sent, but `body` will have HTML stripped and URLs already converted to a link. We need to ensure we modify HS App and Beacon to use `rawBody` when we update the version of HSDS used so this may require at least a minor version bump.

I have added our own small utility to autolink URLs based upon the `UrlFinder` class we have in Chat API. There are some small differences between the implementations but majority of the test cases are the same.


tompedals:
@Charca What is a good way to test how a package would affect the bundle size of Beacon? I'm aware that adding `linkifyjs` will increase the size. We could possibly try to create our own simple library that does just what we need if it's too large.

ItsJonQ:
@tompedals This is the fastest:

https://bundlephobia.com/result?p=linkifyjs@2.1.8

(Love this site BTW)

TLDR: 9.3kb, gzipped.

Let's try to find another solution, if we could :)

ItsJonQ:
🍐 'ed with @tompedals . We're gonna try writing our own regex parser to handle the links. `linkifyjs` looks great, but it's too heavy for what we need.

Marking this PR as WIP

Charca:
Thanks guys! Yes, we're trying to hit our goal of <250kb (gzipped) bundle size, and we still have 120kb to go, so adding an extra 10kb will slow us down, sadly. I'd vote for a custom solution or, even better yet, continue to use the formatted body directly from Chat API, at least on embed. Unless there's a reason why we wouldn't like to do that?

tompedals:
@Charca I'm currently working on our own simple function to autolink like we have on the PHP side. It should be really small so not a problem. Here is some background on this change: https://paper.dropbox.com/doc/Chat-strip-tags-migration--AYvK9sLRbtoPH0GiBrgrP2FGAg-oXyj5y8Y5Duvix0IhMmQT

We are making the change so we don't strip the HTML tags from messages. However to do that we need to escape the output. We could possibly have Chat API return the body escaped for HTML but we use it in plain text emails for the transcript so would need an escaped and non-escaped version. The best solution seemed to be handle escaping and autolinking on the client side.

src/utilities/strings.ts
```diff
@@ -112,3 +112,80 @@ export const camelCase = (string: string): string => {
...
+
+/**
+ * Autolinks URLs in the given string.
+ *
+ * @param {string} string
+ *
+ * @returns {string}
+ */
+export const autolink = (string: string): string => {
+  const pattern = /(?:(https?:\/\/)|www\d{0,3}\.|www-|[a-z0-9.-]+\.[a-z]{2,4}(?=\/))(?:[^\s()<>]+)*(?:[^\s`!-()\[\]{};:'".,<>?«»“”‘’])/giu
```
tompedals: This matches the `UrlFinder` class in Chat API as much as possible. The only case we don't handle is `()` within the path of a URL. There is some regex used in the PHP implementation that isn't supported in JS.

tompedals:
@ItsJonQ @Charca I've removed linkifyjs now.

src/utilities/strings.ts
ItsJonQ: 👍 for me! You've added a bunch of test cases. If we run into any strange link parsing issues, we'll look into then and make this util better!

src/components/ChatTranscript/LineItem.js
```diff
@@ -28,7 +28,11 @@ const LineItem = (props: Props) => {
   ) : null
 
   const contentMarkup = body ? (
-    <span dangerouslySetInnerHTML={{ __html: newlineToHTML(body) }} />
+    <span
+      dangerouslySetInnerHTML={{
+        __html: newlineToHTML(escapeHTML(body)),
```
ItsJonQ: @tompedals I have a thought on tidying this up a bit! We have a neat util function in `react-utils` called [compose](https://github.com/helpscout/react-utils/blob/master/docs/compose.md). 

We can create a function like:

```js
const enhanceBody = compose(newlineToHTML, escapeHTML)

...

dangerouslySetInnerHTML={{
  __html: enhanceBody(body)
}}
```

or


```js
const htmlBody = compose(newlineToHTML, escapeHTML)(body)

...

dangerouslySetInnerHTML={{
  __html: htmlBody
}}
```

This pattern can be repeated for the other uses as well.

💖

src/utilities/strings.ts
```diff
@@ -112,3 +112,80 @@ export const camelCase = (string: string): string => {
...
+}
+
+/**
+ * Autolinks URLs in the given string.
+ *
+ * @param {string} string
+ *
+ * @returns {string}
+ */
+export const autolink = (string: string): string => {
```
ItsJonQ: @tompedals Looking at the name in isolation, I'm not sure what `autolink` does. I mean, I know what it does after :pear:'ing, but it's not as clear as something like `escapeHTML`.

Would you be okay with renaming `autolink` to something like `convertLinksToHTML`?

It's a little bit more verbose, but it's a lot clearer (to me at least!~)

src/utilities/strings.ts
```diff
@@ -112,3 +112,80 @@ export const camelCase = (string: string): string => {
...
+  return string.replace(
+    new RegExp(pattern),
+    (match: string, scheme): string => {
+      let url = match
+      if (!scheme) {
+        // Add http as the default scheme
+        url = `http://${url}`
+      }
+
+      return `<a href="${url}" target="_blank" rel="noopener">${match}</a>`
```
ItsJonQ: 🤔 This one is tricky. I didn't expect `target="_blank" and `rel="noopener"` to be there, but I understand why it is (for chat).

Since we only have one use case for this util, for now, I think it's okay. I'd leave some comments about `target` and `rel` though, just in case we forget about why it's there :D

PR #514, opened 2019-03-06
Feature/table

tinkertrain:
# New component `Table`

👇 

:white_check_mark: Includes 2 themes out of the box (for the 2 styles I saw in the updated doc, called them "default" and "alternative")
:white_check_mark: Custom themes
:white_check_mark: Custom cell content rendering
:white_check_mark: Custom header cell content rendering
:white_check_mark: Clickable rows
:white_check_mark: Handles long text in cells
:white_check_mark: Horizontally scrollable
:white_check_mark: Works with the Pagination component
:white_check_mark: "compound columns": the ability to render multiple data fields in one column
:white_check_mark: Sortable
:white_check_mark: Expandable
:white_check_mark: Lots of stories

The storybook includes a [fun playground](https://deploy-preview-514--hsds-react.netlify.com/?selectedKind=Table&selectedStory=default&full=0&addons=0&stories=0&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) to play with the table options and styles, give it a go!


.gitignore
```diff
@@ -30,3 +30,4 @@ yarn-debug.log*
 yarn-error.log*
 
 storybook-static
+.vscode/settings.json
```
ItsJonQ: @tinkertrain Whoa! How come? This was intentional :D. We want shared project settings

.vscode/settings.json
```diff
@@ -24,6 +24,10 @@
     "statusBar.foreground": "#ffffff",
     "statusBar.background": "#3197D6",
     "statusBar.noFolderBackground": "#3197D6",
-    "statusBar.debuggingBackground": "#3197D6"
+    "statusBar.debuggingBackground": "#3197D6",
```
ItsJonQ: Not sure what happened here...

.prettierrc
```diff
@@ -0,0 +1,3 @@
+semi: false
```
ItsJonQ: We should already have this file. We have a `.prettierrc.yml`

ItsJonQ:
@tinkertrain This is really awesome !! Really liking the direction of this so far! 😍 

P.S. The draft PR feature is neat!

tinkertrain:
[Create Sortable Table component](https://trello.com/c/65KytPoe/36-create-sortable-table-component)

.vscode/settings.json
```diff
@@ -24,7 +24,11 @@
     "statusBar.foreground": "#ffffff",
     "statusBar.background": "#1292EE",
     "statusBar.noFolderBackground": "#1292EE",
-    "statusBar.debuggingBackground": "#1292EE"
+    "statusBar.debuggingBackground": "#1292EE",
+    "titleBar.activeBackground": "#00b3e6",
```
ItsJonQ: @tinkertrain  Hmm? Not sure where this is coming from 😅

Can we remove it? It looks a little disruptive for me

src/components/Table/HeaderCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+        align={column.align}
+        cellWidth={column.width}
+        aria-sort={
+          sortedInfo &&
+          sortedInfo.columnKey === column.columnKey &&
+          sortedInfo.order
+            ? sortedInfo.order
+            : 'none'
+        }
+        onClick={() => {
```
ItsJonQ: @tinkertrain Can this be refactored to be a class method instead of an inline function? 😊

ItsJonQ:
@tinkertrain HaiiiiiiI!!! This looks **amazing** 😍 !

I know this is still a WIP, so I'm not sure how fine-grain you want the feedback to be 😊 .

Some feedback/ideas to float your way...

Customization/flexibility/composition looks good!

### onClick callback

What are your thoughts on providing the event as an argument? Maybe something like:

```
onClick: (event: Event, props: Row | Header | Cell) => void
```

That way, you can do `event.preventDefault` and things, if needed (and still have access to the data associated with the thing you clicked).

### Make sorting a little more magical

By that, I mean maybe make it so the consuming component doesn't have to take care of `isLoading`.

One (pretty good) way of doing this is for the table to manage it's own `isLoading` state, (which can be forcefully overridden via props, if desired)

When a sort event happens.. maybe the `onSort` prop is a Promise.
It can be sequenced like:

1. Click a header to sort
2. setState({ isLoading: true })
3. Call this.props.onSort()
4. this.props.onSort().then(() => this.setState({ isLoading: false })

☝️ something like that

Would love to 🍐 on this of course!

Hope this helps 💃

.vscode/settings.json
plbabin: I wonder if we should just removed settings.json from git. There is local customizations that we might want to do (per developper) and it should not be on the repo. If there is a way to separate local vs repo settings that would be even better

src/components/Table/Cell.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { CellUI } from './styles/Table.css'
+import { TABLE_CLASSNAME } from './Table'
+
+import { CellProps } from './types'
+
+export default class Cell extends React.PureComponent<CellProps> {
+  render() {
+    const { column, row } = this.props
+
+    if (Array.isArray(column.columnKey)) {
```
plbabin: I would split this render function.

if (Array.isArray(column.columnKey)) {
    this.renderListOfCells()
}else{
   this.renderSingleCell()
}

src/components/Table/__tests__/utils.js
```diff
@@ -0,0 +1,235 @@
...
+    jobTitle: faker.name.jobTitle(),
+    companyName: faker.company.companyName(),
+    days: faker.random.number(100),
+    lastSeen: derived(props => {
+      const { days } = props
+      return `${days} days ago`
+    }),
+    emails: emails.length === 1 ? emails[0] : emails,
+  }).generate(amount)
+}
```
plbabin: love the spec here, well done!

stories/Table/Table.stories.js
```diff
@@ -0,0 +1,164 @@
...
+import {
+  defaultTheme,
+  alternativeTheme,
+} from '../../src/components/Table/styles/themes'
+
+const stories = storiesOf('Table', module)
+
+stories.add('default', () => (
+  <div>
+    <PreviewCard style={{ marginBottom: '20px' }}>
```
plbabin: good idea to preview some code! 👍🏻

plbabin:
overall, amazing job 😍

something I noticed while testing in storybook: should we find a way to not overlap content on smaller screen: http://c.hlp.sc/5b2cc5813dd3 ?

One other small thing, maybe when changing the sort we could display the new sort direction right away instead of waiting for the result to come back...thoughts?

src/components/Table/HeaderCell.tsx
tinkertrain: It certainly can!

.vscode/settings.json
tinkertrain: I was about to make the same suggestion as @plbabin, I don't think editor settings should live in the repo, as they're or can be personal, people use different editors anyway. I didn't really intended to commit this change, I'm using this vscode extension that made the change https://marketplace.visualstudio.com/items?itemName=johnpapa.vscode-peacock

tinkertrain:
@ItsJonQ Thanks for looking at it and the suggestions! We of course can/should 🍐 , in the meantime let me respond:

> ### onClick callback
> 
> What are your thoughts on providing the event as an argument? Maybe something like:
> 
> ```
> onClick: (event: Event, props: Row | Header | Cell) => void
> ```

Makes sense to me and it would make consistent with other components. Will add!

 
> ### Make sorting a little more magical
> 
> By that, I mean maybe make it so the consuming component doesn't have to take care of `isLoading`.
> 
> One (pretty good) way of doing this is for the table to manage it's own `isLoading` state, (which can be forcefully overridden via props, if desired)
> 
> When a sort event happens.. maybe the `onSort` prop is a Promise.
> It can be sequenced like:
> 
>     1. Click a header to sort
> 
>     2. setState({ isLoading: true })
> 
>     3. Call this.props.onSort()
> 
>     4. this.props.onSort().then(() => this.setState({ isLoading: false })
> 
> 
> ☝️ something like that

I thought about adding that, here's my reasoning for getting this state out of the component itself.

As I started on the sorting feature I went with the intention of doing something similar to what you're suggesting, not exactly the same, but the gist of it: let the component handle it's own loading state.

As I was doing this, I thought, mmmm maybe the table should also get in the loading state when changing pages (using the `Pagination` component) and this is an interesting case, because neither component is aware of each other. `Table` doesn't have any prop related to pagination at all. So how can she (I've decided Table is a she) update itself in this case? And if you think about it, there might be lots of other cases where someone else performs an action that updates her contents without she necessarily knowing about it.

Sorting _could_ be treated as a special case, but I think I prefer to be consistent.

Let me know what you think! You might have a different way to solve/think about this.

cc @plbabin @knicklabs Would love to hear your input on this too

tinkertrain:
Thanks for looking at it @plbabin !

> overall, amazing job 😍
> 
> something I noticed while testing in storybook: should we find a way to not overlap content on smaller screen: http://c.hlp.sc/5b2cc5813dd3 ?

Ah yes, I thought to leave the handling of this to the user of the component by using `renderCell`, we have a `<Truncate>` component that I could have incorporated in the default rendering of the cell contents, but since a column can have any content, I thought you might have some unexpected rendering going on.

Although thinking more about it, the default will always be text, if we want to render an image we need to pass some html... so maybe I can just add it, if the user wants to handle the content in a different way, they can use `renderCell`.

> One other small thing, maybe when changing the sort we could display the new sort direction right away instead of waiting for the result to come back...thoughts?

Not sure about this... let me experiment a bit.

.vscode/settings.json
ItsJonQ: @tinkertrain + @plbabin Sure :) We can remove it. Especially if folks are experience some pain with it.

I'll create a PR for that right now.

ItsJonQ: https://github.com/helpscout/hsds-react/pull/565

ItsJonQ:
> (I've decided Table is a she)
😍 

I think can have the best of both worlds by making `isLoading` a controlled prop.

```
class Table extends React.PureComponent {
  state = {
    isLoading: this.props.isLoading
  }

  ...

  componentWillReceiveProps(nextProps) {
    if (nextProps.isLoading !== this.props.isLoading) {
      this.setState({
        isLoading: nextProps.isLoading
      })
    }
  }

  ...
}
```

So you can have automagic isLoading handling... but CAN override it when needed for reasons.

.vscode/settings.json
tinkertrain: 👏 Thanks!

src/components/Table/Cell.tsx
tinkertrain: Changed!

src/components/Table/HeaderCell.tsx
tinkertrain: Changed!

tinkertrain:
> > (I've decided Table is a she)
> > 😍
> 
> I think can have the best of both worlds by making `isLoading` a controlled prop.
> 
> ```
> class Table extends React.PureComponent {
>   state = {
>     isLoading: this.props.isLoading
>   }
> 
>   ...
> 
>   componentWillReceiveProps(nextProps) {
>     if (nextProps.isLoading !== this.props.isLoading) {
>       this.setState({
>         isLoading: nextProps.isLoading
>       })
>     }
>   }
> 
>   ...
> }
> ```
> 
> So you can have automagic isLoading handling... but CAN override it when needed for reasons.

Hi Q! I totally see what you are saying, I started to see how I would go about it but it just doesn't feel right to me.

If I look at the pagination with sorting story, this the sorter function:

```jsx
//...
const column = [{
          title: 'Customer (sorts by company)',
          columnKey: ['name', 'companyName'],
          sortKey: 'companyName',
          sorter: this.sortAlphabetically
}]

//...

sortAlphabetically = columnKey => {
    const { data, sortedInfo } = this.state

    this.setState({
      isLoading: true,
    })

    // simulate sortData as an API call
    sortData(data, columnKey, sortedInfo.order).then(sortedData => {
      this.setState({
        data: sortedData,
        sortedInfo: {
          columnKey,
          order: sortedInfo.order === 'descending' ? 'ascending' : 'descending',
        },
        isLoading: false,
      })
    })
  }
```

I like how this is super easy to see what is going on, if I extract the `isLoading` state, I don't see why the `data` and the `sortedInfo` should not receive the same treatment. But then the Table would have to also keep track of those internally via state.

The way sorting is implemented, you'll notice there really is no difference conceptually between it and pagination. Both actions are taken and Table doesn't know about them, Table does not paginate, and Table does not sort. She's kind of lazy. The only difference between sorting, paginating or some other action is that  the sorting action by default is fired by clicking an element inside the Table. I added this default to have styles ready, but this can be overridden if you use `renderHeaderCell` and in that case you can provide your own click target, like:

```jsx
renderHeaderCell: (column, { columnKey, order }) {
  	return (
		<div>
			<span>{column.title}</span>
			<button onclick={pleaseSortThisColumn}>
				{order === 'descending' ? '👆'  :'👇'}
			</button>
		</div>
	)
}

//...

function pleaseSortThisColumn() {}


```

Note: I added an example on the sorting story where the 3rd column has a custom render header.

I like how, for example, with the changes @plbabin made to Pagination, we can pass the same `isLoading` value to the Table and the Pagination, to have the same behaviour at the same time in both components without having to worry about it, in my mind sorting is no different.

I guess the biggest difference with this component and table components from other libraries is that all actions pertinent to the table data are external, because they are performed by the API and no the table itself.

BUT, you might disagree with me 😄 

What does @plbabin and @knicklabs think?

Thanks!

ItsJonQ:
@tinkertrain 

> She's kind of lazy

😂 ! Gotcha! That sort alphabetically example is a good point! Point taken :).

For the consuming application, they can create custom tables, using the HSDS: React Table as a base. And in that custom table, they can handle the implementation detail of `isLoading`.

Hopefully I got that right~!

ItsJonQ:
@tinkertrain I re-reviewed the code. Again, absolutely amazing work here. The only feedback I have involve tidying up stuff. Mostly:

* moving the `render() {}` method to the end. That's the convention we have in HSDS: React, so it was slightly jarring seeing it at the top 😅 
* refactoring longer bits of code/logic within inline props/render, like:

```
        {column.renderCell
          ? column.renderCell(cellData)
          : (Object as any)
              .values(cellData)
              .map(d => <div key={d.slice(4)}>{d}</div>)}
```

That's about it! I'm down to 🍐 if needed!

ItsJonQ:
@tinkertrain Also! If you could resolve the `.vscode/settings.json` merge issue, that would be awesome :). I merged in the PR that removes it from the project

src/components/Table/HeaderCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    return (
+      <HeaderCellUI
+        className={`${TABLE_CLASSNAME}__HeaderCell`}
+        align={column.align}
+        cellWidth={column.width}
+        aria-sort={
+          sortedInfo &&
+          sortedInfo.columnKey === column.columnKey &&
+          sortedInfo.order
```
knicklabs: For long and/or non-obvious conditions like this it would be great to break it out into a variable to help some reading the code so that they don't have to interpret it. I believe what's meant here could be expressed as:

```
const isSortedBy = sortedInfo && sortedInfo.columnKey === column.columnKey && sortedInfo.order
...
aria-sort={isSortedBy ? sortedInfo.order : 'none'}
```

Not a blocker though at all.

tinkertrain: Went ahead and refactored that bit, thanks for the suggestion!

tinkertrain:
> @tinkertrain Also! If you could resolve the `.vscode/settings.json` merge issue, that would be awesome :). I merged in the PR that removes it from the project

Done!

tinkertrain:
> @tinkertrain I re-reviewed the code. Again, absolutely amazing work here. The only feedback I have involve tidying up stuff. Mostly:
> 
>     * moving the `render() {}` method to the end. That's the convention we have in HSDS: React, so it was slightly jarring seeing it at the top 😅
> 
>     * refactoring longer bits of code/logic within inline props/render, like:
> 
> 
> ```
>         {column.renderCell
>           ? column.renderCell(cellData)
>           : (Object as any)
>               .values(cellData)
>               .map(d => <div key={d.slice(4)}>{d}</div>)}
> ```
> 
> That's about it! I'm down to 🍐 if needed!

I refactored some bits and moved `render` methods to the bottom of the class, as it is the convention :)

About that a little note:

I personally like `render` at the top, because it's the most important bit of a react component and when I open a file is normally the only thing I need to see. The rest is implementation detail.

You might be able to tell that I'm a big fan of hoisting in JavaScript (these are classes with methods so not hoisting, just the concept applies), not many people are though... also I'm lazy to scroll down haha

Thanks for the reviews and comments! 🤗

src/components/Table/HeaderCell.tsx
```diff
@@ -49,29 +47,32 @@ export default class HeaderCell extends React.PureComponent<HeaderCellProps> {
...
-          {sortedInfo &&
-            sortedInfo.columnKey === colKey &&
-            sortedInfo.order && (
-              <Icon
-                name={
-                  sortedInfo.order === 'descending' ? 'caret-down' : 'caret-up'
-                }
-              />
-            )}
+          {columnSortStatus !== 'none' && (
```
ItsJonQ: @tinkertrain Oo! Just a heads up, if you want to write this inline, I would evaluate it as a ternary, which returns `<Icon /> : null`

Reason! We've experienced weird issues in React where inline `&&` like this actually renders the string `false` in the UI. Wild. I have no idea. We experienced this in HS App.

Since then, I've been ensuring that the case doesn't happen, `null` is returned.
I remember KCD mentioning something about this as well in one of his many screencasts.

😅 

(This applies to all places this logic is used to render)

src/components/Table/Table.tsx
```diff
@@ -51,6 +51,47 @@ export class Table extends React.PureComponent<TableProps, TableState> {
...
+    const tableWrapperClassNames = classNames(
+      `${TABLE_CLASSNAME}__Wrapper`,
+      isLoading && 'is-loading',
+      isTableCollapsed && 'is-collapsed',
+      className
+    )
+    const tableClassNames = classNames(
+      TABLE_CLASSNAME,
+      isLoading && 'is-loading',
+      Boolean(onRowClick) && 'with-clickable-rows',
```
ItsJonQ: @tinkertrain What is this sorcery? I don't think I've ever seen `Boolean()` like that before, haha

src/components/Table/Row.tsx
```diff
@@ -7,6 +7,11 @@ import { generateCellKey } from './Table.utils'
 import { RowProps } from './types'
 
 export default class Row extends React.PureComponent<RowProps> {
+  handleRowClick = e => {
+    const { row, onRowClick } = this.props
+    onRowClick && onRowClick(row, e)
```
ItsJonQ: Refactor suggestion!

I would add a `defaultProp = { onRowClick: noop }` for this component. That way, `onRowClick` is always there :).

As a user, I also expected the argument order to be `onRowClick(event, row)`. I guess I'm just use to `event` being first? 

😊

src/components/Table/HeaderCell.tsx
tinkertrain: Ohh thanks for the heads up, I'll change it

src/components/Table/Table.tsx
ItsJonQ: Whoa! What on earth is `Number('101')`!!!!
Learning so many new things! Hahaha

tinkertrain: Sorry for assuming...! `Number` and `Boolean` are both constructors for primitives in JS (you also have `String` and maybe I'm forgetting more), so you can declare a number in any of these ways:

```
var a = new Number('123')
var b = Number('123')
var c = 123
```

It's not very common to use, except sometimes without new.

From MDN
> In a non-constructor context (i.e., without the new operator), Number can be used to perform a type conversion.

ItsJonQ:
@tinkertrain + @knicklabs + @jaredmcdaniel 

I believe @knicklabs needs this for his CustomerList prototype work. Since we've 👍 the code, would it be okay to merge/release it (minor). And do adjustments (if required) after the fact?

(cc'ing @jaredmcdaniel )

ItsJonQ:
🍐 with @jaredmcdaniel . Okay with merge. Will follow up to make it 1:1 with CustomerList when we can.

PR #515, opened 2019-03-06
Dropdown: Fix ComboBox positioning with shouldDropDirectionUpdate

ItsJonQ:
## Dropdown: Fix ComboBox positioning with shouldDropDirectionUpdate

screen recording 2019-03-06 at 09 46 am

(ComboBox, by default, should always drop down, if defined)

This update fixes the up/down menu position of `ComboBox` based dropdowns
that happens when the list size changes (during filtering).

The solution was to introduce a new prop called `shouldDropDirectionUpdate`.
By default, in `Dropdown`, it resolves to `true`. However, in `ComboBox`,
we'll resolve it to `false`.

The user can adjust this, if required.

`Dropdown` auto positioning works as before:
screen recording 2019-03-06 at 09 45 am

#### Screencast

I recorded a screencast for additional details:

https://www.useloom.com/share/be89d280e6984512a707571e35e15cbc

---

Future enhancement idea:

Ideally, the menu position **should** auto calculate for `ComboBox`, but only on the initial render. Subsequent updates shouldn't change it's positioning. There are more edge cases with this, but I think this is the end-goal.

For now, this "hard coded" solution is simpler and more stable.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -22,8 +22,8 @@
   "eslint.enable": true,
   "workbench.colorCustomizations": {
     "statusBar.foreground": "#ffffff",
-    "statusBar.background": "#3197D6",
-    "statusBar.noFolderBackground": "#3197D6",
-    "statusBar.debuggingBackground": "#3197D6"
+    "statusBar.background": "#1292EE",
+    "statusBar.noFolderBackground": "#1292EE",
+    "statusBar.debuggingBackground": "#1292EE"
   }
 }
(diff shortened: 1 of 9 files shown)


ItsJonQ:
## Demo URL

https://deploy-preview-515--hsds-react.netlify.com/?selectedKind=ComboBox&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

.vscode/settings.json
```diff
@@ -22,8 +22,8 @@
   "eslint.enable": true,
   "workbench.colorCustomizations": {
     "statusBar.foreground": "#ffffff",
-    "statusBar.background": "#3197D6",
-    "statusBar.noFolderBackground": "#3197D6",
-    "statusBar.debuggingBackground": "#3197D6"
+    "statusBar.background": "#1292EE",
```
ItsJonQ: New blue 😎

ItsJonQ:
@tinkertrain Thanks!!

PR #517, opened 2019-03-07
Improve the utility to convert URLs (and emails) to links

tompedals:
https://trello.com/c/zowspdvk/889-use-rawbody-rather-than-body-when-outputting-chat-messages

During testing we discovered some test strings with URLs that were not handled correctly. There were some side effects when escaping the string before converting URLs to links. To resolve this issue we now escape the string for output at the same time as converting URLs to links. I have also added support for converting emails to links.

Lots of extra test cases have been added.

### DIFF
--- a/src/utilities/__tests__/strings.test.js
+++ b/src/utilities/__tests__/strings.test.js
@@ -259,7 +259,7 @@ describe('convertLinksToHTML', () => {
     ],
     [
       'www.example.com/example?fdasfs=24fa3fd32&asdsa=af5t34tw',
-      '<a href="http://www.example.com/example?fdasfs=24fa3fd32&asdsa=af5t34tw" target="_blank" rel="noopener">www.example.com/example?fdasfs=24fa3fd32&asdsa=af5t34tw</a>',
+      '<a href="http://www.example.com/example?fdasfs=24fa3fd32&amp;asdsa=af5t34tw" target="_blank" rel="noopener">www.example.com/example?fdasfs=24fa3fd32&amp;asdsa=af5t34tw</a>',
     ],
     [
       'http://www.example.com/example?fdasfs=24fa3fd32',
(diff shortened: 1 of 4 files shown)


src/utilities/strings.ts
```diff
@@ -184,20 +184,39 @@ export const convertLinksToHTML = (string: string): string => {
...
-  return string.replace(
-    new RegExp(pattern),
-    (match: string, scheme): string => {
-      let url = match
-      if (!scheme) {
-        // Add http as the default scheme
-        url = `http://${url}`
+  const emailPattern = "\\b[A-Z0-9._'%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,4}"
+  const urlPattern =
+    '(?:(?:https?:\\/\\/|www\\d{0,3}\\.|www-|[a-z0-9.-]+\\.[a-z]{2,4}(?=\\/))(?:[^\\s()<>]+)*(?:[^\\s`!-()\\[\\]{};:\'".,<>?«»“”‘’]))'
```
tompedals: Parsing strings for URLs and emails is hard. We took this simplified approach to reduce the bundle size in Beacon. It seems to do a pretty good job now, but there will be some edge cases.

src/utilities/__tests__/strings.test.js
```diff
@@ -359,19 +359,103 @@ describe('convertLinksToHTML', () => {
...
+      'example@example.com',
+      '<a href="mailto:example@example.com">example@example.com</a>',
     ],
     [
-      'www.example.com www.helpscout.com',
-      '<a href="http://www.example.com" target="_blank" rel="noopener">www.example.com</a> <a href="http://www.helpscout.com" target="_blank" rel="noopener">www.helpscout.com</a>',
+      "example.a'ddress@example.com",
+      '<a href="mailto:example.a&#x27;ddress@example.com">example.a&#x27;ddress@example.com</a>',
+    ],
+    [
```
ItsJonQ: Oh wow! New test cases

tompedals:
@ItsJonQ Whoops, I forgot the very important label :) Yep, patch bump for this one please.

PR #521, opened 2019-03-11
New & Updated Icons

plbabin:
This update includes thousands-ish icons....ok **128 new** icons and **61 updated**. All icons are from HSDS-core-usage, and this PR does not include _Small_, _Cards_, _Beacon_, _Ratings_ and _Emoji_ icons. All icons also keep their mask and groups from sketch to minimize the amount of effort of keeping both `hsds-react` and `hsds-core`  in sync.

Image 2019-03-11 at 1 58 26 PM

## What changed?
In this update a couple of icons are different compare to what we actually have in HSDS-react.

Image 2019-03-08 at 2 29 36 PM

## Script all the things
I created a new script to process icon, to speed up the time if takes to export from sketch and normalized/minify an icon

`npm run process:icons -- --path=path/to/icons/`

You can specify another output directory if you don't want to save the files directly in our icons folder.

 `npm run process:icons -- --path=path/to/icons/ --exportPath=path/to/output/`

This script will do in order:
1. Take all svg from the given directory
2. Rename the svg file to a slugified name (Icon Name.svg) to icon-name.svg
3. Minify and normalize the svg code using SVGO
4. Display in message in your terminal if the icon doesn't exists in the output directory
5. Save the icon

## Storybook
I also take some time to improve our icons page in storybook. I always thought that this page could have some love. I added a grid and a border around each icon. Everything on each line is align, and readable. 

Hope you like this update!

**After**
image

**Before**
image


ItsJonQ:
@plbabin Hmm! Looks like some of the tests are broken. I think it can be fixed by adding some guards around the regex <3

plbabin:
> @plbabin Hmm! Looks like some of the tests are broken. I think it can be fixed by adding some guards around the regex <3

sorry about that, I pushed an update that should fix the issue

jaredmcdaniel:
@plbabin or @ItsJonQ Comin' in hot with another n00b question! 😄 

Do Sketch and HSDS-React really need to be in sync on this? The masking/grouping definitely allow for flexibility on the design side, but ultimately generate 🤢 svg code on export. Doesn't seem like we should be looking to that as production ready markup. Instead, shouldn't we be leveraging the most optimal svg code for react? Additionally, it seems like since there's not really any ongoing maintenance with icons that taking the time to do this right up front would be ideal.

Example: Moon icon

image

Sketch:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 24 24">
  <defs>
    <path id="hsds-icons-moon-a" d="M13 6c-3.309 0-6 2.691-6 6s2.691 6 6 6c.78 0 1.542-.15 2.254-.44A8 8 0 0 1 13 12c0-2.09.822-4.08 2.254-5.56A5.962 5.962 0 0 0 13 6m0 14c-4.411 0-8-3.589-8-8s3.589-8 8-8c1.642 0 3.223.498 4.572 1.44a1.001 1.001 0 0 1 0 1.64A6.01 6.01 0 0 0 15 12a6.01 6.01 0 0 0 2.572 4.92.999.999 0 0 1 0 1.64A7.949 7.949 0 0 1 13 20"></path>
  </defs>
  <g fill-rule="evenodd">
    <mask id="hsds-icons-moon-b">
      <use xlink:href="#hsds-icons-moon-a"></use>
    </mask>
    <use xlink:href="#hsds-icons-moon-a"></use>
    <g mask="url(#hsds-icons-moon-b)">
      <rect></rect>
    </g>
  </g>
</svg>
```

vs optimized svg markup

```
<svg id="Layer_1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
    <path d="M17.6 7.1c.2-.2.4-.5.4-.8 0-.3-.2-.6-.4-.8-1.4-1-3-1.5-4.6-1.5-4.4 0-8 3.6-8 8s3.6 8 8 8c1.6 0 3.2-.5 4.6-1.4.3-.2.4-.5.4-.8 0-.3-.2-.6-.4-.8-1.6-1.2-2.6-3-2.6-5s1-3.8 2.6-4.9zM13 12c0 2.1.8 4.1 2.3 5.6-.8.2-1.5.4-2.3.4-3.3 0-6-2.7-6-6s2.7-6 6-6c.8 0 1.5.2 2.3.4C13.8 7.9 13 9.9 13 12z"/>
</svg>
```

Happy to help get you what you need if we think it's worth it.

plbabin:
@jaredmcdaniel yes that would be my prefered option for sure. I will always root for a better optimize code in production.

Originally, that's what we tried to shoot for. The `process:icons` stripped all `mask`, `g` and `xlink` but sometime the mask is too tied with the `path` (extruding some shape to create the icon). The goal was to find a way to have a repeatable pattern when extracting icons from Sketch so any developer or even designer can easily add new icons to `hsds-react`. 

If we could have a Sketch document containing all icons, without any `mask` (only `paths`), that we could export/import with the SVGO plugins to strip all Sketch trash code, I think that would be the best option.

Happy to discuss that further tomorrow when we are all online!

digitaldesigner:
This is a super cool update and I'm all in... **but**... I'm worried about
>In this update a couple of icons are different compare to what we actually have in HSDS-react.
Because what you've shown there in the "old" row are also totally valid icons that we currently use. The search icon, for example, sometimes we use the small variant, sometimes the large — depends on the component it's contained within (e.g. the search in the main nav is big, but when used inside an input, it's small). Same for arrows, stars etc.

Maybe I've misunderstood, but when we created HSDS we didn't change any icon — we only added a couple. To serve as an example, here are a couple of places we use icons to denote chat. These are 3 or 4 separate icons I think: 

chat

If it helps, I've created a Sketch file which contains the raw shapes from HSDS without their masks: http://c.hlp.sc/b62a54c6a83f

plbabin:
> This is a super cool update and I'm all in... **but**... I'm worried about
> 
> > In this update a couple of icons are different compare to what we actually have in HSDS-react.
> 

First thank you for the Sketch file, I will do some testing with it.

For the icons that are different, the sizing can be set by the `Icon` component. Also for search we have three different svg size `search-small`, `search-medium`, `search`. So when I say they changed, it's doesn't mean that the icon doesn't exists anymore, it just that for the same name it will not render exactly the same icon as before (based on the name in sketch and the name we have in hsds-react)

The two icons that are outstanding are chat and user. If you open storybook, `assign` seem to look like the old user, and `chat-square` will do for the "old" chat icon. We can do the swap where the icons are actually used, or we can update the name in sketch.

Hope that clarify everything!

ItsJonQ:
🍐 with PL! Gonna to hold on review until new SVG icon shapes have been added.
THanks a bunch @digitaldesigner  and @jaredmcdaniel !

plbabin:
I did update this PR with all icons sent by Buzz. All SVGs are optimize through SVGO. 

I also added a couple of missing icons (all the one under the Emoji sub group in sketch)

.vscode/settings.json
```diff
@@ -25,5 +25,6 @@
     "statusBar.background": "#1292EE",
     "statusBar.noFolderBackground": "#1292EE",
     "statusBar.debuggingBackground": "#1292EE"
-  }
+  },
+  "svg.preview.background": "black"
```
ItsJonQ: Oh, neat!!

ItsJonQ:
@plbabin This is 🚀 from me! Thank you for taking this on.

@digitaldesigner + @jaredmcdaniel Haiii!!

If you folks are 👍 on the changes, we can merge/ship this :)

Direct link to deploy preview:

https://deploy-preview-521--hsds-react.netlify.com/?selectedKind=Icon&selectedStory=icons&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

ItsJonQ:
Awesome! I'll get this merged up!

Issue #528, opened 2019-03-13
Reduce bundle impact of Icon, Illo, Huzzah


ItsJonQ:
![b826ed855582ea31d59a94b77d6ae8b2 _Image 2019-03-13 at 3 06 21 PM](https://user-images.githubusercontent.com/2322354/54304219-10771d80-459b-11e9-9ba6-24f85b2a941f.png)

ItsJonQ:
Working on this now to unblock the v2.17.x releases

ItsJonQ:
Working branch:

https://github.com/helpscout/hsds-react/tree/icon-reduce-bundle-size

ItsJonQ:
Sigh... no luck. Tried doing dynamic `require` a bunch of different ways. It either includes the all, or non at all (and breaks)

ItsJonQ:
🍐 'ed with @Charca . We're gonna try leveraging `env` variables to determine which icon set to load.

Rough notes:

```
Icons
  <- 128 Icons


<Icon name="..." />


import ActivityIcon from '../../../'


ENV to determine imports


Icons.js
  if (process.env.HSDS_ENV === 'beacon') {
    require('./BeaconIcons')
  } else {
    require('./AllIcons)
  }

build: 'HSDS_ENV='beacon' webpack build...'

```

PR #539, opened 2019-03-16
Reduce import size (of SVG) with adapters

ItsJonQ:
## Reduce import size (of SVG) with adapters

This update (dramatically) improves the size of HSDS: React for apps like Beacon2 by making the `svg` based `Icon` and `Illo` sets **opt-in**.

This is done by importing the appropriate adapter (similar to how Enzyme 3 requires an adapter).

The adapter would be added to the entry point of your app, typically the `index.js` file:

```
import '@helpscout/hsds-react/adapters/embed'
```

### Loading / Unloading SVG

To manually load SVG sets without an adapter, you can use the `load` function from this component:

```js
import { load } from '@helpscout/hsds-react/components/Illo'

const customSVGSet = { ... }

load(customSVGSet)
```

To unload, or reset, the SVG set, you can use the `unload` function:

```js
import { unload } from '@helpscout/hsds-react/components/Illo'

unload()
```

### Results:

These are the webpack bundle analyzed results:

**Before (2.17.0)**

Parse Size: 98kb
GZipped: 29kb

**After (2.18.0-0)**

Parse Size: 20kb
GZipped: 6kb

**Better After (2.18.0.2)**

Parse Size: 8kb
GZipped: ~3kb

Screen Shot 2019-03-18 at 10 19 17 AM

That's a saving of almost 90kb! This is just for `Icon`. There are addition (minor) savings from `Illo` as well.

## The Downsides

With this method, there is one:

(Thanks to @knicklabs , there's only 1. Previously, there was 2, which involved adding the icon set to global, lolol)

1. It requires the consuming app to add an adapter. It's not a huge deal. Just an extra step, but it's something. (`Illo` and `Icon` still work without the adapter. They just render blanks)

## Things I've tried

This was the only thing that worked across many (many) things I've tried. Some of these failed attempts included:

* Converting svg -> React components, then dynamically importing them
* Requiring certain sets based on an `process.env` (great idea @Charca , but it didn't work 😭 )
* Trying to create a compiler that could encode svg -> tiny strings, then decode during render (crazy, but I tried, lol)
* Tried tree shaking with Webpack

I've also considered creating a custom babel plugin to handle the transformations during compiling.

:crylaughing:

---

## Review

One way to test this out is to adjust the adapter `.storybook/config.js` brings in. By default, it'll use the `src/adapters/app.js` adapter. This can be swapped for embed.

Alternatively, you can install `@helpscout/hsds-react@2.18.0-1` within `beacon2` and give it a spin. Be sure to include the adapter in the main `index.js`!

@knicklabs , @tinkertrain , @Charca , @plbabin , would love your opinion on this approach. If you guys have a better idea, I'm all 👂 !!!

This idea was inspired by the initializer/adapter-like patterned I 🍐 'ed with @knicklabs on. It isn't exactly like what we discussed... but spiritually similar :).

Resolves: https://github.com/helpscout/hsds-react/issues/528


knicklabs:
What if we published the icon sets as separate modules on NPM? We could call them `IconPacks` and we could have `hs-app-iconpack` and `hs-beacon-iconpack`.

Why do this? 

We could keep the bundle size of HSDS-React to a minimum and ship it without icons. We can add the icons that we need only if we need them. This could come in handy should we ever use HSDS-React in another project and we have no requirement for icons.

Here's what I'm thinking. We take advantage of the fact that ES6 modules are singletons and we use a private variable to hold the map of icons. We export a setter for that variable so that the icons can be set at run time. All you need to do is have an entry point early in the app that invokes the setter with an IconPack before the Icon component is used. 

Here's a code sandbox with an example of how I'm imaging this to work.

https://codesandbox.io/s/lyzj9jmx7q?fontsize=14

ItsJonQ:
@knicklabs This is it!!!!!!! 😍 

I'll give this a shot when I can!!!

ItsJonQ:
@knicklabs Updated my implementation! Thank you so much. Also updated tests + docs

src/components/Icon/icons.embed.js
```diff
@@ -0,0 +1,101 @@
...
+  search,
+  spinner,
+  'tick-large': tickLarge,
+  'tick-small': tickSmall,
+  tick: tickLarge,
+  trash,
+  user,
+}
+
+export default ICONS
```
Charca: This list seems to include more icons than those needed in embed. I just audited our usage of Icons in the codebase and came up with this list:

```
fullscreen
search
helpscout-logo
arrow-left image
attachment
minimize
image-add
meatball
cross-small
tick tick-small
chat
```

If we could reduce it to only these icons, the component would be a third of its current size :)

ItsJonQ: @Charca Would you be able to double check that list? From that list, I'm sure we have (at least) the follow in Embed:

* `search`
* `helpscout-logo`
* `chat`
* `attachment`
* `image-add`
* `meatball`
* `arrow-left`
* `tick-small`

I just did a global search, and was able to find all the ☝️ from my list

Charca: Haha yes! Sorry I wasn't clear about that!

> P.S. I need to include certain icons like alert, spinner, and cross as it comes with other components (like error UIs and Modal)

That's perfect. Thank you!

ItsJonQ:
@Charca With that update, icons in Beacon Embed are now:

**8.31kb, 2.76kb gzipped!**

Screen Shot 2019-03-18 at 10 19 17 AM

src/components/Icon/README.md
```diff
@@ -2,6 +2,42 @@
...
+### Loading / Unloading SVG
+
+To manually load SVG sets without an adapter, you can use the `load` function from this component:
+
+```js
+import { load } from '@helpscout/hsds-react/components/Icon'
+
+const customSVGSet = { ... }
+
+load(customSVGSet)
```
knicklabs: Excellent that this is exported to make this flexible for future needs! ❤️

ItsJonQ:
Thanks @knicklabs + @Charca !!!!

src/components/Icon/__tests__/Icon.utils.test.js
```diff
@@ -0,0 +1,33 @@
+import { renameSVGIds } from '../utils'
+
+describe('renameSVGIds', () => {
```
plbabin: thank you adding tests for this utils! ❤️

PR #540, opened 2019-03-19
Updates Scrollable component to support horizontal scrolling

tinkertrain:
This PR updates `Scrollable` to allow the same functionality it provides but for horizontal scrolling.

This allows the following UI:

Slack - Help Scout 03-19 18-36

Horizontal mode is responsive, and it is enabled by using either the prop `fadeLeft`,`fadeRight` or both.

Vertical functionality should work as before.

As an extra improvement, the ScrollLock component does not work for firefox, this PR adds the `overscroll-behavior` rules and now Scrollable behaves as expected on most browsers with the combination of both solutions.

More info on [overscroll-behavior](http://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior)

## Screenshots:

Horizontal (on load):
Storybook 03-19 18-40

Horizontal (both sides):
Storybook 03-19 18-39

All in!
Storybook 03-19 18-41

Added stories and tests.


src/components/Scrollable/Scrollable.css.ts
```diff
@@ -8,6 +8,7 @@ export const ScrollableUI = styled('div')`
   box-sizing: border-box;
   display: flex;
   max-height: 100%;
+  height: ${props => (props.fadeSides ? '100%' : 'auto')};
```
plbabin: ah that's neat, I didn't know about that feature. 💯 _implement that right away in the SideNavigation component_

tinkertrain:
> Impressive work!
> 
> One thing that I am not 💯sure is that you can't drag the scrollbar unless you are at the end of the overflowing div. http://c.hlp.sc/9d38b9021207
> 
> Not sure if it was like that before your update or not (cc @ItsJonQ). If it's not a regression then I am fine shipping your update (and then we could tackle improving that later)

Thanks and good catch! I don't think is a regression, but I was able to fix it easily by adding one simple css rule on the faders (`pointer-events: none)

plbabin:
> Thanks and good catch! I don't think is a regression, but I was able to fix it easily by adding one simple css rule on the faders (`pointer-events: none`)

awesome, thank you!

src/components/Scrollable/Scrollable.css.ts
tinkertrain: I think this is my favourite feature of styled components, makes the js in css worth it imo!

ItsJonQ: 😍

src/components/Scrollable/Scrollable.tsx
```diff
@@ -188,13 +270,24 @@ export class Scrollable extends React.PureComponent<Props> {
...
       <ScrollableUI
         {...getValidProps(rest)}
+        fadeSides={fadeLeft || fadeRight}
         className={componentClassName}
         innerRef={innerRef}
       >
+        {(fadeLeft || fadeRight) && (
+          <EventListener
+            event="resize"
+            handler={() => {
```
ItsJonQ: @tinkertrain Would you be able to make this a class method 😊 (for performance reasons). Creating an inline-arrow functions means a new `Object` every time, which triggers prop change, which triggers React render cycle stuff 🤣

stories/Scrollable.stories.js
```diff
@@ -38,3 +37,60 @@ stories.add('Default', () => (
     </Scrollable>
   </Card>
 ))
+
+const MainContainer = styled('div')`
+  ${baseStyles} width: ${props => (props.fullWidth ? '100%' : '300px')};
```
ItsJonQ: @tinkertrain OoO! I found that adding a `;` at the end of things like:

```
${baseStyles}
```

helps with Prettier indentation 👍

src/components/Scrollable/Scrollable.tsx
tinkertrain: Good catch, just changed!

stories/Scrollable.stories.js
tinkertrain: That drove me crazy, thanks!

tinkertrain:
Thanks for the feedback @ItsJonQ , I have pushed a commit with the changes you suggested, the uncovered code was not straightforward to test, had to max my google-fu 😆

ItsJonQ:
🍐 'ed with @tinkertrain on the final testing this. We should have an update soon!
Once that's pushed + Travis is green, I'll merge it up :)

PR #549, opened 2019-03-22
Dropdown: Fix top/left position calculation with window.scroll values

ItsJonQ:
## Dropdown: Fix top/left position calculation with window.scroll values

Screen Recording 2019-03-22 at 01 06 PM

This update fixes the issue with Dropdown (V2) where it did not factor in
`window.scrollY` and `window.scrollX` with it's `top` and `left`
calculations.

The fix involved creating a special util function to generated the computed
`top` and `left` values, which are used within the `Dropdown.MenuContainer`
position methods.


ItsJonQ:
Will patch fix 2.16.x and 2.20.x

Issue #524, opened 2019-03-12
Esc close handling of a Dropdown within a Modal


ItsJonQ:
Identified by @knicklabs  and @plbabin

ItsJonQ:
![Screen Recording 2019-03-14 at 08 01 PM](https://user-images.githubusercontent.com/2322354/54399298-20265d00-4694-11e9-8b87-f60a28baa1b2.gif)

Complete Modal, Dropdown, Popover, Tooltip, ESC sequence awareness!!!

ItsJonQ:
@ryan-mulrooney I think you gonna like dat ☝️

ItsJonQ:
Working branch:

https://github.com/helpscout/hsds-react/tree/esc-stack-dropdown-modal

PR #550, opened 2019-03-22
Esc Stack: Add support for Modals, Dropdowns, Tooltips, and Popovers

ItsJonQ:
## Esc Stack: Add support for Modals, Dropdowns, Tooltips, and Popovers

Screen Recording 2019-03-14 at 08 01 PM

This update adds complete Modal, Dropdown, Popover, Tooltip, ESC sequence awareness!!!

The solution was to leverage `PortalWrapper` for any `Portal` mechanism within the various components. `PortalWrapper` has an internal manager system to coordinate and track the render sequence of components.

The "press ESC to close" logic has been removed from Tooltip, Popover, and Dropdown, since `PortalWrapper` now handles that.

In order to have.. no sequential closing.. for example, if you had a situation like:

```
Modal
  Dropdown (Opened)
```

And you closed the Modal via clicking `X` or the Overlay, we needed to add a new "force" close mechanism to Portal.

Resolves: https://github.com/helpscout/hsds-react/issues/524


config/jsdomPolyfills.js
```diff
@@ -7,21 +7,34 @@
...
-    clearTimeout(id);
-  };
+    clearTimeout(id)
+  }
 
-if (!window.SVGElement)
-  window.SVGElement = window.Element
+if (!window.SVGElement) window.SVGElement = window.Element
+
+if (!document.createRange)
```
ItsJonQ: `document.createRange` is used in the `popper.js` library. Need to stub it because JSDOM doesn't support this API

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
```diff
@@ -310,6 +310,15 @@ export class MenuContainer extends React.PureComponent<Props> {
     this.didOpen = false
     // End the reposition cycle
     cancelAnimationFrame(this.positionRAF)
+    this.props.closeDropdown()
+    this.focusTriggerNode()
+  }
+
+  focusTriggerNode = () => {
+    const { triggerNode } = this.props
```
ItsJonQ: `PortalWrapper` does have a refocus trigger mechanism. However, since Dropdown doesn't set the trigger node in the standard `PortalWrapper` manner, we have to handle it ourselves.

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
```diff
@@ -418,41 +427,40 @@ export class MenuContainer extends React.PureComponent<Props> {
 
     return (
       <div className="DropdownV2MenuContainerRoot" ref={this.setWrapperNode}>
-        {isOpen && (
-          <Portal
-            id={this.id}
-            onOpen={this.onPortalOpen}
-            onClose={this.onPortalClose}
+        <MenuPortal
```
ItsJonQ: Always renders vs. conditional rendering on `isOpen`. Instead, the `children` content will conditionally render based on `isOpen={isOpen}`

src/components/Dropdown/V2/__tests__/Dropdown.test.tsx
```diff
@@ -21,77 +21,6 @@ jest.mock('../Dropdown.MenuContainer', () => {
   }
 })
 
-describe('ESC: Keyboard events', () => {
```
ItsJonQ: Now being handled by `PortalWrapper`

src/components/Pop/Pop.js
```diff
@@ -1,13 +1,10 @@
 // @flow
 import type { PopProps } from './types'
 import React, { Component } from 'react'
-import EventListener from '../EventListener'
```
ItsJonQ: Unused 😎

src/components/Pop/__tests__/Arrow.test.js
```diff
@@ -82,3 +82,26 @@ describe('getPosition', () => {
     expect(getPosition('left-end')).toBe('end')
   })
 })
+
+describe('sanitizeStyles', () => {
+  test('Returns NaN as zero', () => {
```
ItsJonQ: This only occurs within Jest.. cause `popper.js` + JSDOM reasons.

plbabin:
nicely done Q!! Did some testing on Netlify and it seems to be working as expected. 

One quick question, we would still have to integrate the onPortalOpen/onPortalClose callbacks with our hs-app stacks utility to make them work together right?

ItsJonQ:
@plbabin Thanks!!

Hmm! I'm not sure. If the stacked components are all HSDS: React, than you shouldn't need to do anything. But we may need to do something to enable BackboneMarionette views + React components.

src/components/Pop/Arrow.js
```diff
@@ -123,4 +123,16 @@ export const getPosition = (placement: string) => {
...
+// This is primarily for popper.js within the test JSDOM environment
+// https://github.com/FezVrasta/popper.js/issues/478
+export const sanitizeStyles = style => {
+  const { left, top } = style
+
+  return {
+    ...style,
+    left: isNaN(left) ? 0 : left,
+    top: isNaN(top) ? 0 : top,
+  }
```
knicklabs: We should maybe make a util function for this. Like `safeParseInt` that returns a zero for NaN values.

ItsJonQ: @knicklabs OoOOooO! I like that!

ItsJonQ:
@plbabin  + @knicklabs Thanks for the review!!

ItsJonQ:
Will merge this in next week (unless urgent).
Reason is to double check that recent HSDS: React updates this week get integrated smoothly 😎

PR #552, opened 2019-04-01
Tag: Close icon fix

ItsJonQ:
## Tag: Close icon fix

Screen Shot 2019-04-01 at 8 59 44 AM

This update fixes the close icon within the Tag component. The solution
was to provide the alias for `cross`, which was removed during the icon
refactor.

The Toolbar component story was updated for improved rendering.


ItsJonQ:
@Charca Thanks!

PR #556, opened 2019-04-02
Button: Add new size and state to V2

TerrenceLJones:
## Button: Add new size and state to V2

This update adds a new size and primary state to the V2 Button.

For contrast reasons, the button color for the new `warning` state uses a base of `600` instead of `500`. 

A new size was added to Button. Size `lgxl`. This enables the button to render with a height of `50px`, which matches the requirements for the Beacon Embed design.

### DIFF
--- a/src/components/Button/Button.css.js
+++ b/src/components/Button/Button.css.js
@@ -44,6 +44,17 @@ export const config = {
       colorHover: 'white',
       colorActive: 'white',
     },
+    warning: {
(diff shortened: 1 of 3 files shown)


TerrenceLJones:
[PreviousMessages: Add Reply](https://trello.com/c/QEs7kDGs/144-previousmessages-add-reply)

TerrenceLJones:
Thanks @ItsJonQ. 🥇

PR #563, opened 2019-04-04
Blank slate component

plbabin:
## Blank slate component

This update add a new BlankSlate component. The component can displays a heading, an illustration and / or a message based on what is passed in the props

### Illustration
We can display any illustration from the [Illo](https://hsds-react.netlify.com/?selectedKind=Illo&selectedStory=Illo&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) component. If we need a new illustration for a given context, we can simply add the illustration to the component

### Customization
The content will be vertically and horizontally centered in the available area. We can top aligned using the `alignTop` prop. 

A light background can also be applied to the component with the `lightBackground` prop.

Screen Recording 2019-04-04 at 12 21 PM


digitaldesigner:
We should probably update the Chat UI blank slate (e.g. the "Chat is Off" screen) to use this component. Currently I think we freestyled it. cc/ @ItsJonQ

src/components/BlankSlate/BlankSlate.css.js
```diff
@@ -0,0 +1,57 @@
...
+`
+
+export const BlankSlateUI = styled('div')`
+  display: flex;
+  width: 100%;
+  height: 100%;
+  justify-content: center;
+  text-align: center;
+  flex-direction: column;
+  padding: 56px 0 68px;
```
tinkertrain: For the container we might want to add some padding on the sides too, to avoid these scenarios:

Storybook 2019-04-05 13-09-47

plbabin: nice catch, I'll update the padding.

ItsJonQ:
Awesome job! @plbabin ! Thanks for the review @tinkertrain + @digitaldesigner 

Merging/releasing!

PR #565, opened 2019-04-05
Remove VSCode Settings

ItsJonQ:
## Remove VSCode Settings

This update removes localized `.vscode/settings.json` from the project


knicklabs:
@ItsJonQ I think these are pretty useful default settings for the project. Any reason that we do not want to share them?

ItsJonQ:
@knicklabs Discussion here:

https://github.com/helpscout/hsds-react/pull/514/files#r272274668

ItsJonQ:
Gonna merge this in! Can always add it back if we feel like we need :)

PR #566, opened 2019-04-05
Improve pagination component

plbabin:
This update add 3 small improvements to the Pagination component

### Keyboard shortcut
Pressing `j` or `k` will update the current page
Screen Recording 2019-04-05 at 01 57 PM

### Formatting number
Current range and total of items are now formatted with the us format: 111,111.00
Image 2019-04-05 at 1 47 18 PM

### isLoading state
If the prop isLoading is active, all navigations button will be disabled
Screen Recording 2019-04-05 at 01 58 PM

### DIFF
--- a/src/utilities/number.js
+++ b/src/utilities/number.js
@@ -7,3 +7,10 @@ export const getMiddleIndex = (number: string): number => {
 
   return isOdd(number) ? middle : middle - 1
 }
+
+export const formatNumber = num => {
+  if (num === null || num === undefined) {
+    return num
+  }
+  return num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,')
+}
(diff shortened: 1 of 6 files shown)


src/utilities/number.js
```diff
@@ -7,3 +7,10 @@ export const getMiddleIndex = (number: string): number => {
 
   return isOdd(number) ? middle : middle - 1
 }
+
+export const formatNumber = num => {
+  if (num === null || num === undefined) {
```
ItsJonQ: @plbabin There's a `isDefined` util from `utilities/is` that checks for this :D

ItsJonQ: If you need a number, you can also check for `isNumber()`

plbabin: oh nice!! I updated the util in 85d5c88a

PR #567, opened 2019-04-08
TabBar component

plbabin:
This update add a new `TabBar` component. It's a simple wrapper around `Toolbar` and `Nav` to make it easy on the developer to create a tab navigation component

Image 2019-04-08 at 1 39 21 PM

### Secondary content
This component support a secondary content placeholder where we can render pretty much anything. Here is an example with a dropdown:
Screen Recording 2019-04-08 at 01 39 PM

### Alignment

The component support aligning the navigation menu to the left, right or center

Screen Recording 2019-04-10 at 11 37 AM


ItsJonQ:
> an easy variant of this component where it can be centrally aligned

@digitaldesigner Ooo! Good idea.

@plbabin Maybe we can add a `align` prop to this component to support `left`/`center`/`right` alignment?

plbabin:
@digitaldesigner I added a prop to align (left, right, center) the navigation.

secondary content will be move accordingly on the left/right version. Not sure what to do for the center one though

Screen Recording 2019-04-10 at 11 37 AM

digitaldesigner:
@plbabin I think in the case of the centered version it would be safe to have the secondary content remain right-aligned — we really only ever use this space for a single toggle (Greeters) or nothing at all

knicklabs:
@plbabin @ItsJonQ or @digitaldesigner I feel like there shouldn't be any non-clickable space between the tabs in the bar. It feels like I need to be too precise with mouse movement to click on a tab. I would expect that as I move the mouse across the x-axis that it would always be clickable with the click target highlighted by the hover state. But that's currently not the case. As a consequence, it's possible to move the mouse into the tab bar and miss any target. 

Here's an example to illustrate what I mean:

Example

What do you all think?

plbabin:
Agree, it’s on my list of things to do today. I wanted to sync with Q first because it will impact the Nav component

ItsJonQ:
@plbabin Let's 🍐 !

plbabin:
I updated the styling of the nav to remove the margin

src/components/TabBar/TabBar.css.ts
```diff
@@ -0,0 +1,69 @@
...
+  switch (align) {
+    case 'right':
+      return 'row-reverse'
+    default:
+      return 'row'
+  }
+}
+
+export const TabBarUI = styled('nav')`
+  --BlueConfigGlobalFontSize: 14px;
```
ItsJonQ: @plbabin We have a handy font util now to allow you to set the font-sizes for things:

Example:
https://github.com/helpscout/hsds-react/blob/82b707b52548e0feeb522963472c99eaae707f36/src/components/GreeterCard/GreeterCard.css.ts#L38

I would use this in this `.css.ts` file 😎  (within `SecContentUI` as well)

plbabin: fixed in b651d7f

PR #576, opened 2019-04-11
Heading: Add wordWrap (and other props from Text)

ItsJonQ:
## Heading: Add wordWrap (and other props from Text)

This update enhances the `Heading` component to have similar props as the
base `Text` component, including `wordWrap`, `noWrap`, `truncate`, and
`lineHeightInherit`.

This update also moves the `Heading.css.ts` file to the root `Heading/` directory.

Resolves: https://github.com/helpscout/hsds-react/issues/575


plbabin:
code looks good! should we add some stories to demo the heading in different situation?

ItsJonQ:
@Charca Thanks !

@plbabin Good idea :D. Will update the story for the Header (with knobs). Might be tricky to visually see things like line-height changes, but will add anyway!

ItsJonQ:
@plbabin Here we go!

Screen Recording 2019-04-15 at 10 51 AM

PR #578, opened 2019-04-12
Add Cypress tests for Pagination component

plbabin:
This update adds a list of cypress tests for the Pagination component

Image 2019-04-12 at 4 19 10 PM

### Notes
This is the first time I am working with Cypress. Feel free to give advice or request some changes to this PR

Next step would be to add cypress to travis-ci, and then add Percy to the integration suite

### DIFF
--- a/cypress/integration/Pagination.js
+++ b/cypress/integration/Pagination.js
@@ -1,0 +1,94 @@
+beforeEach(() => {
+  cy.visitStory('Pagination', 'in action')
+})
+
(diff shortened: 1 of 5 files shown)


ItsJonQ:
@plbabin This looks great 😍 . We currently don't have Travis running Cypress, but we can. A Travis + Cypress test run will take approx 10 mins (vs. the 6-7 mins we have now). Not a huge deal, especially if we wanna rely on it more for tests 💪

PR #580, opened 2019-04-15
Dropdown V2: Add Multi-select support

tinkertrain:
## Dropdown V2: Add Multi-select support

This update allows Dropdown to have multiple option selection, by adding the `allowMultipleSelection` prop

It adds defaults styles for selected elements following the design spec that can be found here: https://paper.dropbox.com/doc/Spec-Filter-component--AbRUHsgn~ImYFuvbjwqgxOIIAg-9SLZ97eyZqrZPvq55GBK9

## Selection "Clearer"

In this PR is also included a feature to render a "selection clearer" as the first item in the dropdown menu.

To use, add the prop `selectionClearer` with any string value, this value will be used as a label for the item, for example: "All Items"

When present, if nothing is selected, this item is selected:

Storybook 2019-04-18 14-07-56

Selecting anything else, deselects it:

Storybook 2019-04-18 14-08-41

### onSelect

An item can be selected and deselected (if the dropdown is stateful) when clicked, so I added an extra prop (`deselected`) to the object argument on the `onSelect` callback so the user can know about this. I added this to the documentation.

I also added a `selection` field, which contains an array with the whole selection

### Fixes

While working on this I found a couple of undesirable things that happened when `closeOnSelect` was set to `false` and the user selected an item:

1. The dropdown menu would immediately scroll to the top
2. The focus would switch to the 1st item on the list

These are now fixed and the interactions feel more natural as a result.


ItsJonQ:
@tinkertrain Recorded a screencast with troubleshooting details 😊 

https://www.loom.com/share/97f6f21e1b1440898b7fc4abcc3fb231

src/components/Dropdown/V2/Dropdown.Item.tsx
```diff
@@ -82,8 +83,13 @@ export class Item extends React.PureComponent<Props> {
 
   handleOnClick = (event: Event) => {
     const { onClick } = this.props
+    const state: any = this.props.getState()
 
-    onClick(event, { hasSubMenu: this.hasSubMenu() })
+    if (state && state.allowMultipleSelection && state.selectionClearer) {
+      onClick(state, event)
```
ItsJonQ: @tinkertrain 🤔 Hmm!! Is there a reason the argument order is reversed here? `(state, event)` vs `(event, state)`

src/components/Dropdown/V2/Dropdown.ItemSelectedCheck.tsx
```diff
@@ -0,0 +1,34 @@
...
+    isClearerActive && 'is-selectionClearer-active'
+  )
+
+  return (
+    <ItemSelectedCheckUI className={componentClassnames}>
+      <span className="c-ItemSelectedCheck__value">{props.value}</span>
+      {props.isActive || isClearerActive ? <Icon name="check" /> : null}
+    </ItemSelectedCheckUI>
+  )
+}
```
ItsJonQ: @tinkertrain I'm cool with using an SFC here. No need to try to TypeScript this thing... From what I remember, trying to type an SFC was nightmare-ish.

I would add some `defaultProps` as a static though! 😊

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
```diff
@@ -217,9 +229,43 @@ export class MenuContainer extends React.PureComponent<Props> {
...
+      allowMultipleSelection,
+      clearSelection,
+      focusItem,
+      selectionClearer,
+    } = this.props
+
+    if (allowMultipleSelection && selectionClearer) {
+      const clearerItem = {
+        value: selectionClearer,
+        id: createUniqueIDFactory('hsds-dropdown-v2-theallclearer')(),
```
ItsJonQ: @tinkertrain Curious!

Can this ID factory be moved to the new `ItemSelectedCheck.tsx` component? If not, can the resulting function (from the factory) be cached at the top of this `MenuContainer`? And that fn be instantiated inline here

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
```diff
@@ -217,9 +229,43 @@ export class MenuContainer extends React.PureComponent<Props> {
...
+      }
+
+      return [clearerItem].concat(items).map((item, index) => {
+        /* istanbul ignore next */
+        const indexProp = withIndex ? index : undefined
+        if (item.id === clearerItem.id) {
+          return (
+            <Item
+              {...this.getItemProps(item, indexProp)}
+              className="c-SelectionClearerItem"
```
ItsJonQ: @tinkertrain Just a heads up, `getItemProps` returns a `className` key with some computed class names. To pass in `c-SelectionClearerItem` into the mix, you'd have to.. combine it with `getItemProps().className`

If this logic is too complicated here... perhaps we can add a `isClearerItem` prop to `<Item />`, which then adds it's own special className?

src/components/Dropdown/V2/Dropdown.Renderer.tsx
```diff
@@ -282,7 +305,7 @@ class Renderer extends React.PureComponent<any> {
     // It may feel "wrong"... But, this is FAR cheaper than
     // relying on React to do it.
 
-    // That is because we're doing with a single (more or less) calculcation
+    // That is because we're doing with a single (more or less) calculation
```
ItsJonQ: @tinkertrain LOL thank you for the typo fix <3

src/components/Dropdown/V2/Dropdown.types.ts
```diff
@@ -24,6 +24,7 @@ export interface DropdownState extends DropdownMenuDimensions {
 
 export interface DropdownProps extends DropdownMenuDimensions {
   activeClassName: string
+  allowMultipleSelection?: boolean
```
ItsJonQ: @tinkertrain If it's a boolean, we don't need to make it optional. We'll just add a defaultProp of `false` :)

src/components/Dropdown/V2/Dropdown.utils.ts
```diff
@@ -336,3 +359,10 @@ export const filterNonStoreProps = props => {
     return nextProps
   }, {})
 }
+
+export const isSelectedItemEmpty = selectedItem => {
+  if (selectedItem == null) return true
```
ItsJonQ: @tinkertrain We have some handy `is` based helpers from `utilities/is`.. stuff like `isDefined`, `isArray`. It might help here 💪

src/components/Dropdown/V2/__tests__/Dropdown.Item.test.tsx
```diff
@@ -157,6 +202,33 @@ describe('Events', () => {
...
+      allowMultipleSelection: true,
+      selectionClearer: 'All Items',
+    }
+    const wrapper = mount(<Item onClick={spy} getState={() => state} />)
+
+    wrapper.simulate('click')
+
+    expect(spy).toHaveBeenCalledWith(
+      state,
+      expect.objectContaining({
```
ItsJonQ: @tinkertrain Whoa! I've never seen this before 😮

src/components/Dropdown/V2/__tests__/Dropdown.Item.test.tsx
```diff
@@ -226,6 +298,43 @@ describe('renderItem', () => {
...
+      />
+    )
+
+    const el = wrapper.find('div.ron')
+
+    expect(el.length).toBeTruthy()
+  })
+
+  test('Renders default markup (itemSelectedCheck) for item when multiselect enabled', () => {
+    const wrapper = shallow(<Item label="Champ" value="hello" />)
```
ItsJonQ: @tinkertrain Curious! How come this one is shallow mounted? :)

src/components/Dropdown/V2/Dropdown.css.js
```diff
@@ -147,6 +147,15 @@ export const ItemUI = styled('div')`
     }
   }
 
+  &.c-SelectionClearerItem {
+    border-bottom: 1px solid ${getColor('grey.600')};
```
ItsJonQ: Screen Recording 2019-04-18 at 11 32 AM

@tinkertrain UI wise, is there a way to add the `Divider` component in between the clearer + list of items? The reason is because it's already styled with the desired margins :).

If not, would we be able to update these styles to look the same? Specfically margin/padding:

Screen Shot 2019-04-18 at 11 32 13 AM

(Notice the selected BG is touching the divider line)

Thanks!! <3

src/components/Dropdown/V2/Dropdown.Item.tsx
tinkertrain: oh, mmm, in dropdown.actions, I created `clearSelection` following `selectItem`, which is defined as: `export const selectItem = (state, event, eventTarget) => { //... }`

So it's going to be switched in one place or the other! 😁

src/components/Dropdown/V2/Dropdown.ItemSelectedCheck.tsx
tinkertrain: 👍

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
tinkertrain: Cached to the top, what I'm doing is creating a "special" item to prepend it to the rest of the items before we decide to use ItemSelectedCheck

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
tinkertrain: I see, thanks!

src/components/Dropdown/V2/Dropdown.Renderer.tsx
tinkertrain: hehe!

src/components/Dropdown/V2/Dropdown.types.ts
tinkertrain: Ah yeah... thanks!

src/components/Dropdown/V2/Dropdown.utils.ts
tinkertrain: Handy!

src/components/Dropdown/V2/__tests__/Dropdown.Item.test.tsx
tinkertrain: Me neither to be honest! I wanted to check that this function was called with the correct state, and didn't care about the shape of `Event`, there's no "partial" `toHaveBeenCalledWith` but I found this, which allows you to just check that an object has a certain field, I just chose one that is present in `Event` and voila!

src/components/Dropdown/V2/__tests__/Dropdown.Item.test.tsx
tinkertrain: no reason! I was trying something and this is a leftover

src/components/Dropdown/V2/Dropdown.css.js
tinkertrain: Done!

tinkertrain:
> This looks really good. I didn't find anything else to fix in the code.
> 
> However I found a bug with the keyboard navigation. If you hit KEY/Down you can navigate to the last entry of the list:

Good catch!!

tinkertrain:
> One other UI thing, I wonder if we should deactivate the autocomplete in the `AutoDropdown` search input:
> Screen Recording 2019-04-19 at 09 34 AM

Ahh I never see that because I have form autofill deactivated on my browsers, but I think is a good idea. Let's ask @jaredmcdaniel

plbabin:
with the keyboard nav fix, it looks good to me 👍

Issue #581, opened 2019-04-15
Avatar: Safari Image rendering issue


ItsJonQ:
![Screen Recording 2019-04-15 at 01 47 PM](https://user-images.githubusercontent.com/2322354/56153811-03c45b80-5f85-11e9-9d36-a6259b499e46.gif)

☝️ dramatically slowed down.

Can be seen in Mac/iOS Safari.

Looks like `overflow: hidden` plus transition breaks things.

PR #582, opened 2019-04-15
Avatar: Render fix for Safari (Mac/iOS)

ItsJonQ:
## Avatar: Render fix for Safari (Mac/iOS)

Screen Recording 2019-04-15 at 02 08 PM

This update fixes the Avatar image crop rendering for Safari on iOS and Mac OS.
The solution involved migrating the transition styles to a parent DOM node,
in this case, another wrapping `div`, instead of applying it directly on
the `div.image`.

To simplify the (pretty complex) Avatar component, `Avatar.Image` and `Avatar.Crop`
sub-components were created to contain the rendering logic for those pieces
of UI.

This has been tested on both Mac OS and iOS (iPhone 8).

Resolves: https://github.com/helpscout/hsds-react/issues/581

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -4,16 +4,15 @@ import propConnect from '../PropProvider/propConnect'
 import { StatusDotStatus } from '../StatusDot/types'
 import { AvatarShape, AvatarSize } from './Avatar.types'
 import StatusDot from '../StatusDot'
-import VisuallyHidden from '../VisuallyHidden'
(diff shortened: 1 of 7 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -184,42 +186,6 @@ export class Avatar extends React.PureComponent<Props, State> {
     return <TitleUI className={componentClassName}>{text}</TitleUI>
   }
 
-  getContentMarkup = () => {
```
ItsJonQ: All of this logic was migrated to the sub-components

PR #585, opened 2019-04-16
Route: Fork react-router/Route to remove dependency

ItsJonQ:
## Route: Fork react-router/Route to remove dependency

This update forks the `Route` component from `react-router-dom@4.3.1` to be
used within `NavLink`. This allows for us to remove a (peer) dependency for
`react-router-dom`. It also allows for us to have better control for the
desired version of `Route` to interface with other HSDS components.

This also allows you to import components from HSDS: React in a tree-shakable fashion
without needing `react-router-dom` as a dependency.

Tested in Storybook:
Screen Recording 2019-04-16 at 01 36 PM

Screen Recording 2019-04-16 at 01 37 PM


ItsJonQ:
☝️ Neat! Thanks Travis for catching my FAILURES! :lolcry:

knicklabs:
@ItsJonQ I think maintaining a fork of React Router's Route may make this less robust because its possible to set peer dependencies that are compatible with multiple versions, but now we'll be restricted the whatever the version was when it was forked / last maintained. 

I wish NPM had optional peer dependencies. But alas. What do you think about leaving it as a peer dependency but also having a fork of this version. You could also import `{ Route as ReactRouterRoute} from 'react-router-dom'` and fallback to the forked `Route` component if `ReactRouterRoute` is undefined. The peer dependency warning could then be safely ignored. 

Just throwing an alternative idea out there. I'm happy to proceed with this approach if others feel its best. I think ever approach here leaves something to be desired due to shortcomings with the way dependencies work.

ItsJonQ:
@knicklabs 🤔 I'm not sure how to make that work.

For example, at the moment (not this PR), if you do something like this:

```
import Nav from '@helpscout/hsds-react/component/Nav'
```

and you don't have `react-router-dom` installed, things break.

The only thing I can think of.. is to `try` to require the module from `react-router-dom`? And use that, otherwise, fall back to our fork?

(Hope that makes sense!)

knicklabs:
Yeah, you're right. I forgot that it will fail if the library was missing. It will only fail silently if you try to import something that doesn't exist from a library that does. 

Back to the drawing board, yeah, a require inside a try block would work. That feels even less palatable though. On top of the fact that it's generating warnings even though we'd be handling it when it's missing. NPM so needs optional peer deps.

Anyway, I just needed to talk this out. I think we can go forward with this.

ItsJonQ:
@knicklabs 

> Anyway, I just needed to talk this out. I think we can go forward with this.

I'll pause until next week :). Maybe give us more time to think about it.

No real rush on this one!

ItsJonQ:
Changing it to a minor bump. Just in case!

PR #590, opened 2019-04-23
Improve auto linking to allow longer TLDs

tompedals:
## Improve auto linking to allow longer TLDs 

https://trello.com/c/ybWLWC8N/1162-beacon-20-chat-design-domain-not-recognized-as-email-add

We were only matching domains with TLDs that were up to 4 characters long. With all of the new TLDs the maximum length is now 24 characters. This PR fixes this.

While fixing the TLD length issue I also found that we had some other problems with the regular expression and we were matching things like `www.test./path`. I have reworked the regular expression and added more test cases.

### DIFF
--- a/src/utilities/strings.ts
+++ b/src/utilities/strings.ts
@@ -293,7 +293,7 @@ export const convertLinksToHTML = (string: string): string => {
 
         if (url.match(new RegExp(`^${domainPattern}$`, 'ui'))) {
           // Only matched a domain name (without subdomain)
-          // Skip this as it could be the end/start of a sentance without whitespace.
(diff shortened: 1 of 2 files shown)


src/utilities/strings.ts
```diff
@@ -202,6 +290,14 @@ export const convertLinksToHTML = (string: string): string => {
 
         // Matched a URL
         let url = value
+
+        if (url.match(new RegExp(`^${domainPattern}$`, 'ui'))) {
+          // Only matched a domain name (without subdomain)
+          // Skip this as it could be the end/start of a sentance without whitespace.
```
TerrenceLJones: `sentance` is misspelled.

PR #594, opened 2019-04-24
Upgrades Jest to v24

tinkertrain:
## Upgrades Jest to v24

Jest 24 is the latest version of the testing framework, we'd like to keep this dependency up-to-date.

More info [here](https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly) and changelog [here](https://github.com/facebook/jest/blob/master/CHANGELOG.md#2400)

This version includes some support for typescript, but it is only traspilation through babel, so we will continue to use ts-jest, which gets upgraded in this PR too. [Here's](https://github.com/kulshekhar/ts-jest/issues/961) some context from that repo regarding Jest 24

Some minor changes to the jest config were necessary and for the most part things just worked.

I did find some issues with `async/await` tests, where I found a couple of problems, among them "regenerator runtime" not found. After searching for solutions I found a few issues that touched on this problem:

- https://github.com/facebook/jest/pull/7595
- https://github.com/facebook/jest/issues/6516
- https://github.com/facebook/jest/issues/7579
- https://github.com/vuejs/vue-cli/issues/3678
- https://github.com/facebook/jest/pull/7599

And some others.

The solution I came up with that fixed the issue was to add the babel plugin: `@babel/plugin-transform-runtime` but only to the 'test' environment to avoid including the runtime on our builds.

Other solutions that I read about was to update the babel configuration file, specifically the `preset-env.target.node` option to 'current', at the moment we have '8' and I assume we want to leave it like that, so I didn't try that (my node v is 8 anyways)

More info on babel and the runtime:
- https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator-aliasing
- https://babeljs.io/docs/en/v7-migration#remove-proposal-polyfills-in-babel-polyfill-https-githubcom-babel-babel-issues-8416

Watcher:

`npm run test`:


package.json
```diff
@@ -124,6 +124,7 @@
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
     "@babel/plugin-syntax-import-meta": "^7.0.0",
     "@babel/plugin-transform-flow-strip-types": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.4.3",
```
ItsJonQ: Hmm! I am suspicious about this one! I've had issues with this in the past. Checking out the compiled files now.. 🤞

ItsJonQ: @tinkertrain Ugh... yea. It's what I feared.

Check out the babel compiled code:

```
var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");

exports.__esModule = true;
exports.default = void 0;

var _extends2 = _interopRequireDefault(require("@babel/runtime/helpers/extends"));

var _objectWithoutPropertiesLoose2 = _interopRequireDefault(require("@babel/runtime/helpers/objectWithoutPropertiesLoose"));
```

Prior to `@babel/runtime`, these functions were not dependencies.
With this set up.. that means that consuming integrations HAVE to also use `@babel/runtime` to compile things.

I remember running into these issues during the Brigade v1.x work.

Is there a way we can get Jest 24 set up, without `@babel/runtime`? 

😊

ItsJonQ:
Dang it Netlify! What's going on

package.json
tinkertrain: I was afraid of something like that!

I have pushed a solution... feels a little bit like cheating...

ItsJonQ:
@tinkertrain Thanks for making that babel update! Will check again in a sec :)

package.json
ItsJonQ: It's not cheating if it works! Haha

babel.config.js
```diff
@@ -1,55 +1,35 @@
...
-          '@babel/react',
-        ],
-        plugins: [
-          '@babel/plugin-proposal-class-properties',
-          '@babel/plugin-transform-flow-strip-types',
-          'inline-svg',
-          'emotion',
-        ],
-      }
+module.exports = api => {
```
ItsJonQ: Much better 😎 . Take advantage of dat JS

tinkertrain:
[Upgrade HSDS React to Jest 24](https://trello.com/c/si3NkT9v/188-upgrade-hsds-react-to-jest-24)

ItsJonQ:
Oh darn. Netlify is struggling a bit again. Trying the build again (deleting cache + rebuild)

ItsJonQ:
Awesome! Thanks for the review all! And of course, big thanks to @tinkertrain 🚂 for making this upgrade 😍

Issue #595, opened 2019-04-24
data-cy selectors are wrong


ryan-mulrooney:
Recently cypress tests have been failing because the selectors we use have changed to something like `data-cy="propConnected(Heading)"` (usually they're `data-cy="greeter-new-button"`)

ItsJonQ:
@ryan-mulrooney Thanks a bunch!! Will take care of this today! 💪

ItsJonQ:
Making an update in Fancy:
https://github.com/helpscout/fancy/pull/45

Once that's merged. Will take care of that in HSDS: React. Then finally, in the app.

ItsJonQ:
Something is up with the `data-cy` passing during the propConnect phase.

Alternative fix. Sanitize the namespace before passing it through.

If a `prop['data-cy']` is defined... and if there are HOC wrappers (e.g. `connected()`), unwrap it to use the base namespace.

Rough code idea:
```
const matches = namespace.replace('/\(|\)/g', '|').split('|').filter(n => n)
const sanitizedNamespace = matches[matches.length - 1]
```

PR #597, opened 2019-04-26
Data Cy: Unwrap Namespaces

ItsJonQ:
## Data Cy: Unwrap Namespaces

This update fixes the `propConnect` data-cy attribute to ensure that any
"wrapping" namespaces, typically generated by HOCs, are stripped.

For example, `propConnected(Header)` becomes `Header`.

`connected(Text)` becomes `Text`.

Nested wrappers are also stripped, so `withRouter(connected(Text))` becomes
`Text`.

Before:

Screen Shot 2019-04-25 at 8 46 58 PM

After:

Screen Shot 2019-04-25 at 8 49 07 PM

Manually tested by loading the beta build into the app!

Fixes:

https://github.com/helpscout/hsds-react/issues/595


src/components/PropProvider/propConnect.tsx
```diff
@@ -108,10 +115,10 @@ function propConnect(name?: ConfigGetter, options: Object = {}) {
 
         return {
           'data-cy-component': namespace,
-          'data-cy': namespace,
           ...namespacedProps,
           ...rest,
           className,
+          'data-cy': this.getDataNamespace(),
```
ItsJonQ: Moved down to override `[data-cy]` that my exist in `...rest`. The `this.getDataNamespace()` method takes `this.props` into consideration.

PR #602, opened 2019-04-29
Enhancement/table updates

tinkertrain:
This PR fixes/updates/adds a couple of things that were found

## Updates

**Design**

From [Paper Doc](https://paper.dropbox.com/doc/Customer-List-QA-Design-feedback--AcIYhYgzvaa~IznWCyUxgUOqAg-f7vrUp0ApgNN3dJ336dIg)

- [x] Table rows should be 62px tall
- [x] First table row should have top left and right border radius of `4px`. Same for the last row, bottom left and right border radius.

## Fixes

**Styles**

After HS-App bullied the table with its global styles, I added `border-collapse: separate` to the styles to guard against that 🙅‍♂ .

**Scrolling**

Under the hood Table uses Scrollable, which in turn uses ScrollLock. I intended that by default Table should have the scroll locked in the "x" direction, as it stands it works fine in Firefox (ScrollLock doesn't work in firefox, it achieves it with pure css I added previously) but it wasn't working properly in Chrome and Safari 🤦‍♂ . In this PR I added a `scrollLockDirection` prop to the Scrollable component and in turn exposed `isScrolledLocked` for the Table, in case someone needs to disable the behaviour completely. Docs updated to reflect the additions.

**Cell Keys**
Michal found that adding 2 columns with the same "columnKey" would throw the 👻  feared "react children with the same key" ⚠️ warning in the console. I tweaked how keys are generated to make sure this use case is possible 💪 

Thanks @kierate for spotting and reporting a couple of these! 🙇


tinkertrain:
> code looks good 🚀
> 
> Playground in storybook doesn't seem to work right now though [deploy-preview-602--hsds-react.netlify.com/?path=/story/table--playground](https://deploy-preview-602--hsds-react.netlify.com/?path=/story/table--playground)

Good catch! I think I broke it in another commit... fixed!

ItsJonQ:
Will merge this one up to unblock! Can follow up with UI fixes (if needed)

Issue #571, opened 2019-04-09
Input: Update focus ring for error/warning states


ItsJonQ:
Updates designs:

https://github.com/helpscout/hsds-product-ui/wiki#forms

Raised by @digitaldesigner

PR #604, opened 2019-04-29
Input: Improve state focus ring colours

ItsJonQ:
## Input: Improve state focus ring colours

This update improves the `Input`/`Select` focus ring colors to support all
of our main UI states, including `error`, `success`, and `warning`.

The default `Input` story was enhanced for better UI testing.

#### Input
Screen Recording 2019-04-29 at 11 53 AM

#### Select
Screen Recording 2019-04-29 at 12 00 PM

Resolves: https://github.com/helpscout/hsds-react/issues/571


src/components/Input/styles/BackdropV2.css.ts
```diff
@@ -152,10 +151,7 @@ export const FocusUI = styled('div')`
     top: -1px;
   }
 
-  &.is-error {
-    box-shadow: 0 0 0 ${config.focusErrorOutlineWidth}px
-      ${config.focusErrorOutlineColor};
-  }
+  ${makeFocusStateStyles};
```
plbabin: just curious, emotion will execute the function even though there is no direct evocation here? )

ItsJonQ:
>  totally removing Flow from HSDS 💯
That's the plan! One file at a time...

PR #605, opened 2019-04-29
Dropdown: Multi-select: Match active item (without id)

ItsJonQ:
## Dropdown: Multi-select: Match active item (without id)

Screen Recording 2019-04-29 at 12 50 PM

This update fixes the Dropdown: Multi-select active rendering, specifically
for items without an `id` prop.

This update provides the matcher util with the `indexMap` id for both `id`
and `value` (as fallback) to determine the `is-active` UI state.


ItsJonQ:
Deploy preview:

https://deploy-preview-605--hsds-react.netlify.com/?path=/story/autodropdown--stateful-with-multiple-selection

PR #608, opened 2019-04-30
Avatar: Remove background when image is loaded

ItsJonQ:
## Avatar: Remove background when image is loaded

Screen Recording 2019-04-30 at 04 14 PM

This update adjusts the `Avatar` component to remove the background color
once the Image is loaded in. This prevents unexpected cropping issues
around the edge of rounded/circular avatars.


ItsJonQ:
@jaredmcdaniel Here's the deploy preview:

https://deploy-preview-608--hsds-react.netlify.com/?path=/story/avatar--v2-default

Click on the `image` checkbox on the right sidebar to toggle

PR #611, opened 2019-05-02
Accordion: Body Overflow Fix

ItsJonQ:
## Accordion: Body Overflow Fix

Screen Recording 2019-05-02 at 10 43 AM

This update fixes the render issues/inconsistencies with Accordion. The solution was to ensure that Accordion (and the underlying Collapsible) components both used `css.js` for styles. Also, that `overflow: hidden` was present for `Accordion` 

Lastly, the blue `outline` ring was removed from `Accordion` during the focus state.

Resolves: https://github.com/helpscout/hsds-react/issues/572


src/components/Accordion/Accordion.css.ts
```diff
@@ -35,6 +35,7 @@ export const BodyUI = styled('div')`
   ${baseStyles};
```
tinkertrain: In https://github.com/helpscout/hsds-react/pull/613, this file has been moved to its own `styles` folder... might want to do it here to avoid merge conflicts 👍

src/components/Collapsible/Collapsible.css.ts
```diff
@@ -0,0 +1,9 @@
+import styled from '../styled'
```
tinkertrain: ditto!

src/components/Accordion/Accordion.css.ts
ItsJonQ: @tinkertrain Will do x 2

ItsJonQ:
@tinkertrain Moved the `.css.ts` file to the `styles/` directory!

Issue #538, opened 2019-03-15
Add IconButton component


ItsJonQ:
To help resolve the scaffolding of

```
<Button>
  <Icon />
</Button>
```

PR #612, opened 2019-05-02
Add IconButton Component

ItsJonQ:
## Add IconButton Component

Screen Recording 2019-05-02 at 01 12 PM

This update adds a new `IconButton` component, which combines `Button` (v2) and `Icon` together. Under the hood, it still renders a `Button` so all of the `Button`'s properties and behaviours still work.

This update also refactors `Button` from JS -> TS, and updates Cyan to the latest v0.9.1 for improved testing!

Resolves: https://github.com/helpscout/hsds-react/issues/538


src/components/Button/Button.css.js
```diff
@@ -185,6 +185,10 @@ export const config = {
   focusOutlineWidth: 2,
```
tinkertrain: To avoid merge conflicts from https://github.com/helpscout/hsds-react/pull/613, can you move this file to it's own styles folder? ❤️

src/components/Button/Button.tsx
```diff
@@ -1,41 +1,40 @@
...
 import { namespaceComponent } from '../../utilities/component'
 import { classNames } from '../../utilities/classNames'
 import { noop } from '../../utilities/other'
 import RouteWrapper from '../RouteWrapper'
-import { COMPONENT_KEY } from './utils'
+import { COMPONENT_KEY } from './Button.utils'
+import { UIState } from '../../constants/types'
 
-type ButtonSelector = 'a' | 'button' | 'input'
+export type ButtonSelector = 'a' | 'button' | 'input'
```
tinkertrain: Can you move the types to the type folder? And while you are it, rename `Props` to `ButtonProps`. Thanks!

src/components/Button/ButtonV2.tsx
```diff
@@ -15,34 +12,39 @@ import {
...
-  isSuffix: boolean,
-  size: ButtonSize,
-  spinButtonOnLoading: boolean,
-  state?: UIState,
-  submit: boolean,
-  theme?: string,
+import { ButtonKind, ButtonShape, ButtonSize } from './Button.types'
+import { UIState } from '../../constants/types'
+
+export interface Props {
```
tinkertrain: Same as previous 🙇

ItsJonQ: Will do! 🙇‍♀

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+    )
+  }
+
+  getIconSize(): number {
+    const { iconSize, size } = this.props
+
+    if (size === 'xs') {
+      return 16
+    }
+    if (size === 'sm') {
```
plbabin: so we only support xs and sm sizes?

src/components/IconButton/__tests__/IconButton.test.tsx
```diff
@@ -0,0 +1,186 @@
+import * as React from 'react'
```
plbabin: it looks soooooo easy to test the component with cyan

src/components/IconButton/IconButton.tsx
plbabin: oh based on the story we can have other size. Maybe we need a default icon size, and overwrite it when it's `xs` and `sm` ?

ItsJonQ: @plbabin Yup! That's what's going on :). We have default `iconSize` of `24`, allowing the user to supply their own size. BUT... if the (Button) `size` is `sm` or `xs`, we force an iconSize.

plbabin: oh right default size is coming from the default prop value

I will blame it on ☕️not being strong enough 🤣

ItsJonQ:
@tinkertrain I moved the `IconButton` styles and `Button` types as you suggested for the Flow -> TS update!~ 🙌

Issue #496, opened 2019-02-25
Update `react-sortable-hoc`


ItsJonQ:
We'll need to update `react-sortable-hoc` to the lates version.

There is a collapse issue with `SidebarCollapsibleCard` though. The card doesn't collapse during the sort phase. It looks like the library changed it's rendering methods to clone the DOM node during sorting. It doesn't look like we can get React to re-render things when sorting happens.

Hopefully we can tap into the new `updateBeforeSortStart` callback to resolve this.

ItsJonQ:
Bump, for @plbabin

PR #613, opened 2019-05-06
Remove flow

tinkertrain:
## Remove flow. Replace with TypeScript

OK, brace yourselves this is kind of a big PR, practically touches every file in HSDS! 😅 

## Goal

Remove flow and prop-types from HSDS

## Achieved?

Yes

## That's right buddy boy

So what does this beast actually achieves?

Glad you asked!

## Removes flow and proptypes

From all components

## Folder and file naming conventions

All components now follow this structure:

```
hsds-react/
  └── src/
      └── components/
          └── Table/
              └── __tests__/
                  ├── Table.Cell.test.js
                  └── Table.test.js
              └── styles/
                  ├── Table.Cell.css.js
                  └── Table.css.js
              ├── index.ts
              ├── Table.tsx
              ├── Table.Cell.tsx
              ├── Table.utils.ts
              └── Table.types.ts
```

## Prop and state types

I moved most of the types out of the component files in their corresponding `Component.types.ts` file.

Some Components have a huge list of props so this make the Component file so much cleaner.

A lot of the types (prop and state types) I converted to the convention of:

```ts
export type TableProps = {
  /* ... */
}
export type TableState = {
  /* ... */
}
export type TableCellProps = {
  /* ... */
}
export type TableCellState = {
  /* ... */
}
```

## TypeScript

This PR also turned into converting the components to TypeScript, I fixed and converted a lot of TS issues, but not all!

I left the comments:

```
// TODO: fix typescript complains
// @ts-ignore
```

To signal stuff I didn't convert. The idea would be to gradually fix this, if necessary at all.

## .css.js vs .css.ts

You'll notice I left the css files as '.js'. As Q mentioned, we need to update TypeScript and other deps to achieve this, we'll leave for another task.

## Updated Contributing guide

Added the guidelines.

## Tests

🔬 Tests and Storybook updated and work as expected, coverage 100%

## Next steps

To have a 100% uniform codebase (which is ideal but not necessary) things to do:

🦔 Move all the types from the component file to their respective `.types.ts` file and import
🦔 Find a way to convert `.css.js` to `.css.ts`
🦔 Fix everywhere there is a `// TODO: fix typescript complains` (my count is ~85 instances 😅)

## Thanks

To @ItsJonQ for helping me taking this over the finish line! 🏁


__mocks__/popper.js.js
```diff
@@ -0,0 +1,12 @@
+import PopperJs from 'popper.js'
```
ItsJonQ: dat filename 😂 

Hopefully it doesn't cause any weird... build issues

docs/contributing/creating.md
```diff
@@ -159,18 +171,28 @@ Wonderful 🙏! You've created the base for `Strong`, that's performant, easy to
...
 We'll need to export `Strong` to make it simpler to import and use. This is all done in our `index.js` file:
 
 ```jsx
-// @flow
 import Strong from './Strong'
 import { propConnect } from '../PropProvider'
+import { COMPONENT_KEY } from './Strong.utils.ts'
 
-export default propConnect('Strong')(Strong)
+export default propConnect(COMPONENT_KEY)(Strong)
```
ItsJonQ: @tinkertrain Thank you! 🙏

ItsJonQ:
@tinkertrain You're a hero! 🙏 

Looking through the PR now!

> I spent a lot of time (and I mean a LOT) trying to fix this, babel settings, typescript settings, different dependencies... I gave up.

You should have asked me!! I could have provided context.

TLDR: Nesting components via `css.js` is only supported by Babel (via the Emotion plugin). Typescript does **not** use Babel. Therefore... the transformations can't happen.

P.S. TypeScript **can** go through Babel nowadays. In fact, Zero has this set up by default!
We don't have it set up in HSDS React (yet).

__mocks__/popper.js.js
tinkertrain: I know... has to be the name of the module 🙃

ItsJonQ:
🍐 'ed with @tinkertrain ! I think we're passed some of the trickier bits :).
Just need to make Jest happy with 100% test coverage, and Travis should be good!

ItsJonQ:
@tinkertrain As a follow up (not this PR!) we can do this:
https://github.com/helpscout/hsds-react/issues/616

:D!

ItsJonQ:
@tinkertrain Alrighty!! I fixed up `Pop` . Merged yours in, and fixed up test coverage! We should be all good now!

Next steps.. will make a beta release, and we'll test it out integration in Codesandbox :)

ItsJonQ:
@tinkertrain 

`v2.40.0-no-flow-0`

Here we go!

ItsJonQ:
@tinkertrain Cool! Looks like we have some problems :)

https://codesandbox.io/s/j7rq83zrjy

Good to know! Feel free to fork this and give things a try

ItsJonQ:
Recent updates also take care of https://github.com/helpscout/hsds-react/issues/496

ItsJonQ:
@tinkertrain Phew! I think with the latest changes, we should be okay!!

https://codesandbox.io/s/j7rq83zrjy

tinkertrain:
Pushed one more commit:

- Actually removing flow dependencies
- Doing that required to either remove types from `utils` or convert them to TS, did the latter

🌮 Tests, coverage and storybook remain working 🌮

ItsJonQ:
@plbabin Thanks for the review 🙌 !!

ItsJonQ:
Doin' it!!!

PR #615, opened 2019-05-07
Reduce size of the auto link regex

tompedals:
## Reduce size of the auto link regex

In an attempt to resolve some issues with the auto link regex not matching accented characters or non latin character sets we added in a really long regex string. This worked well but was causing problems with the size of the bundle. This change reduces the pattern to a simple `a-z0-9` so some URLs will no longer be matched.


ItsJonQ:
@tompedals Ruh roh! Looks like one of the tests broke 😅

ItsJonQ:
It's a 🚀 from me! @Charca if it's good with you, I can merge and (patch) release!

Issue #564, opened 2019-04-04
Page.Action: Add Sticky Feature


ItsJonQ:
Codepen prototype:
https://codepen.io/buzzusborne/live/29ae774172071600027ad54b97ee30ab

ItsJonQ:
Make it smart + awesome with `IntersectionObserver`
https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

We don't need to support IE for this, so we should be good!

ItsJonQ:
Coming soon... to a theatre near you:

![Screen Recording 2019-05-02 at 06 02 PM](https://user-images.githubusercontent.com/2322354/57110003-0ead1900-6d05-11e9-9148-8be94ff76565.gif)

Fully responsive, and is only sticky when it needs to be:

![Screen Recording 2019-05-02 at 06 05 PM](https://user-images.githubusercontent.com/2322354/57110015-15d42700-6d05-11e9-97c8-f6e0dbbe9dc5.gif)

PR #617, opened 2019-05-09
Page.Action: Add Sticky Feature

ItsJonQ:
## Page.Action: Add Sticky Feature

Screen Recording 2019-05-09 at 12 13 PM

This update adds a new sticky feature for the `Page.Action` component, which is typically used in Settings/Forms.

The stickiness is aware of the responsive setting from the (parent) `Page`. The interaction is accomplished by using the `IntersectionObserver`. This handy (native) JavaScript class works beautifully and is hyper performant. It does not support IE, but that's okay, because this component will only be used within the main application.

At the moment, the default setting is for sticky to be `false`. We can change it so sticky is on by default. I kept it on `false`, just in case.

cc'ing @digitaldesigner for thoughts on sticky by default, as well as UI/interaction review.

Resolves: https://github.com/helpscout/hsds-react/issues/564


config/jsdomPolyfills.js
```diff
@@ -1,40 +1 @@
-'use strict'
```
ItsJonQ: No longer needed. This all comes from Cyan 😎

src/components/Page/Page.Card.tsx
```diff
@@ -1,12 +1,12 @@
 import * as React from 'react'
 import getValidProps from '@helpscout/react-utils/dist/getValidProps'
+import propConnect from '../PropProvider/propConnect'
```
ItsJonQ: Some refactors to the `Page` components to use `propConnect`

src/components/Page/__tests__/Page.Actions.test.js
```diff
@@ -1,81 +1,222 @@
 import * as React from 'react'
-import { mount } from 'enzyme'
+import { cy } from '@helpscout/cyan'
```
ItsJonQ: Bye bye Enzyme. Hallo 🐱 Cyan! 🐱

digitaldesigner:
I think we should make it sticky by default, so all settings pages everywhere benefit from this update

ItsJonQ:
@digitaldesigner + @tinkertrain  

Deploy preview direct links:

https://deploy-preview-617--hsds-react.netlify.com/?path=/story/page-example--responsive

Or for the iFrame version

https://deploy-preview-617--hsds-react.netlify.com/iframe.html?path=/story/page-example--responsive

ItsJonQ:
@digitaldesigner I fine tuned the numbers! The transition should be smooth:

Screen Recording 2019-05-09 at 04 51 PM

(GIF demo'ing the transition if you scroll slow enough)

If you scroll super fast... than it might not be 100% smooth, cause of MATH + rendering reasons.

src/components/Page/Page.StickyActions.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  observerStart() {
+    const { offset } = this.props
+    const observerOptions = {
+      root: null,
+      rootMargin: `0px 0px ${offset * -1}px 0px`,
+      threshold: 1.0,
+    }
+    this.observer = new IntersectionObserver(
```
tinkertrain: This is so awesome, I have had to implement this manually in the past and it wasn't that much fun

src/components/Page/__tests__/Page.Actions.test.js
tinkertrain: 😍

ItsJonQ:
@tinkertrain Haii! Thanks for testing this stuff in Firefox! I guess Firefox has a bit to go to get IntersectionObserver working right 😱 .

I was able to fix the.. sometimes sticky issue you pointed out.

Re: the mobile issue, I'm not able to get it working correctly in Firefox.


Screen Recording 2019-05-13 at 06 02 AM

(In the GIF, you'll notice that the IntersectionObserver log never fires. More details in my git commit message below)

That's okay, since this component (or rather, the entire `Page` component) doesn't target mobile to begin with. At least, not yet.

From my git commit:

---

Sticky Actions: Fix implementation for Firefox

This update adjusts the sticky detection by using the `intersectionRatio` prop,
rather than `isIntersecting` prop. This is due to the differences between how
Firefox calculations the intersection value vs. Chrome/Safari.

Firefox also has a strange issue re: horizontal calculation. Once the element
is intersection horizontally (y axis), the IntersectionObserver no longer
detects vertical (x) based intersections.

Because of this, I've hidden the sticky version of the Page Actions element
for Firefox for smaller viewports. The value is determined by the Page's
min media query support.

tinkertrain:
@ItsJonQ Works great now! I think the mobile trade-off is great, TBH I don't particularly like sticky headers/footer on phones, as they just take real state screen...

🚀

ItsJonQ:
@tinkertrain Thank you!!!!

PR #622, opened 2019-05-14
TypeScript: ES Module Interop

ItsJonQ:
## TypeScript: ES Module Interop

This update adds the `esModuleInterop` option to the TS config. This ensures
that JS -> TS (then compiled back to JS) modules get imported/exported
correctly.

This is important for integrating the library with a consuming application.


ItsJonQ:
Hmm! Causing lots of compiling issues. Gonna close this

PR #628, opened 2019-05-16
Page.Action: Add guard for IntersectionObserver

ItsJonQ:
## Page.Action: Add guard for IntersectionObserver

This update adds a guard for `IntersectionObserver`, which is used for the
sticky feature in `Page.Action`.

The issue would only occur JSDOM. Target browsers (Edge+) supports this
feature

### DIFF
--- a/src/components/Page/Page.Actions.tsx
+++ b/src/components/Page/Page.Actions.tsx
@@ -72,7 +72,6 @@ export class Actions extends React.PureComponent<
       zIndex,
       ...rest
     } = this.props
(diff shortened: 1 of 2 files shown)


src/components/Page/Page.Actions.tsx
```diff
@@ -72,7 +72,6 @@ export class Actions extends React.PureComponent<
       zIndex,
       ...rest
     } = this.props
-    const { isStickyActive } = this.state
```
ItsJonQ: Unused variable

PR #629, opened 2019-05-16
Styled: Adding "Extras"

ItsJonQ:
## Styled: Adding "Extras" (aka. `_`)

With the [`@helpscout/fancy@2.6.0`](https://github.com/helpscout/fancy/releases/tag/v2.6.0) update, we're able to add extra arguments to the `styled` function. These "extras" are handy utilities used to style components within this library.

Extras is unique to this library, and isn't a default feature of a library like [Emotion](https://emotion.sh/docs/introduction).

### Example

**Without Extras**

```jsx
import styled from '../styled'
import { getColor } from '../../../styles/utilities/color'

const Brick = styled('div')`
  background-color: ${getColor('red.500')};
`
```

**With Extras**

```jsx
import styled from '../styled'

const Brick = styled('div')`
  background-color: ${styled._.getColor('red.500')};
`
```

**Or**

```jsx
import styled from '../styled'
const { _ } = styled

const Brick = styled('div')`
  background-color: ${_.getColor('red.500')};
`
```

**Or...**

```jsx
import styled from '../styled'

const Brick = styled('div')`
  ${({ _ }) => `
    background-color: ${_.getColor('red.500')};
  `};
`
```

Extras include the following utilities:

* `addFontSmoothing`
* `baseStyles`
* `darken`
* `forEach`
* `getColor`
* `getShadow`
* `isBeacon`
* `isHSApp`
* `lighten`
* `makeFontFamily`
* `makeFontFamilyMono`
* `resetStyles`
* `rgb`
* `rgba`
* `setFontSize`


ItsJonQ:
@knicklabs I was piggy backing off the vibe of Underscore/lodash, in that.. these are a collection of useful utils.

I can see how it may be confusing 😅 

Any other variables for namespacing you recommend? 😊

tinkertrain:
> @knicklabs I was piggy backing off the vibe of Underscore/lodash, in that.. these are a collection of useful utils.
> 
> I can see how it may be confusing 😅
> 
> Any other variables for namespacing you recommend? 😊

How about :bowtie:.getColor(), that seems pretty fancy to me...

:shipit:

ItsJonQ:
@tinkertrain I mean... you technically could use emojis... 😂

ItsJonQ:
For context, I livestream'ed the initial implementation of this.
Some other namespaces I thought of

`$`
`$$`
`_`
`__`

The `$` just looked weird when used with string interpolation.
`_` felt like a good fit.

Also, when you do `._.` or `{ _ }`, they look like faces. Nice side effect bonus.

I'm still open to changing the namespace!

knicklabs:
This is totally not a deal breaker for me, but what if instead of using a symbol, we used a short word that signalled intent, like `util`? But again, I don't feel that passionately about this so also feel free to ignore me and merge 😄

ItsJonQ:
Thanks all!!

PR #635, opened 2019-05-21
Improve table performance

plbabin:
## Improve table performance

This update brings some modifications on how the table is rendered on the page.

### Why
Right now the table could be rendered numerous time without having any props change.

### How
We introduce a couple of change on the structure of the table component. On the outside, nothing will change and the render UI will be the same.

##### Converting `PureComponent` to `Component`. Add a shouldComponentUpdate lifecycle method.
By doing this, we can do a deep-equal on the columns and the data, making sure that those two props really change before triggering a render. Since we are passing objects/array as props, PureComponent will re-render unless the object point to the same memory allocation, and as a generic component we can't force that.

##### Introducing a `Head` and `Body` sub-component
Those component wrap the `.map` function that render each cellHeader or each row. By creating a wrapper around those two loop, we can again add a `shouldComponentUpdate` and make sure the data or columns really change before triggering a render. That way, we can re-renderer the `Head` without touching the `Body`

##### Changing the `isLoading` UI state

Before, the isLoading state was fading out the whole table, triggering a browser redraw of each dom element. We added an overlay `<div>` on top of the table. That overlay will be toggle based on the `isLoading` prop value

##### Slicing the collapsed data is done in `Body`
Before, that slicing action was done in the `Table` render function causing a re-render on the children because slice will clone the array. Now, the slicing action is done in the `Body` state, an action that will be executed once when the component is initialize, or when the component will received a new list of items.

----

I hope those changes will slightly improve the table performance, especially the initial rendering time when there is a lot of state mutation from redux. If anybody has more or better suggestions on how to improve the rendering time (like with cell rendering, which could become a problem with multiple cell renderers + tons of rows), please reach out!


ItsJonQ:
@plbabin Resolved the merge conflicts 🙌

package.json
```diff
@@ -178,6 +179,7 @@
     "eslint-plugin-jsx-a11y": "5.1.1",
     "eslint-plugin-react": "7.9.1",
     "eslint-scope": "3.7.1",
+    "fast-deep-equal": "^2.0.1",
```
ItsJonQ: @plbabin Since it'll be used in `Table.Body`, we'll need to add this to `dependencies`, not `devDependencies` 😊

src/components/Table/Table.Body.tsx
```diff
@@ -0,0 +1,88 @@
...
+export default class Body extends React.Component<BodyProps, BodyState> {
+  columnsCache = []
+  state: BodyState = { rows: [] }
+
+  constructor(props) {
+    super(props)
+    this.columnsCache = this.getColumnsCache(this.props.columns)
+
+    this.state.rows = this.getRows(props)
+  }
```
ItsJonQ: @plbabin This can be written without `constructor`:

```jsx
export default class Body extends React.Component {
  state {
    rows: this.getRows(this.props)
  }
  columnsCache = this.getColumnsCache(this.props)

  ...
}
```

src/components/Table/Table.Body.tsx
```diff
@@ -0,0 +1,88 @@
...
+      let key = c.columnKey
+      if (Array.isArray(key)) {
+        key = key.join(',')
+      }
+      return key
+    })
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { rows } = this.props
```
ItsJonQ: @plbabin Since we're update state based on `rows`, would it be good to do a comparison between `this.props.rows` vs `nextProps.rows`?

src/components/Table/Table.Body.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return columns.map(c => {
+      let key = c.columnKey
+      if (Array.isArray(key)) {
+        key = key.join(',')
+      }
+      return key
+    })
+  }
+
+  componentWillReceiveProps(nextProps) {
```
tinkertrain: Is there a way to avoid using `componentWillReceiveProps` it's use is discouraged and will not work in future versions http://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

ItsJonQ: @tinkertrain I'm aware of the deprecating willReceiveProps as well. 🤔 

Thanks for the reminder! I'll be more aware the next time I feel like I need to use it :)

src/components/Table/Table.Body.tsx
plbabin: nice I didn't know about that pattern!

src/components/Table/Table.Body.tsx
plbabin: we could, but then we will run the deep comparison in both page (here when the props value changed, and then in shouldComponentUpdate). I wanted to reduce those `equal` that as much as possible

src/components/Table/Table.Body.tsx
plbabin: fixed in cfa50283

src/components/Table/Table.Body.tsx
plbabin: TIL I learned about getDerivedStateFromProps :)

I updated the component to use that lifecycle method in cfa50283

ItsJonQ: It's a React 16 thing :). So.. up until pretty recently, we wouldn't have been able to use it, heh

PR #643, opened 2019-06-12
Editable Input Fields

tinkertrain:
Editable Fields
===============

📚 https://deploy-preview-643--hsds-react.netlify.com/?path=/story/editablefield--text

A special type of inputs that allow you to edit _inline_, see stories to see exactly what that means.

## Features

- Types: 'text', with special handling for: 'email', 'number' and 'url'
- Clear (delete) value button
- Multi-value fields
- Add and remove values from multi-value fields
- Fields with options (think phone types: work, home, etc)
- Disabled state
- Actions: 'delete' and 'link' included, ability to pass custom actions
- Enter and Escape handling


src/components/EditableField/EditableField.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+export class EditableField extends React.PureComponent<
+  EditableFieldProps,
+  EditableFieldState
+> {
+  static className = 'c-EditableField'
+  static defaultProps = {
+    innerRef: noop,
+    initialValue: '',
+    type: 'text',
```
knicklabs: What do you think about using constants for values like this, like `inputTypes.TEXT` or something like that? I think its helpful to guard against typos but also can be helpful in testing to have object of constants that can be iterated over to produce inputs of all the different types.

src/components/EditableField/EditableField.tsx
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  getClassName() {
+    const { className } = this.props
+    return classNames(EditableField.className, className)
+  }
+
+  getNewValue = ({ inputValue, name }) => {
+    const { value } = this.state
+    const idx = Number(name.split('_')[1])
```
knicklabs: What does this do here?

src/components/EditableField/EditableField.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const { name, type } = this.props
+    const { value } = this.state
+
+    if (Array.isArray(value)) {
+      return (
+        <div>
+          {value.map((val, idx) => (
+            <EditableFieldInput
+              name={`${name}_${idx}`}
+              key={`${name}_${idx}`}
```
knicklabs: Would be nice to have a helper function for making these strings.

src/components/EditableField/EditableFieldInput.tsx
```diff
@@ -0,0 +1,151 @@
...
+          onChange={this.handleChange}
+          onFocus={this.handleFocus}
+          onKeyDown={this.handleKeyDown}
+        />
+        <FieldStaticValueUI
+          className="c-EditableField__staticValue"
+          innerRef={node => {
+            this.spanRef = node
+          }}
+          tabIndex={isEditing ? '-1' : '0'}
```
knicklabs: What do you think about creating constants for these values to improve readability?

tinkertrain:
Thanks for having a look @knicklabs ! As I go I will be refactoring some stuff taking into consideration your suggestions 👍

src/components/EditableField/EditableField.tsx
tinkertrain: I'll make it more readable, but in a nutshell, it determines the new value for the appropriate input in case of Fields with multiple values

src/components/EditableField/EditableField.types.ts
```diff
@@ -0,0 +1,101 @@
...
+    name: string
+    action: FieldAction
+    event: Event
+  }) => void
+}
+
+export interface TruncateProps {
+  string: string
+  splitter?: string
+}
```
plbabin: well done here, I trully love what you did with the Types!!

src/components/EditableField/EditableField.tsx
```diff
@@ -0,0 +1,477 @@
...
+      activeField: name,
+    })
+
+    onInputFocus({ name, value: fieldValue, event })
+  }
+
+  handleInputBlur = ({ name, event }) => {
+    const { onInputBlur } = this.props
+
+    return new Promise(resolve => {
```
plbabin: interesting strategy with the promise here. I like that!

plbabin:
Stellar work!

In my testing so far, I think I found 2 bugs:

In the multiple list, if you add a new line and commit it (pressing enter) **then remove the line**. If you add another new line and press ESC, the value of the removed line will be displayed

Screen Recording 2019-07-12 at 04 35 PM

Second bug _might be how it should behave though_:
Blurring out of a new item will not commit it, so when I focus back and press ESC, the line will be removed
Screen Recording 2019-07-12 at 04 38 PM

plbabin:
Did another round of testing this morning

1. There is a little jump with the multiple type options (when changing the dropdown value)
Screen Recording 2019-07-15 at 08 42 AM

2. You can tab/focus in a disabled multiple option field. Also I wonder if we should remove the hover state when disabled (and change the color of the text) 🤔 
Screen Recording 2019-07-15 at 08 44 AM

jaredmcdaniel:
Thanks @plbabin! Just want to make sure we're collecting all the feedback in the same spot for @tinkertrain. Can you compare that with or add it to the latest feedback in the Paper doc?

https://paper.dropbox.com/doc/Inline-Editor-feedback--Ag~QPZ0LnvuznwxemUj9_fkLAg-p9yy5KN76fNNVJDu3xHQ7

plbabin:
@jaredmcdaniel  will do!

Adding the link for future reference

https://paper.dropbox.com/doc/Inline-Editor-feedback--Ag9mF_70PQin8mzcDf5WyL9CAg-p9yy5KN76fNNVJDu3xHQ7

src/components/EditableField/EditableField.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  if (isArray(value)) {
+    fieldValue = value.map(val =>
+      createNewFieldValue({ value: val, name }, defaultOption)
+    )
+  } else {
+    fieldValue.push(createNewFieldValue({ value, name }, defaultOption))
+  }
+
+  return fieldValue
```
knicklabs: We could go an immutable route here which is a bit more succinct and means we don't need to build up the value of fieldValue over time.

E.g.

```
return isArray(value)
  ? value.map(value => createNewFieldValue({ value, name }, defaultOption)
  : [createNewFieldValue({ value, name }, defaultOption)]
```

Thoughts?

plbabin: I like that!

src/components/EditableField/EditableField.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+export function generateFieldActions(actions): FieldAction[] | [] {
+  let newActionsArray: FieldAction[] = []
+
+  if (actions !== null) {
+    if (actions === undefined) {
+      newActionsArray = [deleteAction]
+    } else {
+      newActionsArray = Array.isArray(actions) ? actions : [actions]
+
+      let isDeleteActionPresent =
```
knicklabs: Should this be a `const`?

src/components/EditableField/EditableField.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export function createNewValueFieldFactory(uuidFn) {
+  return function createNewValueFieldObject(
+    { value, name },
+    defaultOption?
+  ): FieldValue {
+    if (isObject(value)) {
+      const fieldObj = { ...value, id: uuidFn(`${name}_`) }
+
+      if (defaultOption != null && !Boolean(value.option)) {
```
knicklabs: Should we give `defaultOption` a default value of `null` and then use a strict equality check for `!== null` here?

src/components/EditableField/EditableField.utils.ts
```diff
@@ -0,0 +1,78 @@
...
+        newActionsArray.filter(action => action.name === 'delete').length > 0
+
+      newActionsArray = isDeleteActionPresent
+        ? newActionsArray
+        : newActionsArray.concat(deleteAction)
+    }
+  }
+
+  return newActionsArray
+}
```
knicklabs: I'm finding this method a bit difficult to follow because of the nested `if` statements. 

I think we could employ early returns to make it all easier to follow what the final value being returned is.

```
if (actions === null) {
  return []
}

If (actions === undefined) {
  return [deletedAction]
}

if (!Array.isArray(actions)) {
  return [actions]
}

const isDeleteActionPresent = actions.filter(({ name }) => name === 'delete').length > 0
return isDeleteActionPresent
  ? actions
  : actions.concat(deleteAction)
```

Thoughts?

knicklabs: As a follow-up question, is there a good use case for returning an empty set? I find it a bit unclear having different paths through code based on the difference between `null` and `undefined`. If there is a good case, could we add a comment?

src/components/EditableField/styles/EditableField.Input.css.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  .is-disabled &:hover {
+    cursor: initial;
+  }
+
+  .is-active & {
+    pointer-events: none;
+    z-index: 2;
+    border-bottom-color: transparent !important;
+  }
```
knicklabs: Just a question for you @tinkertrain, what do you think about using props inside the CSS instead of class names?

knicklabs:
@tinkertrain this is looking fantstic! There's a tremendous amount of great work here. I left a few questions and thoughts for you that I'd love your feedback on. I also updated the feedback document (multiple fields section) with some info about an issue I found regarding deleting entries via keyboard.

src/components/EditableField/EditableField.utils.ts
tinkertrain: Me too, thanks!

src/components/EditableField/EditableField.utils.ts
tinkertrain: yeah, I tend to use `something != null` because it checks for both `undefined` and `null`, strict comparison would just check `null`. In this case, I think making the default = `null` works, because I only care about `defaultOption` if it is defined. Thanks!

src/components/EditableField/EditableField.utils.ts
tinkertrain: In this case I do need different handling of `null` and `undefined`, because by default the "delete" action gets added (so if actions is `undefined`, you get "delete") and if you explicitly want to remove this behaviour, you can pass `null`. I'll put a comment there.

I agree that it can be structured in a better manner though. Thanks!

tinkertrain: One more thing, you can also override the delete action, so I need to check for that too

src/components/EditableField/styles/EditableField.Input.css.ts
tinkertrain: I think it depends on the situation, I like classes because:

- you can override the styles externally if you need to with minimal effort
- Is more familiar with more developers
- Styled components will create new classes for each variation, which I'm guessing with no real evidence that it might not be the most performant (kind of like @extend in sass? maybe I'm way off on this one though)

I like props because of their dynamism and cool things you can do, sometimes they are the right tool for the job, there's no easy way to achieve what line EditableField.Input.css.ts:215 does with classes, and I think that is awesome

knicklabs: Great points @tinkertrain! I guess with the way we've been using classes too, they help identify what state something is in, like `is-disabled`, for example so that's helpful for testing. Cheers!

plbabin:
Added two new comments on the feedback doc (13.b & 13.c). After that it's approved for me!

PR #647, opened 2019-06-19
Add time of day icons. Update fill color

plbabin:
This update adds two new elements to our icons set:

Image 2019-06-19 at 12 17 06 PM

It also changed a little bit how the currentColor is applied to all our icons. It verified that a fill attributes already exists on a path before enforcing "currentColor"

The update also set `fill="currentColor"` on some icons


ItsJonQ:
@plbabin Nice!!

Curious about the `path[fill]` target.

Would we want both?

`path, path[fill]`?

I originally added that rule to allow for the icon colours to seamless absorb things declared from a parent selector/node (e.g. a Button).

Just checking :)

plbabin:
reason is that sometime you need our css to leave alone a path that don't need a fill color (like the sun icon in this PR)

ItsJonQ:
@plbabin Gotcha! Thanks for context! :D

PR #652, opened 2019-06-26
New Filtered List component && StatusBadge update

plbabin:
This update add a new component called Filtered List. 

Screen Recording 2019-06-26 at 10 51 AM

A FilteredList component will displays a list of items based on a limit number. If the number of items goes over the limit, a badge will be displayed after the list.
Hovering on the badge will show all remainder items inside a tooltip

### Inline
Image 2019-06-26 at 10 51 06 AM

This component can be rendered inline. It can be done using the `inline` prop

```javascript
<FilteredList items={items} inline />
```

### Custom renderer

Image 2019-06-26 at 10 50 56 AM

There is also the possibility to use a custom renderer within the component to customize what will be displaying as the item (it default to a `<Text />` component. The `itemKey` prop will be use to determine if the value change or not during the component lifecycle.

```javascript
const customRender = item => {
    return <a href={item.href}>{item.label}</a>
}

const items = [
  {href="https://google.com/", label: "Google"}
]

<FilteredList items={items} renderItem={customRender} itemKey="label" />
```

## New badge UI
This update change the `StatusBadge` and the Badge isSquare UI. The border, background and text color slightly change to match the actual design for Customer List. This change will impact the conversation card in beacon

**old**
Image 2019-06-26 at 4 29 46 PM

**new**
Image 2019-06-26 at 4 31 28 PM


src/components/FilteredList/styles/FilteredList.css.js
```diff
@@ -0,0 +1,55 @@
...
+  align-items: center;
+  color: ${getColor('charcoal.400')};
+`
+
+export const BadgeUI = styled(Badge)`
+  &.is-square {
+    background-color: ${getColor('grey.300')};
+    box-shadow: 0 0 0 1px ${getColor('grey.500')} inset;
+    margin-left: 5px;
+  }
```
knicklabs: What do you think of moving this option to the `Badge` component? We need to use this badge variant in other places.

plbabin: 👍

plbabin:
I updated the PR with a new styling for the StatusBadge / Badge.isSquared

src/components/FilteredList/FilteredList.tsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  shouldComponentUpdate(nextProps) {
+    const { items } = this.props
+
+    if (nextProps.items.length !== items.length) {
+      return true
+    }
+
+    const itemKey = items
```
TerrenceLJones: Ohh, this is a niffy way of doing this check!

plbabin: Since it was part of the customer list table, we had to ensure this component render only when needed. Arrays/objects in props are not helpful for PureComponent so we went with a custom shouldComponentUpdate

PR #657, opened 2019-07-09
Refactor EmojiPicker active/hover state styling

TerrenceLJones:
This is a proposal for an alternative way of handling styles for Emoji Picker active/hover states. Previously, it was difficult to get the correct spacing for the background color for each emoji. This was inspired by how Github handles it's emoji picker for reactions. 

## Before
Screen Recording 2019-07-05 at 06 10 PM

## After
Screen Recording 2019-07-10 at 12 31 PM

### DIFF
--- a/src/components/EmojiPicker/styles/EmojiPicker.css.ts
+++ b/src/components/EmojiPicker/styles/EmojiPicker.css.ts
@@ -3,14 +3,6 @@ import Icon from '../../Icon'
 import { getColor, rgba } from '../../../styles/utilities/color'
 
 export const config = {
-  colors: {
-    grey: 'grey.300',
-    purple: 'purple.300',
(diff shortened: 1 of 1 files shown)


msjuliejulie:
@TerrenceLJones I think your solution will be a better fix in the short term. I would suggest maybe making the scale value a bit less, so the effect is a bit more subtle. Otherwise we would need to do resizing to all the icons to make sure they all have a consistent size within the bounds of the circle or add additional classes to target the icons with issues.

digitaldesigner:
The hover scale is causing the other emojis to bounce around which makes it look pretty strange. I would prefer for there to be no scale at all.

digitaldesigner:
Similarly if we could add +5px padding to the popover I think that'd make it feel a bit nicer

PR #663, opened 2019-07-17
Fixed error with Tooltips in Embed

Charca:
## Problem

When attempting to upgrade hsds-react in Beacon embed, I ran into an issue that was preventing the app from rendering.

I found out that the issue originated in version 2.43.1, which added an `EventListener` to the `Pop` component: https://github.com/helpscout/hsds-react/releases/tag/v2.43.1

## Solution

I'm not sure if this issue doesn't happen in hs-app, and if that's the case it might be because embed is still in React 15, but the issue seemed to be that "Providers" can only have a single child component, and with this update we were now returning two: `EventListener` and `PopUI`.

To fix this, I've wrapped both components in a `div`, which is the only React 15-compatible solution I know of, but if you know of a better one, please let me know!

### DIFF
--- a/src/components/Pop/Pop.tsx
+++ b/src/components/Pop/Pop.tsx
@@ -249,22 +249,24 @@ class Pop extends React.Component<Props, State> {
 
     return (
       <Manager>
-        <EventListener
-          event="click"
-          handler={this.handleOnBodyClick}
-          scope={document.body}
-        />
-        <PopUI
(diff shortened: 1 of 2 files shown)


tinkertrain:
Looks fine! :shipit:

Charca:
Thanks for fixing those, @TerrenceLJones!

PR #665, opened 2019-07-19
Editable Field Follow Up Fixes

tinkertrain:
This PR is a quick follow to the just merged Editable Fields.

- [x] Render as block: adding a new item would not automatically focus the new field
- [x] Hover over the actions would not render the border on the field
- [x] Refactors and fixes `calculateWidth` for Fields of dynamic width
- [x] Tweaked actions interactive area
- [x] Fixes interactions with keyboard on multi value fields: escape and enter had a couple of buggy behaviours on different scenarios

### Revision July 22nd

- [x] Increased the Field margin bottom from 20 to 40px
- [x] Blank slate text color changed to "charcoal.300"
- [x] Placeholder text color changed to "charcoal.300"
- [x] Blank slate border color changed to "grey.800"
- [x] Border bottom on blank slate is the width of the text instead of full-width
- [x] Tighten the spacing between the content and the borders by 3px
- [x] Static Value color changed to "charcoal.800"
- [x] Adjusted icon sizes and types
- [x] Pushed delete icon all the way to the right
- [x] On fields with options, hovering static state, now there is a gap between the option and the value (same as the space on active state)
- [x] Hovering the option shows the caret cursor instead of pointer

### Revision July 23rd

- [x] Fixed the blur effect when selecting an option a second or third time
- [x] Added the not-allowed cursor on disabled fields
- [x] Refactored class names handling
- [x] Made the width of the dashed border on ENTER or ESCAPE the width of the text instead of full width
- [x] Shielded input styles from other globally applied rules
- [x] Decreased space between add button and fields

### Revision July 24nd
- [x] Uniformed colours for various states that had slight discrepancies
- [x] Extracted colours into constants
- [x] Removed border on hover when fields are disabled

### DIFF
--- a/src/components/EditableField/EditableField.Input.tsx
+++ b/src/components/EditableField/EditableField.Input.tsx
@@ -25,8 +25,9 @@ import propConnect from '../PropProvider/propConnect'
 import getValidProps from '@helpscout/react-utils/dist/getValidProps'
 import {
   ACTION_ICONS,
-  COMPONENT_KEY,
+  EF_I_COMPONENT_KEY,
   isEllipsisActive,
+  getComponentClassNames,
 } from './EditableField.utils'
 import { classNames } from '../../utilities/classNames'
 import { key } from '../../constants/Keys'
@@ -38,11 +39,13 @@ import {
   EditableFieldInputState,
 } from './EditableField.types'
 
+const CLASSNAMES: any = getComponentClassNames(EF_I_COMPONENT_KEY)
+
 export class EditableFieldInput extends React.Component<
   EditableFieldInputProps,
   EditableFieldInputState
 > {
-  static className = 'c-EditableFieldInput'
+  static className = CLASSNAMES.component
   static defaultProps = {
     isActive: false,
     disabled: false,
@@ -169,11 +172,12 @@ export class EditableFieldInput extends React.Component<
     const staticValueNode = this.staticValueRef
 
     const contentNode =
-      staticValueNode && staticValueNode.querySelector('.c-Truncate__content')
+      staticValueNode && staticValueNode.querySelector(`.${CLASSNAMES.content}`)
     const emailNode =
-      staticValueNode && staticValueNode.querySelector('.TruncateFirstChunk')
+      staticValueNode &&
+      staticValueNode.querySelector(`.${CLASSNAMES.firstChunk}`)
 
-    /* istanbul ignore if */
+    /* istanbul ignore next */
     if (
       (contentNode && isEllipsisActive(contentNode)) ||
       (emailNode && isEllipsisActive(emailNode))
@@ -192,10 +196,10 @@ export class EditableFieldInput extends React.Component<
         0
     const staticContentNode = this.staticContentRef
     const staticOptionNode = staticContentNode.querySelector(
-      '.EditableField__staticOption'
+      `.${CLASSNAMES.staticOption}`
     )
     const staticValueNode = staticContentNode.querySelector(
-      '.EditableField__staticValue'
+      `.${CLASSNAMES.staticValue}`
     )
     const staticOptionWidth = staticOptionNode
       ? staticOptionNode.getBoundingClientRect().width + 10
@@ -214,7 +218,7 @@ export class EditableFieldInput extends React.Component<
     temporaryValueNode.textContent = staticValueNode.textContent
 
     // To get an accurate width, apply the font styles of the field
-    temporaryValueNode.classList.add('is-temporary-value')
+    temporaryValueNode.classList.add(CLASSNAMES.isTemporaryValue)
     editableFieldInputNode.appendChild(temporaryValueNode)
     contentWidth = Math.ceil(
       temporaryValueNode.getBoundingClientRect().width + staticOptionWidth
@@ -245,7 +249,7 @@ export class EditableFieldInput extends React.Component<
         : /* istanbul ignore next */
           0
       const placeholderNode = this.staticContentRef.querySelector(
-        '.is-placeholder'
+        `.${CLASSNAMES.isPlaceholder}`
       )
       const staticContentWidth = placeholderNode
         ? placeholderNode.getBoundingClientRect().width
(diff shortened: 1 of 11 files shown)


knicklabs:
@tinkertrain I added this to the Paper Doc too, but when a field is disabled, should we use the not-allowed cursor.

Issue

As seen above, it looks like you should be able to click into the input and I wonder if it will be obvious to users that the field is disabled.

Here, I've changed the cursor for illustration:

Possible Solution

src/components/EditableField/EditableField.Input.tsx
```diff
@@ -175,54 +184,86 @@ export class EditableFieldInput extends React.Component<
...
+      '.EditableField__staticValue'
+    )
+    const staticOptionWidth = staticOptionNode
+      ? staticOptionNode.getBoundingClientRect().width + 10
+      : 0
+    let contentWidth = 0
+
+    // A reliable way to get the width of dynamic content
+    // is by creating a temporary element
+    // and get its width
```
knicklabs: Nice comment!

src/components/EditableField/EditableField.tsx
```diff
@@ -414,7 +427,26 @@ export class EditableField extends React.Component<
     }
   }
 
-  renderInputFields() {
+  renderAddButton = () => {
+    const { disabled } = this.props
+    const { fieldValue, multipleValuesEnabled } = this.state
+
+    const isLastValueEmpty = fieldValue[fieldValue.length - 1].value === ''
+    const isSingleAndEmpty = fieldValue.length === 1 && isLastValueEmpty
```
knicklabs: I believe `isSingleAndEmpty` is redundant because `isLastValueEmpty` will return `true` if the array has one item and is empty AND it will return true if it has n items and the last one is empty.

src/components/EditableField/EditableField.tsx
```diff
@@ -414,7 +427,26 @@ export class EditableField extends React.Component<
...
+
+    return multipleValuesEnabled && !isSingleAndEmpty && !disabled ? (
+      <AddButtonUI
+        className="EditableField_addButton"
+        type="button"
+        onClick={this.handleAddValue}
+        disabled={isLastValueEmpty}
+      >
+        <Icon name={ACTION_ICONS['plus']} size="20" />
```
knicklabs: You can just use `ACTIONS_ICONS.plus` here.

src/components/EditableField/EditableField.utils.ts
```diff
@@ -5,9 +5,9 @@ import { find } from '../../utilities/arrays'
 export const COMPONENT_KEY = 'EditableField'
 
 export const ACTION_ICONS = {
-  delete: 'cross-medium',
-  link: 'new-window-large',
-  plus: 'plus-medium',
+  delete: 'cross-small',
+  link: 'new-window',
+  plus: 'plus-small',
   valueOption: 'chevron-down',
```
knicklabs: There are constants for these defined in `icons-all.js` that are exported. Could we reuse these?

For example:

```
export const ACTION_ICONS = {
  delete: ICONS.crossMedium,
  ...
}

jaredmcdaniel:
@knicklabs nice call on disabled. My initial thought was to just remove the dashed border bottom. Think that would help clarify? If you still prefer the disabled cursor, lets chat about it. Felt a bit heavy to me on first pass ... but might also be the clearest option. 👍 Cc @tinkertrain

knicklabs:
@jaredmcdaniel I can see that the not-allowed cursor is a tad heavy handed. I think my issue just stems from how we get a text cursor, so even though the input looks disabled, the text cursor suggests I can interact with it by insertion or selection. What if we just set the cursor to "default" here instead, so that it hover over it is no different than hovering over any other non-input element?

jaredmcdaniel:
@knicklabs love it. Much better. Cc @tinkertrain.

src/components/EditableField/EditableField.utils.ts
tinkertrain: I think `ICONS` is an an object with keys that tie to the svg:

`ICONS.folder = "<svg ... />"`

src/components/EditableField/EditableField.tsx
tinkertrain: Ah yeah, thanks I removed the suffix and now I can change that 👍

src/components/EditableField/EditableField.tsx
tinkertrain: The reason I have both, is because I use one to decide to render, and the other to decide if it should disabled, I know it looks a little bit weird... 😅

src/components/EditableField/EditableField.utils.ts
knicklabs: I was hoping we could use this https://github.com/helpscout/hsds-react/blob/master/src/components/Icon/icons.all.js#L212 but, nevermind, it turns out the keys and values are in the reverse order that we'd need.

src/components/EditableField/EditableField.tsx
knicklabs: Right, I see the minutia there. Thanks!

jaredmcdaniel:
@tinkertrain couple things ...

1) On disabled ... any reason we're keeping the border-on-hover? Seems odd that there's a hover state but its un-editable.
2) Can we confirm that _all_ disabled text colors are the same? It looks like the label on `with options` is a lighter hex:

image

3) I'm still not seeing the caret on hover for the `with options` label. We still waiting on that?

Thanks man!

tinkertrain:
@jaredmcdaniel 

1. I removed the hover state
2. Same colours now (found a couple of other instances where there were some discrepancies)
3. Not sure why you are not seeing it? Here's what I see on both chrome (recorded) and firefox: http://c.hlp.sc/a40a7fb24d13/Screen%20Recording%202019-07-24%20at%2011.22%20AM.mov (in that movie, beacuse of the close up you can see the dropdown option was pure black, that is one of the colours I found and changed)

Cheers!

PR #673, opened 2019-08-01
Fix remaining pill issues

ryan-mulrooney:
After @tjbo expertly created and implemented the pill for character counting on Form Inputs, we noticed a few issues to address in this follow up.

1. When the pill appears, all the content below is pushed down: http://c.hlp.sc/281f55b675ca
2. When the pill first appears, there's a staggered effect: http://c.hlp.sc/04b2ad07a850
3. When reducing the character count to less than 250, the new count is briefly shown before the pill disappears. The pill should just disappear without displaying the new count: http://c.hlp.sc/281f55b675ca

To address 1) @ryan-mulrooney moved the pill `div` inside the InputUI component, and then `position: absolute`
2) and 3) is @tjbo's bag :)

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -581,7 +581,7 @@ export class Input extends React.PureComponent<InputProps, InputState> {
         >
           <Badge
             status={isTooMuch ? 'error' : 'success'}
-            style={{ fontWeight: 100 }}
+            style={{ fontWeight: 100, minWidth: 75 }}
           >
             {this.state.charValidatorText}
           </Badge>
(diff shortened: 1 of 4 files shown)


tjbo:
Fixed #3.

#2 is due to not having a min-width, I think we will need to add a min-width because since we are also not using a MonoSpaced font, this issue can rear it's head in other ways, for example when character width changes:

A) http://c.hlp.sc/e6b916f3eb5c

B) if we have a different string length ie, `99 / 100` has a completely different length than `100/100`

@digitaldesigner , How do you feel about a min-width to guard against these situations?

digitaldesigner:
I'm fine with a min-width provided the contents are centered

PR #681, opened 2019-08-13
Composite editable fields

tinkertrain:
# Composite Editable Fields

[Spec](https://paper.dropbox.com/doc/Customer-Profile-Edit-Name--Ai2vQ2HZ__RCuMNclAn46nVdAg-oqhqXYSAGmhckhERnSWuN)

This PR adds a new component under `EditableField`: `EditableFieldComposite`

### `EditableFieldComposite`

This component allows the grouping and in-lining of multiple single value Editable Fields.

`EditableFieldComposite` is in essence a Higher Order Component that decorates its children to enable grouping them.

### API

```jsx
	<EditableFieldComposite placeholder="Add a name" size="lg">
      <EditableField
        label="First Name"
        name="first_name"
        type="text"
        placeholder="First Name"
        value="Johnny"
        onInputFocus={() => {
          console.log('execute!')
        }}
      />
      <EditableField
        label="Last Name"
        name="last_name"
        type="text"
        placeholder="Last Name"
        value="Cash"
      />
    </EditableFieldComposite>
```

It only has a few props:

- placeholder: this is the text you see when all fields are empty
- size: default is "md", pass "lg" for the large option
- separator: By default, each value will be separated with a non-breaking space, but you can pass any string here, like ","

Use each `EditableField` as normal, including events

`EditableFieldComposite`:

- Decorates each field with an `inline` prop
- Removes labels
- Adds "floating labels" to fields http://c.hlp.sc/c4aa5753474f/floating_3.mp4

### Large size

This PR enables rendering regular `EditableField` in large size, pass a prop of size="lg" if you need it, for Composite there is no need to put this prop in each field, only on the Composite (see example above).


jaredmcdaniel:
@tinkertrain nice to see the pieces start to come together!

Know we've got plenty of styling to go yet, but the only thing Im seeing at this point is a slight discrepancy between the border-bottom and focus state for the `Add a pet` blank slate. Looks like there's a `px` diff or so?

src/components/EditableField/EditableField.tsx
```diff
@@ -124,11 +126,13 @@ export class EditableField extends React.PureComponent<
...
   getClassName() {
-    const { className, disabled } = this.props
+    const { className, disabled, size } = this.props
+
     return classNames(
       EditableField.className,
       className,
-      disabled && STATES_CLASSNAMES.isDisabled
+      disabled && STATES_CLASSNAMES.isDisabled,
+      size === 'lg' && STATES_CLASSNAMES.isLarge
```
knicklabs: Should we make constants for these sizes?

src/components/EditableField/EditableFieldComposite.tsx
```diff
@@ -0,0 +1,301 @@
...
+       *
+       * When do we hide the mask? On focus, always
+       * When do we show the mask? Can't be always on blur, because although it would technically work,
+       * we would get a flash of the mask when we move the focus from "name" to "city"
+       *
+       * We know that moving the focus from one input (name) to another (city) triggers this sequence:
+       * blur name, focus city
+       *
+       * Editable Field already takes care of adding a class when it's active, we just need to wait a bit
+       * so that the focus event gets triggered right after the blur, and that is when we act
```
knicklabs: Fantastic comment 👏

jaredmcdaniel:
Feeling really nice @tinkertrain! Couple things ...

1) The animation on de-focus feels a bit jumpy and I think it's due to the two lines closing ... and they're not really aligned with anything because the first and last name transition back together (vs 50% width). Would it be possible to either **a)** remove the last name border animation on de-focus or **b)** remove the border animation for _both_ first and last name? I think I prefer **a** but let me know your thoughts.



2) I know we'll have wider real estate in the modal and full profile, but just raising this:

tinkertrain:
> Really nice work! I like that you have it focus the input based on segment was clicked.

Thanks! I'm so glad someone noticed that little nugget :)

tinkertrain:
> Feeling really nice @tinkertrain! Couple things ...
> 
>     1. The animation on de-focus feels a bit jumpy and I think it's due to the two lines closing ... and they're not really aligned with anything because the first and last name transition back together (vs 50% width). Would it be possible to either **a)** remove the last name border animation on de-focus or **b)** remove the border animation for _both_ first and last name? I think I prefer **a** but let me know your thoughts.
> 
> 
> 
> 
>     1. I know we'll have wider real estate in the modal and full profile, but just raising this:
> 
> 
> 

I just made an update addressing both comments :)

src/components/EditableField/EditableFieldComposite.tsx
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    return (
+      <span
+        className={classNames(
+          COMPOSITE_CLASSNAMES.maskItem,
+          STATES_CLASSNAMES.isPlaceholder
+        )}
+        onClick={() => {
+          this.handleMaskClick('placeholder')
```
plbabin: having the onClick created like that, would that not declare a lot of unnecessary functions in memory each time masks are re-rendered? (same with line 243).

I get that we need to pass custom arguments based on which mask was clicked. Not sure I have a solution for that too...maybe caching each callback in an object could help 🤷‍♂

src/components/EditableField/styles/EditableField.Input.css.ts
```diff
@@ -34,51 +37,95 @@ export const ComponentUI = styled('div')`
...
+    will-change: transform, font-size;
+    transition-property: transform, font-size;
+    transition-timing-function: linear;
+    transition-duration: 0.2s;
+    pointer-events: none;
+  }
+
+  .${STATES_CLASSNAMES.hasActiveFields} & {
+    &::before {
+      content: ${({ placeholder }) => `"${placeholder}"`};
```
plbabin: I really like that solution to pass the custom placeholder. Another good example of how great styled component could be!

src/components/EditableField/EditableFieldComposite.tsx
tinkertrain: Hi PL, welcome back!

This subject of inline functions in React is a bit touchy, there is a bit of a dogma that you should just NEVER do it. But most of the time this is said without any evidence. The reason people give is that react has to create a new function every time `render` runs and thus it is BAD for performance.

I believe in staying away from premature micro optimizations. Creating and destroying functions in JavaScript is _fast_, arrow functions are native and/or transpiled with babel to normal functions. Before, the pattern was to use `bind`, and that _was_ slow (although I think JS engines have make it less so).

In any case, if we find a problem and after measuring and finding this to be the culprit, we address it at that point. Right now, it will require some kind of convoluted code just to avoid it, and don't think is worth investing the time.

I like this article by Ryan Florence: https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578, which puts more eloquently what my thoughts are.

In searching for the article, is amazing the number of articles that start with "ANTIPATTERN", "NEVER DO THIS", hahahaha  ¯\_(ツ)_/¯

It's definitely an interesting subject and we can talk more about it!

plbabin: thank you for your input, I will read the article for sure. I agree that micro optimizing that kind of stuff is trivial and that we should do it only if it's required

PR #685, opened 2019-08-21
Add Avatar action overlay

plbabin:
This update adds a new overlay when hovering an avatar.

Screen Recording 2019-08-20 at 04 53 PM

This overlay is fully customizable and will follow the shape of the avatar (circle, rounded or square).

The Avatar dom element will be a button when it is `actionable`. This will add a focus state to the avatar

Screen Recording 2019-08-21 at 03 23 PM

It adds 4 new props to the `<Avatar />` component:

* **actionable** Activate the action overlay
* **actionIcon** Change the actual icon of the overlay (default to the `trash`icon)
* **actionIconSize** Set the size of the overlay icon
* **onActionClick** Callback for then the overlay was clicked

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -18,6 +18,8 @@ import {
   StatusUI,
   TitleUI,
   ActionUI,
+  AvatarButtonUI,
+  FocusUI,
 } from './styles/Avatar.css'
 import {
   COMPONENT_KEY,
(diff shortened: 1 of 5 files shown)


jaredmcdaniel:
Mentioned to @plbabin in Slack, but just looking for a focus state on click. Otherwise looking great. Love that the overlay is consistent with the desired shape 👍

plbabin:
I added a focus state as requested by @jaredmcdaniel. I had to change the dom element used when the avatar is actionable to make it focusable.

PR #690, opened 2019-08-26
Editable Fields validation

tinkertrain:
WIP


knicklabs:
I made a beta release of this for testing. `npm i -S @helpscout/hsds-react@editable-fields` if you want to test it in a project.

knicklabs:
@tinkertrain Great news: the editable fields update when they receive new props. Bad news is that the composite editable fields do not.

tinkertrain:
> @tinkertrain Great news: the editable fields update when they receive new props. Bad news is that the composite editable fields do not.

Thanks for testing! I pushed a fix :)

tinkertrain:
@knicklabs I have added the new data for onCommit on this PR as I'd like you to have more time take it for a spin and make sure it's ok before I leave for vacations. I added a story and updated the description.

knicklabs:
I think this is looking good and I've been using a beta build of it and its feeling really good. I'd be happy to see this merged up. @jaredmcdaniel I know you mentioned you may have some design feedback. Did you want to address it in this PR or merge and do a follow-up a PR?

src/components/EditableField/EditableField.types.ts
```diff
@@ -74,21 +94,27 @@ export interface EditableFieldProps {
...
   onAdd: (args: { name: string; value: FieldValue[] }) => void
-  onCommit: (args: { name: string; value: FieldValue[] }) => void
+  onCommit: (args: {
+    name: string
+    value: FieldValue[]
+    data: CommitData
+  }) => void
   onDelete: (args: { name: string; value: FieldValue[]; event: Event }) => void
   onDiscard: (args: { value: FieldValue[] }) => void
+  validate: (args: { value: string; name: string }) => Promise<Validation>
```
knicklabs: @plbabin I'm having some difficult integrating this for network errors. I'm thinking we should either accept validationInfo as props to override the internal state so that we can control the component or removing the validate function in favor of passing in the props, only. What do you think?

PR #691, opened 2019-08-27
Fix Avatar background color & Add remove avatar animation

plbabin:
This update fixes an issue with the avatar button background color. In #685 we add replace the `<div>` with a `<button>` when the `Avatar` component is actionable. By doing that the default button background was still present. Now it will be a transparent background. 
 
**Before**
Image 2019-08-27 at 3 07 12 PM

**After**
Image 2019-08-27 at 3 06 38 PM

### Remove avatar animation
This update add an animation sequence to remove an avatar. It will add a border animation, force load the fallback image and fade out the actual avatar. When the animation is completed, the avatar will not be clickable anymore

Screen Recording 2019-08-29 at 03 44 PM

Right now this animation is only valid for the circle shape.

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -20,6 +20,8 @@ import {
   ActionUI,
   AvatarButtonUI,
   FocusUI,
+  BorderAnimationUI,
+  CircleAnimationUI,
 } from './styles/Avatar.css'
 import {
   COMPONENT_KEY,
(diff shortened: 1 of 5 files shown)


plbabin:
I pushed another update with the full remove avatar animation. This last commit completed what was needed for the Avatar

plbabin:
@tinkertrain I fixed the issue on firefox. It seems that FF does not support modifying the svg circle attributes from CSS, I create a helper that would allow us to set the attribute directly on the svg element

I'll go ahead with merging the PR

PR #695, opened 2019-09-05
Upgrade to colorway 0.3.0 to update all yellow.

plbabin:
Image 2019-09-05 at 11 15 05 AM

This update upgrade colorway to v0.3.0. This version contain the latest yellow color palette.

**noote** 
At some point we'll have to update seed to a newest version to get the latest _SCSS_ colors. It was not done for the blue update so I am leaning toward waiting that all colors were migrate before upgrading seed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1682,13 +1682,15 @@
       }
     },
     "@helpscout/colorway": {
-      "version": "0.0.11",
-      "resolved": "https://registry.npmjs.org/@helpscout/colorway/-/colorway-0.0.11.tgz",
-      "integrity": "sha512-0OutxeVLTMjDEMw9lQFyzUJN+v000FxI7obVdvKA6w77iEJLBZYpyiiPzcsosCUBxL8A3A/eBWO1KVnP3SR/qw==",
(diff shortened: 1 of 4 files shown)


plbabin:
Pretty much only the button V1 that the color are not updating (and are using the old Seed colors). I think it's minor since we are using V2 a lot, and it should probably be the app that is using V1 that should do the color update (it was like that with the `blue` so I will follow the same path)

digitaldesigner:
Everything looks the same to me...

plbabin:
@digitaldesigner Deploy preview is broken for whatever reason

I took a screenshot of prod and local storybook:

Image 2019-09-06 at 9 33 51 AM
Image 2019-09-06 at 9 34 03 AM

digitaldesigner:
I'm happy if you are, all I'd ask is you do QA on the Notes editor — that's the most critical component that uses yellow heavily

plbabin:
> I'm happy if you are, all I'd ask is you do QA on the Notes editor — that's the most critical component that uses yellow heavily

are you talking about the chat editor and the convo editor in hs-app?

PR #701, opened 2019-09-13
Add keypress event callbacks to EditableField

knicklabs:
We need a way to be able to add event listeners to EditableField inputs on key events. This update adds `onInputKeyDown`, `onInputKeyPress`, and `onInputKeyUp` event listeners.

### DIFF
--- a/src/components/EditableField/EditableField.Input.tsx
+++ b/src/components/EditableField/EditableField.Input.tsx
@@ -29,7 +29,6 @@ import {
   OTHERCOMPONENTS_CLASSNAMES,
   TRUNCATED_CLASSNAMES,
   STATES_CLASSNAMES,
-  COLOURS,
 } from './EditableField.utils'
 import { classNames } from '../../utilities/classNames'
 import { key } from '../../constants/Keys'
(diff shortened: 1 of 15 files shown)


knicklabs:
Not sure why this build is failing. Will look into it.

knicklabs:
Build failure is all sorted out.

src/components/EditableField/EditableField.tsx
```diff
@@ -252,7 +256,7 @@ export class EditableField extends React.Component<
             value: fieldValue,
             data: {
               cause: 'BLUR',
-              operation: 'UPDATE',
+              operation: OPERATION.UPDATE,
```
plbabin: yay!!

src/components/EditableField/EditableField.Input.tsx
```diff
@@ -175,15 +176,19 @@ export class EditableFieldInput extends React.Component<InputProps> {
   /* istanbul ignore next */
   handleKeyDown = event => {
     const isEnter = event.key === key.ENTER
-    const isEscape = event.key === key.ESCAPE
     const isDropdownTrigger = event.target.classList.contains(
       OTHERCOMPONENTS_CLASSNAMES.dropdownTrigger
     )
```
plbabin: I am curious on why the Escape part was removed?

knicklabs: Before this input was only allowing `enter` and `escape` key events to bubble up to the parent. In the case of enter, it was only bubbling up when no in a dropdown trigger. With the exception of the dropdown case, we now want all key events to bubble up.

PR #703, opened 2019-09-16
Add "splitter" prop to <Truncate />

cen10:
The splitter prop allows text before the splitter to be truncated, which shortens long usernames on emails and keeps the entire domain name visible (e.g. `averylongem...@gmail.com`).

It's based off [EditableField.Truncated](https://github.com/helpscout/hsds-react/blob/master/src/components/EditableField/EditableField.Truncated.tsx). The EditableField code was not imported intentionally since changes to EditableField can reasonably be expected only to affect EditableField and people may not know or may forget if other components rely on it.

For some reason when I drop a gif in it's not animating, but here's a [link](https://p-zkF42X.t2.n0.cdn.getcloudapp.com/items/X6uY1LWL/Screen+Recording+2019-09-16+at+04.10+PM.gif?v=7ba155b717aacd42c29d02168dd3722e).
 
This also edits `Truncate.stories` because in Storybook the example for `Auto` wasn't being truncated -- but if I'm misunderstanding how this example is supposed to work, let me know. When I go to the Truncate default story, the text for the `Auto` example runs off the page unless I provide a width on the `<p>` tag. [This is what I see on master](https://share.getcloudapp.com/KouLX5vA) in Storybook, and you can see what it looks like with my edit in the gif I link to above.

### DIFF
--- a/src/components/Truncate/Truncate.tsx
+++ b/src/components/Truncate/Truncate.tsx
@@ -6,6 +6,8 @@ import styled from '../styled'
 import { truncateMiddle } from '../../utilities/strings'
 import css from './styles/Truncate.css'
 import { TruncateProps, TruncateState } from './Truncate.types'
+import { TruncateWithSplitterUI } from './styles/Truncate.WithSplitter.css'
+import { TRUNCATED_CLASSNAMES } from './Truncate.utils'
 
 export class Truncate extends React.PureComponent<
   TruncateProps,
@@ -101,6 +103,7 @@ export class Truncate extends React.PureComponent<
       ellipsis,
       limit,
       showTooltipOnTruncate,
+      splitter,
       tooltipPlacement,
       tooltipProps,
       tooltipModifiers,
@@ -117,14 +120,37 @@ export class Truncate extends React.PureComponent<
     )
 
     const shouldShowTooltip = showTooltipOnTruncate && this.state.isTruncated
-    const word = this.getTruncatedContent()
+    let truncatedText
 
-    const wordMarkup = (
+    if (splitter) {
+      const str = text || children
+      const [first, second] = str.split(splitter)
+
+      truncatedText = (
+        <TruncateWithSplitterUI
+          className={`${TRUNCATED_CLASSNAMES.component} ${
+            TRUNCATED_CLASSNAMES.withSplitter
+          }`}
+        >
+          <span className={`${TRUNCATED_CLASSNAMES.firstChunk}`}>{first}</span>
+          <span className={`${TRUNCATED_CLASSNAMES.splitterChunk}`}>
+            {splitter}
+          </span>
+          <span className={`${TRUNCATED_CLASSNAMES.secondChunk}`}>
+            {second}
+          </span>
+        </TruncateWithSplitterUI>
+      )
+    } else {
+      truncatedText = this.getTruncatedContent()
+    }
+
+    const textMarkup = (
       <span
         className="c-Truncate__content"
         ref={(ref: any) => (this.contentNode = ref)}
       >
-        {word}
+        {truncatedText}
       </span>
     )
     const content = shouldShowTooltip ? (
@@ -134,10 +160,10 @@ export class Truncate extends React.PureComponent<
         placement={tooltipPlacement}
         title={title || this.getText()}
       >
-        {wordMarkup}
+        {textMarkup}
       </Tooltip>
     ) : (
-      wordMarkup
+      textMarkup
     )
 
     return (
--- a/src/components/Truncate/Truncate.utils.ts
+++ b/src/components/Truncate/Truncate.utils.ts
@@ -1,0 +1,9 @@
+export const TRUNCATED_COMPONENT_KEY = 'Truncated'
+
+export const TRUNCATED_CLASSNAMES = {
+  component: TRUNCATED_COMPONENT_KEY,
+  withSplitter: 'withSplitter',
+  firstChunk: `${TRUNCATED_COMPONENT_KEY}__firstChunk`,
+  splitterChunk: `${TRUNCATED_COMPONENT_KEY}__splitterChunk`,
+  secondChunk: `${TRUNCATED_COMPONENT_KEY}__secondChunk`,
+}
--- a/src/components/Truncate/Truncate.utils.ts
+++ b/src/components/Truncate/Truncate.utils.ts
@@ -1,9 +1,9 @@
-export const TRUNCATED_COMPONENT_KEY = 'Truncated'
+export const TRUNCATED_COMPONENT_KEY = 'c-Truncate__content__chunks'
 
 export const TRUNCATED_CLASSNAMES = {
   component: TRUNCATED_COMPONENT_KEY,
-  withSplitter: 'withSplitter',
-  firstChunk: `${TRUNCATED_COMPONENT_KEY}__firstChunk`,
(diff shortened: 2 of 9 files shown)


src/components/Truncate/Truncate.tsx
```diff
@@ -117,14 +120,37 @@ export class Truncate extends React.PureComponent<
     )
 
     const shouldShowTooltip = showTooltipOnTruncate && this.state.isTruncated
-    const word = this.getTruncatedContent()
```
cen10: Changing this variable name from `word` to `truncatedText` since it may not always be a word that needs truncating and, in the case of email addresses where I'll be using this feature, I think `text` makes more sense. 

Naming it `truncatedText` is a little bit of a misnomer because there are cases when the text will come back unchanged - for example, if the limit is 15 characters and the text is less than that - but I think it helps show that the text is probably being manipulated.

src/components/Truncate/Truncate.utils.ts
```diff
@@ -0,0 +1,9 @@
+export const TRUNCATED_COMPONENT_KEY = 'Truncated'
+
+export const TRUNCATED_CLASSNAMES = {
+  component: TRUNCATED_COMPONENT_KEY,
+  withSplitter: 'withSplitter',
+  firstChunk: `${TRUNCATED_COMPONENT_KEY}__firstChunk`,
+  splitterChunk: `${TRUNCATED_COMPONENT_KEY}__splitterChunk`,
+  secondChunk: `${TRUNCATED_COMPONENT_KEY}__secondChunk`,
+}
```
cen10: This PR is modeled off of EditableField.Truncated, so I've used the same pattern and put these constants into a util file. Ditto for `EditableField.Truncated.css.ts` becoming `Truncate.WithSplitter.css.ts`. The reason for not just reusing EditableField.Truncated was suggested by Nickolas since people may not know if it's reused elsewhere and any changes may then break this feature in the Truncate component.

In order to prevent that scenario from happening but also DRY things up, is there a pattern in this codebase for pulling out components that may be reused in multiple places? The code in this util file and `Truncate.WithSplitter.css.ts` could then be clearly marked as reusable and any changes there would be understood to have farther reaching repercussions.

stories/Truncate.stories.js
```diff
@@ -10,52 +10,67 @@ const limit = 10
 
 stories.add('default', () => (
   <div>
-    <p>
-      Auto:<br />
+    <p style={{ width: '50%' }}>
+      Auto:
+      <br />
```
cen10: This file auto-formatted to move the `<br />` to the next line -- LMK if I should change this back/modify my formatter.

cen10:
@knicklabs The Travis build is failing and I'm not sure what to do about it. I see the same 3 warnings when I run `npm run dev` on the latest master locally, so it's not related to the code changes. I don't see this problem on other open PRs. When this kind of thing happened at my old job I could restart the build and see if it was being flaky, but I don't see that button available to me in the Travis UI. Any suggestions?
```
(node:70556) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error
originated either by throwing inside of an async function without a catch block, or by
rejecting a promise which was not handled with .catch().
```

src/components/Truncate/Truncate.utils.ts
knicklabs: In this particular scenario, what do you think if we do a follow-up PR to utilize this change in the EditableFields component? Then we could get rid of the EditableField.Truncated component.

src/components/Truncate/Truncate.tsx
```diff
@@ -117,14 +120,37 @@ export class Truncate extends React.PureComponent<
...
+          }`}
+        >
+          <span className={`${TRUNCATED_CLASSNAMES.firstChunk}`}>{first}</span>
+          <span className={`${TRUNCATED_CLASSNAMES.splitterChunk}`}>
+            {splitter}
+          </span>
+          <span className={`${TRUNCATED_CLASSNAMES.secondChunk}`}>
+            {second}
+          </span>
+        </TruncateWithSplitterUI>
```
knicklabs: In terms of the class names here, what do you think with sticking with the `bem` approach? Perhaps, `c-c-Truncate__content__chunks` for the wrapper around chunks and `c-Truncate__content__chunks__first` for a chunk. Also, instead of `withSplitter` could we use `with-splitter` for consistency with other modifiers, like `is-visible`?

src/components/Truncate/Truncate.utils.ts
knicklabs: In terms of the class names here, what do you think with sticking with the `bem` approach? Perhaps, `c-Truncate__content__chunks` for the wrapper around chunks and `c-Truncate__content__chunks__first` for a chunk. Also, instead of `withSplitter` could we use `with-splitter` for consistency with other modifiers, like `is-visible`?

cen10:
```
In this particular scenario, what do you think if we do a follow-up PR to utilize this change 
in the EditableFields component?
```
Ah, yep, much cleaner.
```
[ ... ] what do you think with sticking with the bem approach?
```
Happy to make these changes! For my own learning, can you touch on this a bit more? I thought the code here (copy/pasted from EditableFields) might be the current approach since that component has been updated more recently than Truncated. Are we trying to move things over to `bem` as we go, and this just wasn't addressed yet in EditableFields?

cen10:
@knicklabs Looked into this:
1) I'm only seeing a tooltip show up where it shouldn't on the `Auto` example of the `Default` page -- are you seeing it across all examples on that page? The reason the tooltip shows up on the Auto example is because the boolean prop `showTooltipOnTruncate` [was added to that component](https://github.com/helpscout/hsds-react/pull/320/files#diff-dd430a37f3240a2f0fffc5b94c7d9c0aR15). I'm not sure why, but it seems unnecessary now and I can remove it as part of this PR.

2) I also noticed what appears to be a regression, at least in Storybook. You can see [a gif](https://user-images.githubusercontent.com/2322354/43473312-c9ac2096-94bd-11e8-9cee-4ff27740de43.gif) in the PR I link to above that shows how Auto is expected to work -- the text gets truncated as the window resizes smaller, and the font does not change. Currently, the text does not truncate and the font size jumps back and forth -- this is how it looks on master on FireFox (and Safari, but the recording is FF):
Screen Recording 2019-09-18 at 09 44 AM

On Chrome, the text isn't truncated as it should be if the window is too small, and there is slightly different font size behavior but you do see it jump from one to another in certain situations.

Since `auto` is the default behavior, I'm surprised we haven't seen this being an issue live -- maybe it's just a Storybook problem? I can try finding some examples on prod later on today.

knicklabs:
1. Yes, I do only see the tooltip on the `auto` example.
2. Nice catch! I think would revert that change you made and make an issue for this and then we can address that in a separate PR. I wonder if it's an issue with the resize handler. I'm not sure we use auto in the wild in any situations where resizing is impacting the layout.

cen10:
@knicklabs Updates made! Tooltip is missing from auto on the default page. I checked the truncating behavior on Chrome/FF/Safari.

src/components/Truncate/Truncate.utils.ts
knicklabs: That was likely an oversight in the original component. We've been using a variant of bem throughout the project. Only difference is our modifies aren't namespaced. So we've been using `is-visible` as opposed to `component--is-visible` which is more conventional.

PR #704, opened 2019-09-16
EditableField: Programmatic Control of Validation Info

knicklabs:
This update allows for validation info to be set externally via props. This is sometimes necessary, such as in cases where we need to provide feedback via server-side errors.


knicklabs:
Going to do a beta release and try this out in app to make sure it meets our needs.

knicklabs:
Taking a different route.

PR #705, opened 2019-09-17
Fancy -> Styled Components

plbabin:
**HSDS V3 release**
_Still a work in progress_

This update is trying to replace Fancy with the library styled-components. The way of generating css-in-js should be very similar, if not 99.9% the same so it should not impact our way of styling components. That being said, there is a lot of difference between the two libraries and this update is trying to close the gap to keep Help scout project working exactly the same way.

Styled-components offer a smaller bundle and a faster rendering engine compare to what we have right now with fancy (fork of Emotion 9). Moving away from fancy will allow us to leverage the community behind Styled-components that keep improving the library on a daily basis.

Enough talking, let's dive in what this PR is changing to our library:

### Update `styled``
Originally we had a component called `styled` that was importing `fancy` and exposing it to our application.

```javascript
import styled, 
   { ThemeProvider, FrameProvider, ScopeProvider } 
from "@helpscout/hsds-react/components/styled"
```

We update this component to import directly `styled-components`. It will also expose the ThemeProvider from this library we should work exactly like the one in fancy. `FrameProvider` and `ScopeProvider` will be removed from that component

Another update we did is to import directly from styled-component locally instead of doing it from our customize styled component (sorry I know it's confusing). Centralizing the import was nice, but it was a pain when refactoring or importing from a different directory. Now no need to deal with relative path:

```javascript
import styled from "styled-components"
```

### Frame component
Fancy did create a custom FrameProvider component to sync styles between iframe and the root document. That component is decrecated and being replace with a `<Frame />` component. Instead of using a regular `iframe` in your code, simply use the `<Frame />` component and it will take care of rendering the iframe and syncing the styles

```javascript
import Frame from "@helpscout/hsds-react/components/frame"

class MyApp extends React.Component{
  render(){
    return (
      <Frame>
        <div id="App">
          <Card>
            <Text>
              <Tooltip title="Elf!" triggerOn="click">
                Buddy!
              </Tooltip>
            </Text>
          </Card>
      </div>
    </Frame>)
  }
}
```

### Typescript
All our components are written using typescript. The good news is that `styled-components` support typing, but the bad news for us is that our code was not ready for that. This update fix a lot of typing issue from our css files.

When passing props to the styled component, we need a prop interface:
```javascript
export interface Props {
  className?: string    className?: string
  innerRef: (node: HTMLElement) => void   ref: (node: HTMLElement) => void
}

const WrapperUI = styled.div<Props>`...`
```

When adding a `ref` prop, typescript was complaining for a reason still unknown to me. Solution was to force the typing of the method:

```javascript
<ButtonWrapperUI align={align} ref={this.setNodeRef as any}>
```

### innerRef deprecated
`innerRef` was a prop made available by Emotion to get access to the dom element node. With `styled-components`, we need to use directly `ref`. **This might cause some issue with components using the hsds-react library**

```javascript
<ContentUI data-cy="ActionSelectContent" ref={this.setNodeRef}>
```

### remove `namespacedComponent` 
all components declaration in `index.ts`  was declaring himself in our namespacedComponent. It is no longer needed, except when there is propProvider involve. The good news is that PropProvider will push the component in the namespaced object if it was not already declared.

### SC_ATTR
`styled-components` allow us to namespace our `style` element to not interfere with other `styled-components` instance on given page. This env variable needs to be set by the process that will compile the styles (storybook or webpack for example)

`SC_ATTR=HSDS-react start-storybook -p 8900`

## Next steps to complete this PR

**Fix tests & coverage**
- [ ] Dropdown.Item
- [ ] Frame
- [ ] Tooltip Reference (if it used somewhere in hs-app/embed)
- [ ] contextConnect
- [ ] propConnect

**Other stuff to fix**
- [ ] Fix ScopeProvider 
- [ ] Implement globalStyling
- [ ] Deciding of PropProvider destiny
- [ ] Validate that `<Frame />` is working as expected (IE11 as well)
- [ ] Test everywhere


tinkertrain:
This is looking great! I can't thank you enough for taking on this big task, the benefits are clear!

cen10:
This is a serious undertaking, great job!

```
With styled-components, we need to use directly ref. This might cause some issue with components using the hsds-react library
```
^^ How do you think we can test for that?

package.json
```diff
@@ -108,7 +108,9 @@
     "path-to-regexp": "2.4.0",
     "popper.js": "1.14.3",
     "react-sortable-hoc": "^1.10.0",
+    "react-styled-frame": "^1.1.0",
     "react-transition-group": "2.5.3",
+    "styled-components": "5.0.0-beta.8",
```
tjbo: `beta.8`, scares me. I'm guessing we are going to bump this eventually?

src/components/Dropdown/V2/Dropdown.Item.tsx
```diff
@@ -238,11 +238,14 @@ export class Item extends React.PureComponent<Props> {
...
   setNodeRef = node => {
     this.node = node
-    this.props.innerRef(node)
+    this.props.ref(node)
   }
   setActionNodeRef = node => (this.actionNode = node)
   setWrapperNodeRef = node => (this.wrapperNode = node)
-  setMenuNodeRef = node => (this.menuNode = node)
+  setMenuNodeRef = node => {
+    console.log('menuRef', node)
```
tjbo: console.log here

src/components/Dropdown/V2/Dropdown.Menu.tsx
```diff
@@ -71,10 +71,7 @@ export class Menu extends React.PureComponent<Props> {
     )
 
     return (
-      <MenuWrapperUI
-        className="c-DropdownV2MenuWrapper"
-        innerRef={innerWrapperRef}
-      >
+      <MenuWrapperUI className="c-DropdownV2MenuWrapper test" ref={wrapperRef}>
```
tjbo: extra test class name here

src/components/Dropdown/V2/Dropdown.Item.tsx
tjbo: actually many, wip I know....

package.json
plbabin: yes, the goal was to start on this PR with the latest. I expect that v5 will be out when we'll be ready to move forward with our PR. 

beta-9 was out 10 days ago and 4.4 is the last release before v5

src/components/Dropdown/V2/Dropdown.Item.tsx
plbabin: leftovers from debugging why the dropdown was not rendering correctly (actually it was because styles where render in both styled-component and emotion, took me 2 days to figure that out 🤦‍♂ )

tjbo:
This is looking really good. It's a big PR but I have faith in PL and our very large test suite. And I think the styled-components API is a big upgrade on what we have now.

I think it would be great to give QA a heads up on this, as it's probably going to require quite a bit of regression testing and I know Ryan at least likes to know about these things in advance

src/components/Dropdown/V2/Dropdown.Item.tsx
tjbo: That Dropdown causes a tonne of problems.

src/components/Flexy/Flexy.Block.tsx
```diff
@@ -24,6 +24,4 @@ class Block extends React.PureComponent<Props> {
   }
 }
 
-namespaceComponent(COMPONENT_KEY.Block)(Block)
```
Charca: 🙌

plbabin: don't celebrate too quickly :) 

`namespaceComponent` is added back by propConnect, another reason I would love to get rid of this system

Charca: Haha, well at least we'll see this function less often, so I'd still consider it a win :P

plbabin:
pushing another update that fixes almost all tests and coverage. There is component that would require more time to fix all the test:

- [] Dropdown.Item
- [] Frame
- [] Tooltip Reference (if it used somewhere in hs-app/embed)
- [] contextConnect
- [] propConnect

plbabin:
merging in v3, to continue development of the next version. What's left to do or to fix will be done in this branch (there is a lot of cards covering what needs to be done)

PR #707, opened 2019-09-18
Upgrade Code Syntax Highlighting for CopyCode

tjbo:
This PR updates Code Copy to have syntax highlighting. It appears not to have been working. We also customized it to Buzz's specifications (roughly)

We had a component called highlight, but it was only being used by CodeCopy, to simplify things, I removed it.

cc: @digitaldesigner


src/components/CopyCode/CopyCode.tsx
```diff
@@ -27,7 +36,7 @@ class CopyCode extends React.PureComponent<Props> {
     innerRef: noop,
     onCopy: noop,
   }
-  codeNode: HTMLElement
+  node: HTMLElement
```
tjbo: codeNode just sounds too nerdy for me. 😆

digitaldesigner:
Mate this is perfect, really nice work! One tiny thing, can you remove 10px padding from the top and 10px from the bottom — just to reduce the leading and trailing whitespace. So pumped :)

PR #708, opened 2019-09-18
Update Truncate documentation

cen10:
Documentation updates that got left out of #703 accidentally:

This PR updates the `<Truncate />` documentation to include the new `splitter` prop, and removes some unnecessary props from the new splitter example in the Truncate story.


cen10:
@plbabin I forgot to update the README for Truncate in the last release 🤦‍♀ , can you take a look when you get a chance? Also removed some stray props in storybook for this example. The component functionality is unaffected.

PR #710, opened 2019-09-25
EditableField: Fixes email type UI glitches

tinkertrain:
## Problems

There were a couple of visual bugs on email type fields reported by Jared:
1. Extra spacing: http://c.hlp.sc/ac63799be531
2. Weird deletion behaviour when space present in value: http://c.hlp.sc/7a4b054a1277

### 1. Extra spacing

Before, I had a minimum width on the "first chunk" (the part before the _@_), which in cases like a single letter name, would be too much and show extra space.

I removed the minimum width and instead applied a couple of rules to the second chunk (the domain), I also simplified the markup and included the _@_ in the second chunk to avoid some cases in which I noticed there was some extra space added to the first chunk because of how flexbox behaves.

### 2. Character deletion when spaces present

This was a Chrome only bug, present when adding an "email" type to the input, I'm not sure why exactly, but the solution was to give the input a type of text even when given an email type, this shouldn't be a problem because we are using our own validation and not the browser bundled one.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,7 +30,6 @@ module.exports = {
     'no-dupe-class-members': 'off',
 
     // Add TypeScript specific rules (and turn off ESLint equivalents)
-    '@typescript-eslint/no-angle-bracket-type-assertion': 'warn',
     'no-array-constructor': 'off',
     '@typescript-eslint/no-array-constructor': 'warn',
     '@typescript-eslint/no-namespace': 'error',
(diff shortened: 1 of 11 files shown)


jaredmcdaniel:
Welcome back JP! Looks good. Just checking though ... would we throw validation in this case (due to the space)?

image

knicklabs:
@jaredmcdaniel validation is handled outside of the component. The component's responsibility is just to display the errors.

PR #712, opened 2019-09-26
EditableField: Disabling multiple fields at the same time

tinkertrain:
This PR addresses @knicklabs suggestion:

> When we are validating a field asynchronously we disable it and we keep track of the disabled item in the local state. I believe the disabled item should be an array of fields, not a single field, since it is possible that more than one field could be validating at one time.

I updated the validation story 📚 , it includes a field to enter a number in ms to delay the resolving of the validation promise (makes it easier to visualize how the disabling is behaving)


knicklabs:
Closing this since https://github.com/helpscout/hsds-react/pull/714 covers it.

PR #714, opened 2019-09-30
Fix/editable fields validate empty

tinkertrain:
https://helpscout.atlassian.net/browse/HSDS-78

## Problem

EditableField not triggering validate when value is changed to empty

## Solution

✅ I moved the logic responsible for dealing with empty values in the "blur" event handling from the bottom of the method (which was not being validated) to the validation stage, from there the logic is practically the same as it was.

## Multi-value fields

On multi-value fields we have two scenarios:
If we clear the value on one field:
a) There is more than one value present: clear the value, remove the input
b) If there is only one value: clear the value, leave the input

Considering that, this PR follows this flow:
Let's say the validation includes: required and email type
The field has 2 emails already

1. The user clears the 2nd email
2. We remove it, no validation happens. We would only try to validate when the value is something other than ''
3. The user clears the remaining email
4. Validation fails (required)

## Tests

✅ Updated tests to account for the async nature of the validation
📚 Use the validation stories to confirm that validation happen on blur on empty values: https://deploy-preview-714--hsds-react.netlify.com/iframe.html?id=editablefield--validation


src/components/EditableField/EditableField.Input.tsx
```diff
@@ -326,7 +326,7 @@ export class EditableFieldInput extends React.Component<InputProps> {
             innerRef={this.setInputNode}
             name={name}
             placeholder={placeholder}
-            type={type === 'email' ? 'text' : type}
+            type="text"
```
cen10: Would we ever use EditableField for something like a password field, where we would want the type to be `password` in order to obscure the text?

knicklabs: Great point! I've made this change.

knicklabs:
Once this is green I will merge this up.

src/components/EditableField/EditableField.Input.tsx
```diff
@@ -326,7 +326,7 @@ export class EditableFieldInput extends React.Component<InputProps> {
             innerRef={this.setInputNode}
             name={name}
             placeholder={placeholder}
-            type={type === 'email' ? 'text' : type}
+            type={type === 'password' ? type : 'text'}
```
plbabin: we still need to make sure that when it is email we change it to text. I think it was to fix an obscur bug in chrome

something like that could work:
```suggestion
            type={['password', 'email'].includes(type) ? type : 'text'}
```

src/components/EditableField/EditableField.tsx
```diff
@@ -225,71 +227,89 @@ export class EditableField extends React.Component<
...
-      initialFieldValue,
-      multipleValuesEnabled,
-      validationInfo,
-    } = this.state
-    const { validate, onCommit, onInputBlur } = this.props
+    const { activeField, multipleValuesEnabled } = this.state
+    const { validate, onCommit, onDiscard, onInputBlur } = this.props
 
-    if (equal(initialFieldValue, fieldValue)) {
+    if (equal(this.state.initialFieldValue, this.state.fieldValue)) {
```
plbabin: any reason why you moved away from destructuring the state object (just curious)

src/components/EditableField/EditableField.tsx
```diff
@@ -225,71 +227,89 @@ export class EditableField extends React.Component<
 
   handleInputBlur = payload => {
```
plbabin: I know we are short in time to ship EditableField, but I do think it would be nice to cleanup this blur method at some point, separate it in multiple utils or class method just to help with readability

src/components/EditableField/EditableField.Input.tsx
knicklabs: @plbabin do you mean when it is email we should leave it as email? Currently, it changes everything except password to text.

src/components/EditableField/EditableField.tsx
knicklabs: I'm not sure how this change snuck in there :(

src/components/EditableField/EditableField.tsx
knicklabs: I'd like to overhaul this API. I started a paper doc with some ideas, but in a nutshell, I believe validate should be renamed to save and it should be a cancelled promise that resolves when valid and rejects when invalid. Cancelling should trigger escape.

PR #715, opened 2019-10-02
EditableTextarea component

tinkertrain:
New component! Editable Textarea

Prototype: https://codesandbox.io/s/editabletextarea-i2ejs

## Behaviours

- Use Enter to save the note
- Use Shift + Enter to add line breaks
- Use Escape to discard changes
- When the content is larger than the height of the teaxtarea, we add a visual cue (gradient)
- The visual cue disappears when scrolling to the bottom of the textarea
- The visual cue's color is configurable, in case the textarea background is different than white
- Scrolling is locked to the textarea (so if you keep scrolling down, the window does not scroll, which can be annoying, especially since the field is not particularly long)
- Number of rows (max height of the textarea) is configurable (default 5)
- Styles follow EditableFields as close as possible

📚 [Storybook](https://deploy-preview-715--hsds-react.netlify.com/iframe.html?id=editabletextarea--default)
📹 http://c.hlp.sc/ece42d7ad31e/textareas.mp4

## Todo

- [x] Wire up callbacks
- [x] Tests
- [x] Docs


cen10:
@tinkertrain I noticed some behavior in the storybook that seems odd -
https://share.getcloudapp.com/12uEBPGK
1. I can't tell that there's a gradient in the top box where there's already content that extends beyond the height of the element.
2. The gradient does exist in the bottom with the placeholder text, but it's difficult to click into and focus jumps back to the first box unless you manage to hit the sweet spot (maybe this is just an issue with how it's implemented in Storybook and not a problem with the component itself).

src/components/EditableTextarea/README.md
```diff
@@ -0,0 +1,41 @@
+# EditableTextarea
+
+A EditableTextarea it's a sibling of the EditableField, but a textarea...
+
+- It autoresizes it's height based on content to a maximum number of rows (maxRows)
+- Use Enter to save the note
+- Use Shift + Enter to add line breaks
+- Use Escape to discard changes
+- When the content is larger than the height of the teaxtarea, we add a visual cue (gradient)
+- The visual cure disappears when scrolling to the bottom of the textarea
```
cen10: Small typo - visual `cue`

jaredmcdaniel:
@tinkertrain posted an update in slack, but scrapped together a bare-bones [prototype](https://www.loom.com/share/4e0efebaf28442e4bca78e8a67bbeb5b) to help fill in some gaps Im seeing. Namely ...

- Can we remove the horizontal scrollbar? I believe removing `overflow:scroll;` altogether will give us what we need since `textarea` shows scrolling by default and we'll only ever want vertical  (up/down) scrolling.
image

- Once a value exists, lets use a solid border for hover (vs dashed)
- Would like to only show scrollbar on hover and focus
- Custom style scrollbar to remove background frame
- Make fader more prominent (taller) if possible
- Show only four lines of text by default (but love that this is a prop we can customize)

jaredmcdaniel:
@tinkertrain small update. I need to go back on this one, as it looks like we _do_ used dashed borders for hovers across other editable fields. My bad! 

>Once a value exists, lets use a solid border for hover (vs dashed)

Feel free to keep that in and let's see how it comes together. Thanks!

knicklabs:
@tinkertrain here's a diff to a branch with some rough changes regarding the scrollbar and hover/focus states in case it's useful: https://github.com/helpscout/hsds-react/compare/feature/editable-textarea...feature/editable-textarea-tweaks?expand=1

tinkertrain:
> @tinkertrain I noticed some behavior in the storybook that seems odd -
> [share.getcloudapp.com/12uEBPGK](https://share.getcloudapp.com/12uEBPGK)
> 
>     1. I can't tell that there's a gradient in the top box where there's already content that extends beyond the height of the element.
> 
>     2. The gradient does exist in the bottom with the placeholder text, but it's difficult to click into and focus jumps back to the first box unless you manage to hit the sweet spot (maybe this is just an issue with how it's implemented in Storybook and not a problem with the component itself).
> 
> 
> ETA: Apart from this weirdness, this is an awesome addition!

Thanks for the feedback! Although I haven't seen that before and couldn't reproduce :/

Are you still seeing that after the latest update? Which browser?

Thanks!

tinkertrain:
> @tinkertrain small update. I need to go back on this one, as it looks like we _do_ used dashed borders for hovers across other editable fields. My bad!
> 
> > Once a value exists, lets use a solid border for hover (vs dashed)
> 
> Feel free to keep that in and let's see how it comes together. Thanks!

Thanks for the feedback!

Hehe, those hover states I just copied form the EditableFields :)

I pushed an update, scrollbars are looking OK, I tested in Chrome, Safari and Firefox, even Firefox has a little option to improve it although is not perfect (you can see it in the video below, is a bit weird but I think the feature is quite new)

A little demo: in order, chrome, safari, firefox:
📹 http://c.hlp.sc/ece42d7ad31e/textareas.mp4

jaredmcdaniel:
@tinkertrain nice work man ... couple CANIs ...

1. Can the top "blank slate" version say `Add notes` and display inline (border only under text vs full width), like we do for the other text fields?

**Current:** 
image

**Requested:**
image

2. Any opposition to making the following tweaks to the fader `:before` element?
```
height: 50px;
background: linear-gradient(to top, white 0%, rgba(255, 255, 255, 0) 100%);
```

Cc'ing @knicklabs, @plbabin or @cen10 in case they're able to weigh in sooner than later.

src/components/EditableTextarea/EditableTextarea.Mask.tsx
```diff
@@ -0,0 +1,89 @@
+import * as React from 'react'
```
cen10: Ignore this file please, I remove it in the next commit -- was a leftover from an initial attempt that accidentally made its way into staging.

cen10:
@jaredmcdaniel https://share.getcloudapp.com/d5ubxZjp <- Shows the border getting darker on hover.

cen10:
@jaredmcdaniel Please take a look when you have a chance - there's still some weird jumpiness when you enter text and hit "enter", but wanted to get your eyes on it for anything else that might stick out.

cen10:
No jumpiness when clicking into fields, hitting enter, or hitting escape. https://share.getcloudapp.com/NQuezvDk

cen10:
@knicklabs @plbabin Can I get your eyes on this please? Jared has signed off on the design. 🎊 

I'll squash these commits before merge.

knicklabs:
I think one thing we're going to need is validation. I think where we're using this though, there is very low risk of errors. Do we want to push forward with this and do that as a follow-up or should we tackle it now. Thoughts @cen10 @jaredmcdaniel @tinkertrain? Getting this out, will allow us to reduce friction customers are having with the notes field. We also require this for the lines field on the new address block.

cen10:
@knicklabs I'm assuming backend validation already exists for the current notes field and we can hook into that? If so, then I like the idea of getting this out sooner.

knicklabs:
@cen10 yes the logic exists but the error won't be displayed is all without the component supporting it. That said, we're not doing any sort of format-based validation here so it's only going to be a problem if there is a server error. I agree with you both about shipping now and iterating afterwards. It's looking good to me. I need this in my current PR so let me just make a beta build and make sure its going to work properly so we don't have to go back and forth with multiple releases.

src/utilities/debounce.ts
```diff
@@ -0,0 +1,27 @@
...
+        return fn.apply(context, args)
+      }
+    }, wait)
+
+    if (callNow) {
+      // @ts-ignore
+      return fn.apply(this, arguments)
+    }
+  }
+}
```
plbabin: I am surprise we didn't had any debounce function in hsds-react 🤔

knicklabs:
@tinkertrain one thing we'll need to sort out is how to get this working with the EditableFieldComposite. We have a case where we need to do such a thing.

PR #716, opened 2019-10-02
COMMS-299: Truncate the select dropdown labels when they're really long

tompedals:
https://helpscout.atlassian.net/browse/COMMS-299

Before:

Screen Shot 2019-10-02 at 15 11 16

After:

Screen Shot 2019-10-02 at 15 09 11


malikalim:
@tompedals is this on develop? I still see it not fixed on proxy pack http://c.hlp.sc/9db753d951bd

PR #717, opened 2019-10-02
HSDS-77 Fix Truncate component to show tooltip when truncated on splitter'

cen10:
JIRA: https://helpscout.atlassian.net/browse/HSDS-77

### Problem 
1. When using `<Truncate />` with a splitter prop so you can truncate email addresses, the tooltip feature isn't working. It should show the non-truncated email address on hover.

2. If you have an email address like `g@example.com`, with the splitter prop set, it renders out as `g @example.com`. This is because the first span has a min-width of 21px, which is wider than a character.

#### Tooltip feature:
The `<Tooltip />` component will wrap around the truncated content if a) the `showTooltipOnTruncate` prop is true and b) `this.state.isTruncated` is true. The latter makes sure that we only show the Tooltip if the content has been truncated. 
  
For content that isn't using a splitter, one parent `<span>` wraps around the content, and we can easily compare the length of the content to the width of the span and see if it overflows using `this.contentNode.offsetWidth > this.node.offsetWidth`. However, content that IS being truncated based of the presence of the splitter has a different HTML structure and the width of the two spans containing the pre- and post- splitter content is very nested. 

This PR gains access to those nested spans to compare their combined width to that of their container in order to accurately update `this.state.isTruncated` and allow the `<Tooltip />` to wrap the truncated text.

#### Spacing issue
A solution to the spacing problem has already been implemented by Juan Pablo in [this PR](https://github.com/helpscout/hsds-react/pull/710/files#diff-ee19278829622bce00200f9b5ecd5584R8-R20) and has been reimplemented here.

### Testing
I've updated the Truncated => Tooltips story to include an email address. 
* [ ] There should be no tooltip if the email is not truncated. 
* [ ] There should be a tooltip if the email is truncated
https://share.getcloudapp.com/o0ux4wQR

I've updated the Truncated => default story to include an email address with a one character user name.
* [ ] There should be no space between the username and the `@` sign, e.g. `a@domain.com`.

Note:
In testing, I noticed an issue that exists in EditableFields as well. https://share.getcloudapp.com/p9uzEgBO 
An email will truncate, then as the width of the display gets even smaller it displays the first few letters without truncation and things appear awkwardly cut off. This may be related to the fact that the domain names in these examples are very long, so even with truncating the username, the text still spills over. 

A possible solution is to truncate the domain name as well, but this might end up looking even more awkward: `a...@....com`, `a...@along....com`, `a...@along...`. I'll bring this up to Design, but as it's not a regression with this PR it shouldn't be a blocker here.

### DIFF
--- a/src/components/Truncate/Truncate.tsx
+++ b/src/components/Truncate/Truncate.tsx
@@ -76,16 +76,28 @@ export class Truncate extends React.PureComponent<
       this.contentNode.style.display = 'initial'
       // 2. Calculate the differences
       const isContentTruncated =
-        // TODO: fix typescript complains
-        // @ts-ignore
-        this.contentNode.offsetWidth > this.node.offsetWidth
+        props.splitter && props.showTooltipOnTruncate
+          ? this.isSplitContentTruncated(this.contentNode, this.node)
+          : // TODO: fix typescript complains
+            // @ts-ignore
+            this.contentNode.offsetWidth > this.node.offsetWidth
       // 3. Resets the display
       // TODO: fix typescript complains
       // @ts-ignore
-      this.contentNode.style.display = null
 
       return isContentTruncated
     }
+  }
+
+  isSplitContentTruncated = (contentNode: any, node: any): boolean => {
+    const [
+      firstSpan,
+      secondSpan,
+    ] = node.children[1].children[1].children[0].children[0].children[0].children
+
+    return (
+      contentNode.offsetWidth < firstSpan.scrollWidth + secondSpan.scrollWidth
+    )
   }
 
   getText = (props: TruncateProps = this.props) => {
@@ -133,10 +145,8 @@ export class Truncate extends React.PureComponent<
           }`}
         >
           <span className={`${TRUNCATED_CLASSNAMES.firstChunk}`}>{first}</span>
-          <span className={`${TRUNCATED_CLASSNAMES.splitterChunk}`}>
-            {splitter}
-          </span>
           <span className={`${TRUNCATED_CLASSNAMES.secondChunk}`}>
+            {splitter}
             {second}
           </span>
         </TruncateWithSplitterUI>
--- a/src/components/Truncate/Truncate.tsx
+++ b/src/components/Truncate/Truncate.tsx
@@ -90,13 +90,10 @@ export class Truncate extends React.PureComponent<
   }
 
   isSplitContentTruncated = (contentNode: any, node: any): boolean => {
-    const [
-      firstSpan,
-      secondSpan,
-    ] = node.children[1].children[1].children[0].children[0].children[0].children
-
     return (
-      contentNode.offsetWidth < firstSpan.scrollWidth + secondSpan.scrollWidth
+      contentNode.offsetWidth <
+      node.querySelector(`.${TRUNCATED_CLASSNAMES.firstChunk}`).scrollWidth +
+        node.querySelector(`.${TRUNCATED_CLASSNAMES.secondChunk}`).scrollWidth
     )
   }
 
--- a/stories/Truncate.stories.js
+++ b/stories/Truncate.stories.js
@@ -37,10 +37,16 @@ stories.add('default', () => (
       </Truncate>
     </p>
     <p>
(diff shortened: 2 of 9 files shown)


src/components/Truncate/Truncate.tsx
```diff
@@ -76,18 +76,30 @@ export class Truncate extends React.PureComponent<
...
 
       return isContentTruncated
     }
   }
 
+  isSplitContentTruncated = (contentNode: any, node: any): boolean => {
+    const [
+      firstSpan,
+      secondSpan,
+    ] = node.children[1].children[1].children[0].children[0].children[0].children
```
cen10: We access this.node and this.contentNode so nicely using refs, is there a similar way to access these nested spans so that this isn't so ugly?

src/components/Truncate/Truncate.tsx
```diff
@@ -76,18 +76,30 @@ export class Truncate extends React.PureComponent<
...
       // 3. Resets the display
       // TODO: fix typescript complains
       // @ts-ignore
-      this.contentNode.style.display = null
 
       return isContentTruncated
     }
   }
 
+  isSplitContentTruncated = (contentNode: any, node: any): boolean => {
```
cen10: I'm not sure what to put here so TS doesn't complain. Property `offsetWidth` doesn't exist on type Node or Element. I also get the error `HTML collection is not an array type` for type Element.

cen10:
Whoops, need to fix a test.

src/components/Truncate/Truncate.tsx
knicklabs: I think I'm missing something here but why do we need to reach so deeply into the node?

knicklabs: If the spans have classnames and node is an instance of HtmlElement can we use `node.querySelector` here instead?

stories/Truncate.stories.js
```diff
@@ -40,6 +40,7 @@ stories.add('default', () => (
       Truncate by Splitter:
       <br />
       <Truncate splitter="@">longemailaddress@gmail.com</Truncate>
+      <Truncate splitter="@">a@myemail.com</Truncate>
```
knicklabs: Can we add some additional examples here so we can visually see how this looks under more scenarios?

src/components/Truncate/Truncate.tsx
cen10: That's how deeply nested the two spans are that wrap the first and second chunks of the email. We need to get the scroll width of those two spans and compare it to the width of the parent span, and if the combined scroll width is larger than the parent, the text should be truncated. Good call with node.querySelector.

stories/Truncate.stories.js
cen10: Copying the examples from the Editable Fields story.

src/components/Truncate/Truncate.tsx
tinkertrain: Have you tried `HTMLElement`? I wouldn't worry too much about it, sometimes TS just likes to be troublesome and not worth wasting your time.

cen10: With the `HTMLElement` type it complains that "Object is possibly null"  for `this.node` and "Argument of type 'null' is not assignable to parameter of type 'HTMLElement'" for `this.contentNode`. 🤔

cen10:
@knicklabs I tested the (accidental) 2.65.2 release on the branch for HSAPP-89 and saw that the spacing issue is gone and tooltips work. I wasn't able to find the spacing issue in the Customer Management work on sometestco, but maybe I was looking in the wrong spot?

src/components/Truncate/Truncate.tsx
knicklabs: It looks like this is the correct behavior of TypeScript because it can be null. In fact, we actually initialize those variables with null. I found this article on the topic really useful: https://medium.com/@martin_hotell/react-refs-with-typescript-a32d56c4d315 - TLDR: it seems we can satisfy the TypeScript compiler by guarding the use of the variables with if conditions.

PR #718, opened 2019-10-02
Update Avatar loading flow

plbabin:
This update introduce a new way to load the image inside an avatar.

Big thank you to https://github.com/mbrevda/react-image for the loading mechanism

Screen Recording 2019-10-02 at 04 53 PM

### What's new
* We removed  all the internal state to track down if the image was loading, loaded or failed.
* We removed the hidden `` tag that was used to load an image (and determine if it was successful or if it failed).
* All dynamic prop are handle inside styled component, instead of using the `style` prop
* Leverage the `Image` class and `decode()` to load the image. 
* Create a source list of images to load, if the first one fail it will skip to the next one until it works.
* Add a cache layer to skip all the loading process if the image was already loaded at some point
* Add the possibility to disable the fade in animation when the image finish loading

### Todos before releasing
- [x] Test in hs-app Customers
- [x] Test in beacon to make sure avatar are still working fine

### Next up
We could extract the loading image code into his own component so we could reuse it somewhere else

Another option would be to use the react-image library (where all the loading code is coming from anyway)


src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+  onError?: () => void
+  onLoad?: () => void
+  title: string | number
+}
+
+export interface State {
+  currentIndex: number
+  isLoading: boolean
+  isLoaded: boolean
+}
```
tinkertrain: Minor: For consistency and keeping the component file cleaner, can we move these to a Avatar.types.ts file?

src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+    }
+
+    this.state = this.sourceList.length
+      ? // 'normal' opperation: start at 0 and try to load
+        { currentIndex: 0, isLoading: true, isLoaded: false }
+      : // if we dont have any sources, jump directly to unloaded
+        { currentIndex: 0, isLoading: false, isLoaded: false }
+  }
+
+  srcToArray = src => (Array.isArray(src) ? src : [src]).filter(x => x)
```
tinkertrain: Curious, what is the `filter` achieving here?

src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+
+  onLoad = () => {
+    cache[this.sourceList[this.state.currentIndex]] = true
+    /* istanbul ignore else */
+    if (this.image) {
+      this.setState({ isLoaded: true })
+      this.props.onLoad && this.props.onLoad()
```
tinkertrain: I think because you have a default prop of `noop` for `onLoad`, the check is not needed.

src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+      // if we know it doesn't exist, skip it!
+      /* istanbul ignore next */
+      if (cache[src] === false) continue
+    }
+
+    // currentIndex is zero bases, length is 1 based.
+    // if we have no more sources to try, return - we are done
+    if (nextIndex === this.sourceList.length) {
+      this.setState({ isLoading: false })
+      this.props.onError && this.props.onError()
```
tinkertrain: Same as `onLoad` comment above

src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+    /* istanbul ignore else */
+    if (this.image) {
+      this.setState({ isLoaded: true })
+      this.props.onLoad && this.props.onLoad()
+    }
+  }
+
+  onError = () => {
+    cache[this.sourceList[this.state.currentIndex]] = false
+    // if the current image has already been destroyed, we are probably no longer mounted
```
tinkertrain: Useful comments, thanks!

src/components/Avatar/Avatar.Image.tsx
```diff
@@ -1,68 +1,263 @@
...
+    if (this.state.isLoading) this.loadImg()
+  }
+
+  componentWillUnmount() {
+    // ensure that we dont leave any lingering listeners
+    /* istanbul ignore else */
+    if (this.image) this.unloadImg()
+  }
+
+  componentWillReceiveProps(nextProps) {
```
tinkertrain: Just a note, I'm not entirely sure what we are going to do if they actually remove this method from React, `getDerivedStateFromProps` does not behave like this one does 😕

jaredmcdaniel:
n00b question ... but not sure the best way to test this one. Tried firing it up locally and getting errors ...

```ERROR in ./stories/Avatar/Avatar.stories.js
Module not found: Error: Can't resolve 'react-image' in '/Users/jaredmcdaniel/Git/helpscout/hsds-react/stories/Avatar'
 @ ./stories/Avatar/Avatar.stories.js 46:0-30
 @ ./stories sync .stories.(js|ts|tsx)$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true&quiet=true```

Also not seeing a Netlify link. 🤔 All that to say, love we're making progress on this one!

src/components/Avatar/Avatar.Image.tsx
plbabin: 👍 all done in 6048ade6

src/components/Avatar/Avatar.Image.tsx
plbabin: making sure that the each element in the array exists and that the loading flow will not try to load an empty url

src/components/Avatar/Avatar.Image.tsx
plbabin: good question, I am not sure either. We could use a variable just like the cache one, generate an id for each avatar, and store the image src using that id

we'll see when we upgrade to react 17 😄

src/components/Avatar/Avatar.Image.tsx
plbabin: all fixed in a55572bd

src/components/Avatar/Avatar.Image.tsx
plbabin: fixed in a55572bd

plbabin:
@jaredmcdaniel should be good to go now

tjbo:
@plbabin, do you have an HS-APP branch for this already?

plbabin:
Not currently. We did tried out the beta release with customer list/profile last week. There is another beta release available 2.69.0-0 if you want to try it out

tjbo:
Movie time: 🎥😺 

http://c.hlp.sc/bdfa05364bdb

plbabin:
> Movie time: 🎥😺
> 
> http://c.hlp.sc/bdfa05364bdb
> 
> Just a small change when the image loads, if it has an image src, it shouldn't have a background color.

oh so the background color is removed even if the image is not loaded yet?

digitaldesigner:
Yeah this is fine, I'd prefer it went from white -> image than blue -> image :)

PR #720, opened 2019-10-03
Fix Input Width if is Error

tjbo:
This PR fixes a bug with Error states in the Input Component.

In condition builder in Messages we were having a problem in FireFox with slimmer inputs. This fixes that problem, so that the width of the input is calculated if there is an error.

http://c.hlp.sc/997ff0c21c71

### DIFF
--- a/src/components/Input/styles/Input.css.js
+++ b/src/components/Input/styles/Input.css.js
@@ -224,6 +224,10 @@ export function makeFieldStyles() {
     width: 100%;
     z-index: 1;
 
+    &.is-error {
+      width: calc(100% - 24px);
+    }
+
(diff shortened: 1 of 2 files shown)


src/components/Input/styles/Input.css.js
```diff
@@ -224,6 +224,10 @@ export function makeFieldStyles() {
     width: 100%;
     z-index: 1;
 
+    &.is-error {
+      width: calc(100% - 24px);
```
plbabin: I assume 24 is the size of the icon?

tjbo: Exactly.

PR #721, opened 2019-10-04
Add content, heading and maxWidth to the RadioCard component

plbabin:
Screen Recording 2019-10-04 at 01 31 PM

This update adds the ability to render an heading and a content inside a RadioCard. It also add the possibility to limit the size of each card.

### DIFF
--- a/src/components/RadioCard/RadioCard.types.ts
+++ b/src/components/RadioCard/RadioCard.types.ts
@@ -4,16 +4,19 @@ export type InputEvent = Event
 export type RadioCardProps = {
   checked: boolean
   className?: string
+  content: string | Component
+  heading?: string
   icon: string | Component
   iconSize: number
   id?: string
-  inputRef: (node: InputNode) => void
   innerRef: (node: InputNode) => void
+  inputRef: (node: InputNode) => void
   isFocused: boolean
+  maxWidth?: string | number
   onBlur: (event: InputEvent) => void
+  onChange: (value: any) => void
   onFocus: (event: InputEvent) => void
   title?: string
-  onChange: (value: any) => void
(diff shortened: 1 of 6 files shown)


src/components/RadioCard/RadioCard.types.ts
```diff
@@ -4,16 +4,19 @@ export type InputEvent = Event
 export type RadioCardProps = {
   checked: boolean
   className?: string
+  content: string | Component
```
cen10: Do we need to indicate here that `content` is optional?

src/components/RadioCard/RadioCard.types.ts
```diff
@@ -4,16 +4,19 @@ export type InputEvent = Event
...
+  heading?: string
```
cen10: Since `getHeadingMarkup` also handles receiving a function, should this be `string | Component`?

src/components/RadioCard/RadioCard.types.ts
plbabin: good catch, I did improve the typing in 9f771542

src/components/RadioCard/RadioCard.types.ts
plbabin: good catch, I did improve the typing in 9f771542

plbabin:
> This looks awesome! Nice enhancement. Left a couple questions about Typescript types and I'm seeing something weird in the custom icon story? https://share.getcloudapp.com/kpudQwbB It seems like the default chat icon is being applied instead of the search one you called for in the story?

It was indeed not working correctly. I did push a change to render the custom icon if it's a react node.

PR #722, opened 2019-10-07
Fix: Tweak to EditableTextArea

cen10:
Only on Firefox, if you press "enter" in a field with placeholder text the focus indicator animates below the border that sits under the placeholder text instead of across the top of it.

Bug: https://share.getcloudapp.com/Z4uzP2YG

Fix: Move the border onto the textarea instead of the parent div to resolve the issue. This then caused problems across all browsers with the fader covering up the border, so I shifted the fader's position 1px. This caused a problem only on Firefox where you could see some unfaded text peeking between the bottom of the fader and the border, which was resolved when I removed a style specific to Firefox that was adding an additional 3px to the height of the textarea parent.

### DIFF
--- a/src/components/EditableTextarea/styles/EditableTextarea.css.ts
+++ b/src/components/EditableTextarea/styles/EditableTextarea.css.ts
@@ -2,7 +2,6 @@ import styled from '../../styled'
 import baseStyles from '../../../styles/resets/baseStyles.css'
 import { COLOURS, SIZES } from '../../EditableField/EditableField.utils'
 import { FONT_FAMILY } from '../../../styles/configs/constants'
-import { isFirefox } from '../../../utilities/browser'
 
 const { field, focusIndicator } = SIZES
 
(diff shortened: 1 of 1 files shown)


src/components/EditableTextarea/styles/EditableTextarea.css.ts
```diff
@@ -121,11 +115,6 @@ export const EditableTextareaUI = styled('div')`
     height: auto;
     border: 0;
     margin: 0;
-    ${/* 
-      I have no idea why Chrome and safari are adding 3px to the height of the parent of the textarea,
-      so I'm adding those 3px to firefox for parity  ¯\_(ツ)_/¯
-    */
-    isFirefox() && 'margin-bottom: 3px;'}
```
cen10: Is removing this having some undesirable effect that I'm not noticing? I needed to take it away in order for the border and fader to play nicely together on Firefox.

tinkertrain: I added this because chrome and webkit add some pixels to the container of the textarea, making firefox look a bit smaller, it was so that they look almost exactly the same, but I didn't notice the issue of the mismatched borders you're solving here.

If we just remove this, I don't see the issue anymore (maybe I'm missing something), so maybe the rest of the changes are not needed? Having all browsers to look exactly the same to the pixel is often futile, I thought this was an easy win, but it clearly wasn't!

🙈

cen10:
Closing b/c it turns out we only need to remove the FF style rule with margin-bottom and the rest of the changes aren't necessary  -- since there's another [open PR](https://github.com/helpscout/hsds-react/pull/723) that is adding validation, we're folding this small change into that.

PR #724, opened 2019-10-07
Fix avatar styles

tjbo:
https://helpscout.atlassian.net/secure/RapidBoard.jspa?rapidView=153&projectKey=COMMS&modal=detail&selectedIssue=COMMS-330

The user avatars shown on the Chat blank screen are A) blurry and B) have very subtle but noticeable blue pixelation around the image — both of which make them look bad.

This ticket fixes that by adding some styles.

### DIFF
--- a/src/components/Avatar/Avatar.Crop.tsx
+++ b/src/components/Avatar/Avatar.Crop.tsx
@@ -21,7 +21,6 @@ export const AvatarCrop = props => {
   )
 
   const styles = {
-    color: 'transparent',
     transition: `background-color ${animationDuration}ms ${getEasingTiming(
       animationEasing
     )}`,
--- a/src/components/Avatar/Avatar.Crop.tsx
+++ b/src/components/Avatar/Avatar.Crop.tsx
@@ -9,6 +9,7 @@ export const AvatarCrop = props => {
     animationEasing,
     className,
     children,
+    hasImage,
     isImageLoaded,
     withShadow,
   } = props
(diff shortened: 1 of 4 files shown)


src/components/Avatar/Avatar.Crop.tsx
```diff
@@ -21,6 +21,7 @@ export const AvatarCrop = props => {
   )
 
   const styles = {
+    color: 'transparent',
```
knicklabs: I think this transparent color needs to be applied only in the case where there is no image. See this example here https://deploy-preview-724--hsds-react.netlify.com/?path=/story/avatar--initials -- the initials are not visible due to the transparent background.

tjbo: Great catch. Fixed. 👍

tjbo:
@digitaldesigner @knicklabs 

Thanks for the feedback. I didn't know Avatars could be so complicated.

Updated the ticket. 👍

src/components/Avatar/styles/Avatar.css.ts
```diff
@@ -94,7 +94,8 @@ export const ActionUI = styled('div')`
 
 export const CropUI = styled('div')`
   align-items: center;
-  background-color: currentColor;
+  background-color: ${({ hasImage }) =>
+    hasImage ? 'transparent' : 'currentColor'};
```
knicklabs: Nice use of props here!

PR #726, opened 2019-10-08
HSDS-77 Update Truncate to fix Tooltips

cen10:
Missing old commits, I'm starting a new branch to avoid reverting the original revert of HSDS-77.


cen10:
Missing old commits, I'm starting a new branch to avoid reverting the original revert of HSDS-77.

PR #730, opened 2019-10-10
BEMBED-90: Added styles for MessageCard component on mobile viewports

Charca:
New UI with 100% width message, and bigger text overall for viewports < 400px wide.

Screen Shot 2019-10-10 at 12 34 27 PM


Charca:
@digitaldesigner could you take a quick look at the mobile styles for this component? Preview storybook is here: https://deploy-preview-730--hsds-react.netlify.com/?path=/story/messagecard--default

Just want to make sure spacing and font sizes are correct. Measurements outside the Message like positioning and margins will be taken care of in the embed.

image


Thanks!

PR #732, opened 2019-10-14
PROD-390 Add scheme to urls if missing

cen10:
Jira: https://helpscout.atlassian.net/browse/PROD-390

Problem: 
Websites that were missing `http://` or `https://` wouldn't open if you clicked the "new window" button. 

Solution:
Check for the presence of the scheme and add if missing.

To test:
In Storybook, check that the "new window" button in the EditableField URL story will open for websites with and without the `http://` scheme.

Note:
We had similar code inside the `convertLinksToHTML` util function, so I abstracted it into one new util and updated both `convertLinksToHTML` and `EditableField` to make use of it. 

Using a beta release, I checked that the following links opened as expected:
```
google.com
www.google.com
http://www.google.com
https://www.google.com
```

### DIFF
--- a/src/utilities/urls.ts
+++ b/src/utilities/urls.ts
@@ -1,0 +1,13 @@
+export const normalizeUrl = (string: string): string => {
+  if (!string) {
+    return ''
+  }
+
+  let url = string
+
+  if (url.search(/^http[s]?:\/\//) === -1) {
+    url = `http://${url}`
+  }
+
+  return url
+}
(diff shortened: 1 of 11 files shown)


src/utilities/urls.ts
```diff
@@ -0,0 +1,13 @@
+export const normalizeUrl = (string: string): string => {
+  if (!string) {
+    return ''
+  }
```
cen10: I don't think this util should be responsible for validating that what it's being given is a url. That said, if an empty string is given to it then it would return `http://` without this check, which felt weird. This much of a guard made sense to me but happy to change this up if needed.

tinkertrain: Makes sense to me 👍

PR #733, opened 2019-10-14
Add avator selector and new button styles

tjbo:
https://helpscout.atlassian.net/secure/RapidBoard.jspa?rapidView=153&projectKey=COMMS&modal=detail&selectedIssue=COMMS-283

**In this PR we are:**
1 - adding an AvatarSelector Component

2 - adding some styles to buttons to account for the pill shaped buttons

image

cc: @digitaldesigner


src/components/Button/styles/Button.config.css.js
```diff
@@ -0,0 +1,260 @@
+import { getColor } from '../../../styles/utilities/color'
```
tjbo: The other file was getting difficult to read, so seperated this out.

src/components/AvatarSelector/AvatarSelector.tsx
```diff
@@ -0,0 +1,47 @@
...
+  IconCaretUI,
+} from './styles/AvatarSelector.css.js'
+
+export interface Props {
+  image: string
+  isOpen: boolean
+  name: string
+}
+
+export default class AvatorSelector extends React.PureComponent<Props> {
```
tinkertrain: Little typo in class name "AvatorSelector"

src/components/AvatarSelector/AvatarSelector.tsx
```diff
@@ -0,0 +1,47 @@
...
+  static default = {
```
tinkertrain: I think it should be `defaultProps` here

tjbo:
Nice catch, fixed those up 👍

PR #734, opened 2019-10-25
Add missing component API props

TerrenceLJones:
This adds missing component api props, which was leading to the issue described: https://helpscout.atlassian.net/browse/PROD-455

### DIFF
--- a/src/components/EditableTextarea/EditableTextarea.tsx
+++ b/src/components/EditableTextarea/EditableTextarea.tsx
@@ -44,6 +44,8 @@ export class EditableTextarea extends React.PureComponent<
     value: '',
     onCommit: noop,
     onChange: noop,
+    onInputBlur: noop,
+    onInputFocus: noop,
     onEnter: noop,
     onEscape: noop,
     validate: () => Promise.resolve({ isValid: true }),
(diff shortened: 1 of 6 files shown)


knicklabs:
Putting this on hold while QA. If everything is looking good, will merge and release.

src/components/EditableTextarea/EditableTextarea.tsx
```diff
@@ -214,14 +223,27 @@ export class EditableTextarea extends React.PureComponent<
...
+    /* istanbul ignore next */
+    const stop = () => e.preventDefault() && e.stopPropagation()
+
+    // Escape route tested
+    /* istanbul ignore if */
+    if (code === key.ESCAPE) {
+      stop()
+      this.textArea.current.blur()
+    }
+  }
```
knicklabs: This functionally isn't changing anything but it's easier to when integrating the component with other components listening to escape to prevent actions bound to escape from firing based on the state of the component if we delay blurring the input to the keyup event.

knicklabs:
@tinkertrain I added you for a second review since I ended making more changes here beyond what Terrence originally made.

src/components/EditableTextarea/EditableTextarea.tsx
```diff
@@ -96,7 +98,6 @@ export class EditableTextarea extends React.PureComponent<
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.value === this.props.value) return
```
knicklabs: Short circuiting here prevents resetting the component to its original state which is sometimes necessary. As such, I removed this guard.

PR #737, opened 2019-10-31
Modify default color for Avatars

TerrenceLJones:
[Jira Ticket ](https://helpscout.atlassian.net/browse/COMMS-353)

This changes the default color for the Avatar component, when an image is not available.

Screen Shot 2019-10-31 at 17 36 27


src/components/Avatar/styles/Avatar.css.ts
```diff
@@ -251,6 +251,8 @@ export const OuterBorderUI = styled(CropBorderUI)`
 `
 
 function getColorStyles(props: Object): string {
+  console.log(props)
```
tjbo: Can we clean this up?

PR #738, opened 2019-11-01
Make Dropdown.Item polymorphic

knicklabs:
The `Dropdown.Item` component is a `div` but it may also be handy to make it an `a` tag. The reason for this is that if you want to redirect the user to another page you have to currently handle that with a click event. This means that you have to manually handle opening the page in a new page and it is not possible to have the "Open in a New Tab" context menu item on right click of the item. If we use an `a` tag for such items, then we get all of this functionality and more for free from the browser. We should take advantage of what the browser offers us whenever possible.

This is a very small change. If an item has an `href` attribute, the `Dropdown.Item` will be rendered with an `a` tag instead of a `div`. No styles or anything else changes. See the Stories for `DropdownV2/Item/Link` and `DropdownV2/Item/Disabled Link` for examples.

### DIFF
--- a/src/components/Dropdown/V2/Dropdown.css.js
+++ b/src/components/Dropdown/V2/Dropdown.css.js
@@ -125,73 +125,85 @@ export const ActionContentUI = styled('div')`
   max-width: 100%;
 `
 
-export const ItemUI = styled('div')`
-  ${baseStyles};
-  background-color: ${rgba(getColor('grey.400'), 0)};
-  border-radius: 0 !important;
-  color: ${getColor('charcoal.400')};
-  cursor: pointer;
-  display: block;
-  outline: none;
-  position: static;
-  transition: background-color 0.1s ease;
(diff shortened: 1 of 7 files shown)


src/components/Dropdown/V2/Dropdown.css.js
```diff
@@ -125,74 +125,86 @@ export const ActionContentUI = styled('div')`
...
+    [SELECTORS.itemAttribute]: true,
   }
-`
-ItemUI.defaultProps = {
-  [SELECTORS.itemAttribute]: true,
+
+  return ItemUI
 }
 
+export const ItemUI = makeItemUI('div')
```
knicklabs: In case anyone was using this outside of our library, I have continued to export `ItemUI` as a `div` which is backwards compatible.

knicklabs:
To use beta release: `npm i @helpscout/hsds-react@dropdown`

PR #740, opened 2019-11-05
Allow item props to be updated on validate/save

knicklabs:
We use the validate method to do validation and saving since it is async. It may be necessary to update other item properties on save, e.g. to add an id property on create so the next save will update.


knicklabs:
@tinkertrain for full details about why I'm doing this, see this PR: https://github.com/helpscout/hs-app/pull/7565

knicklabs:
I just cleaned up some cruft from the beta release. Will merge and release once this is green.

PR #745, opened 2019-11-07
Change way Dropdown ItemUI is created

knicklabs:
In https://github.com/helpscout/hsds-react/pull/738 we made a change so that a Dropdown.Item could be rendered as an `a` tag instead of a `div` if the item had an `href` prop.

This worked fine, except in Chat UI. The items were being rendered as `div` components still, but most of the time, required two clicks for the `onClick` handler to fire.

To be honest, I'm not sure why this happened. The only thing we did here was change `ItemUI` to a factory that returns the ItemUI as a `div` or `a` tag. 

Nonetheless, reverting that fixed the problem. We can use the `withComponent` API method on a Fancy StyledComponent to dynamically change the div to an `a` when there is an `href` prop and this works.

In essence, we're doing the same thing, but one solution causes problems and the other does not.

### DIFF
--- a/src/components/Dropdown/V2/Dropdown.Item.tsx
+++ b/src/components/Dropdown/V2/Dropdown.Item.tsx
@@ -10,7 +10,7 @@ import Menu from './Dropdown.Menu'
 import {
   ActionUI,
   ActionContentUI,
-  makeItemUI,
+  ItemUI,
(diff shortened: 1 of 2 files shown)


knicklabs:
I tested this with 2.70.5-0 beta release.

PR #748, opened 2019-11-11
Allow Item Props to be Updated on Validate/Save on Input Blur

knicklabs:
This update is a follow-up to: https://github.com/helpscout/hsds-react/pull/740

> We use the validate method to do validation and saving since it is async. It may be necessary to update other item properties on save, e.g. to add an id property on create so the next save will update.

In the previous PR, this was fixed but it only worked when pressing enter. It should also work when blurring the field.

### DIFF
--- a/src/components/EditableField/__tests__/EditableField.test.tsx
+++ b/src/components/EditableField/__tests__/EditableField.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { cy } from '@helpscout/cyan'
-import { mount } from 'enzyme'
+import { mount, shallow } from 'enzyme'
 import { EditableField } from '../EditableField'
 import {
   ACTIONS_CLASSNAMES,
@@ -249,7 +249,7 @@ describe('Value', () => {
     expect(spy).toHaveBeenCalledWith({ event, name, value: initialValue })
   })
 
-  test('handleInputKeyDown enter: commits value', () => {
+  test('handleInputKeyDown enter: commits value', done => {
     const wrapper: any = mount(<EditableField name="company" value="hello" />)
     const initialValue = wrapper.state('fieldValue')
     const name = initialValue[0].id
@@ -274,14 +274,14 @@ describe('Value', () => {
         {
           id: name,
           value: 'howdy',
-          disabled: false,
           validated: true,
         },
       ])
-    })
+      done()
+    }).catch(done)
   })
 
-  test('handleInputKeyDown enter: commits value (with options)', () => {
+  test('handleInputKeyDown enter: commits value (with options)', done => {
     const wrapper: any = mount(
       <EditableField
         name="company"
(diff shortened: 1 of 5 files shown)


src/components/EditableField/__tests__/EditableField.test.tsx
```diff
@@ -1161,35 +1162,7 @@ describe('enter press', () => {
...
-    input.simulate('keydown', { key: 'Enter' })
-
-    const f = flushPromises()
-    jest.runAllImmediates()
-
-    f.then(() => {
-      expect(escapeSpy).toHaveBeenCalled()
-      expect(stateSpy).toHaveBeenCalledWith({ disabledItem: [] })
-    })
-  })
```
knicklabs: @tinkertrain None of these tests were finishing before because we weren't calling `done()` in the callback. I fixed all of the tests up, but can't figure out what we should do with this test here so I've removed it for now. Any idea?

src/components/EditableField/__tests__/EditableField.test.tsx
```diff
@@ -1592,35 +1576,46 @@ describe('Input Blur', () => {
...
-    input.getDOMNode().value = '1234567'
     input.simulate('blur')
-
     const f = flushPromises()
     jest.runAllImmediates()
-
     f.then(() => {
-      expect(escapeSpy).toHaveBeenCalled()
-      expect(stateSpy).toHaveBeenCalledWith({ disabledItem: [] })
-    })
```
knicklabs: @tinkertrain same issue as mentioned above, so I've removed it for now.

knicklabs:
I also fixed up a bunch of tests that were not executing to completion before.

src/components/EditableField/__tests__/EditableField.test.tsx
tinkertrain: Hi! I'm having problems understanding what needs to happen here, running the test as is seems to be running the assertions but I think I'm wrong on some assumption here, considering this and the next tests are rejected...

knicklabs: @tinkertrain The assertions actually aren't running synchronously, so they do fail, but don't break the test because they fail after it finishes. If we add the `done` argument to the test and invoke it in the `f.then` block, then the assertions will run within the test and fail. I cannot for the life of me, get that spy to be called. However, I can verify that it is reached.

tinkertrain: Oh I see, I did try several combinations too, within the `catch` or the second argument on `then` but couldn't make it work either, it's fine by me to delete or comment them out :)

tinkertrain: Maybe leave at least one and comment out and leave a comment of why they're commented? might be usefult at some point in another instance that is similar

knicklabs: Sounds good. I think I'll just leave them deleted then for the same of not having commented out stuff here. We can always refer back to this PR. There is coverage around escaping and the test coverage is not impacted by this removal.

knicklabs:
Once this is green again, I'll merge and release.

PR #749, opened 2019-11-12
HSDS-81: DotStepper Component

Charca:
**[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-81)**

## Summary

This PR adds a new `DotStepper` component, to be used initially in the [Beacon Onboarding](https://paper.dropbox.com/doc/Spec-Beacon-Onboarding--AocVju31C9fWt2jpQ49a1mjUAg-ZoA8t4X2J0XPvruI9cyOa) project.

Since this component is a pretty simple one, we've decided to build it from scratch instead of using the `remake:component` script to bootstrap it, so we can use it as an example of how we want to build components in hsds-react. In particular, what's different about this component is that:

- It doesn't use TypeScript. Just regular JS here.
- No PropConnect.
- Uses the [new folder structure](https://c.hlp.sc/5zu5obqo) with stories, tests, and styles in the same directory as the component code.

Please feel free to comment on any aspect of how the component was built. Happy to use this PR to start discussions around styling and conventions we'd like to keep as a group.

## Design Review

@digitaldesigner I've built the component based on [this prototype](https://codepen.io/buzzusborne/live/7550e34ce040b68b0b1a1f8bcb0dacc2) but since the colors didn't match the ones in colorway, I had to use slightly different ones (`green.300`, `green.500` and `grey.500`)

## Screenshot

image


digitaldesigner:
Perfect, thanks @Charca — yeah always default to colorway, my prototypes can't be trusted . Nice work :p

Charca:
@ryan-mulrooney yeah I'm not sure why knobs are working the first time you change them only :/ It seems to be broken in the `master` branch as well. @tinkertrain @plbabin any ideas?

plbabin:
> Wooo no TS! I like the simplicity, looks great.
> 
> One question, what do you (everyone) think about `propTypes`? Useful, not useful, keep, get rid of...

🤷‍♂ They never catch anything. I see them as documentation more than type validation

Charca:
Good question for the JS Guild! I'll add it to the board.

PR #750, opened 2019-11-12
Global styles & cleaning up SCSS

plbabin:
**HSDS V3 release**
_Still a work in progress_

This second V3 PR introduce the concept of global stylesheets within HSDS-React. By moving away from fancy we can now leverage features from [Styled Components](https://styled-components.com) and that includes [Create Global Style](https://www.styled-components.com/docs/api#createglobalstyle).

Not only this update will make our life easier to make sure HSDS-react code all look the same everywhere, it will also reduce the noise in your Dev Tool of choice when inspecting styles

Image 2019-11-12 at 5 04 05 PM

# Global Styles
We add a new component `<GlobalStyle />` that can be injected anywhere inside a react application. When added to a page, styled-components will add a new `<style>` tag with the content of our `<GlobalStyle />` component. This act as a CSS stylesheet so anything in it can target anything on a page. Right now we are using it to set our global css variable for font size and font family.

## Rename global css var
All css var kept referencing Blue in their name. We moved away from that name a year ago, it just makes sense to rename those variable to have HSDS in their name:

`HSDSGlobalFontFamily`, `HSDSGlobalFontFamilyMono` & , `HSDSGlobalFontSize`.

## `.hsds-react` className
Within the `<GlobalStyle />` there is css code namespaced with the `.hsds-react` className. The reason is that we don't want to have our library targetting CSS from another application. Any React application that want to leverage the full HSDS global styling will need to wrap their app with this className. Keep in mind that global font size and family will work in both situation. Later we will introduce a CSS reset within that namespace to overwrite

```javascript
export class App extends React.Component {
  render() {
    const store = createStore(preloadedState)
    return (
      <div className="hsds-react">
        <GlobalStyle />
        <ReduxProvider store={store}>
          <AppRouter>
              {this.props.children}
          </AppRouter>
        </ReduxProvider>
      </div>
    )
  }
}
```

## Overwrite font-size / font-family
There is two way to change the value for font-size and font-family inside the GlobalStyle component. 

You can first target the css variable and update it

```css
.someComponent{
  --HSDSGlobalFontFamily: "Helvetica Neue", sans-serif;
}
```

or you can pass it as a prop when injecting the component in an app:

```javascript
<GlobalStyle fontFamily="Helvetica Neue", sans-serif" />
```

# Farewell SCSS & Seed
There were leftovers SCSS -> fancy migration for some components. This update migrate all components to styled-components and remove anything related to SCSS and Seed in the app. It removes some build command, files, resets and scripts. We are now 100% under styled-components with this PR 💯

# Sunsetting Button V1
The old button component was removed from HSDS-react. Now there is only one version (all reference to `ButtonV2` will need to be renamed to `Button`

# Stylelint
This update add a new linter for all our Component.css.js file. It used Stylelint and that linter will trigger Prettier when needed.

# `npm run storybook:ci`
Sick of your storybook command opening a new browser window each time you start the process. We got you covered. Use `npm run storybook:ci` to run storybook in the background, and just refresh an already opened window to start testing

# What's left to do
- [x] Update Documentation
- [x] Fix tests 
- [ ] ~Move `linkStyles` to GlobalStyles~
- [ ] ~HSAPP css reset in GlobalStyles~


src/components/ConditionField/styles/ConditionField.css.ts
```diff
@@ -14,19 +11,12 @@ export const FieldCloseWrapperUI = styled(Flexy.Item)`
...
 
 export const AddButtonContentUI = styled('div')`
-  ${baseStyles};
   padding: 5px 0;
 `
 
-export const OrWrapperUI = styled('div')`
-  ${baseStyles};
-`
+export const OrWrapperUI = styled('div')``
```
Charca: There are a lot of these empty `styled('div')` components, which seem to be mostly wrappers. Do you think we can remove these now?

plbabin: Nice catch. I went ahead in 8fa835ed and replace all empty styled components

src/components/ProgressBar/styles/ProgressBar.css.js
```diff
@@ -0,0 +1,76 @@
+import styled from 'styled-components'
+import { getColor } from '../../../styles/utilities/color'
+import forEach from '../../../styles/utilities/forEach'
+
+/*
```
TerrenceLJones: Should this commented out code be removed?

src/components/StatusBar/styles/StatusBar.css.js
```diff
@@ -0,0 +1,100 @@
...
+import Centralize from '../../Centralize'
+
+const statusStyles = {
+  error: 'red',
+  info: 'blue',
+  success: 'green',
+  warning: 'yellow',
+}
+
+const statusBarStyleBold = color =>
```
TerrenceLJones: What are your thoughts on refactoring these to have explicit returns (as we are doing here: https://github.com/helpscout/hsds-react/pull/750/files#diff-8e4af190189ebd03f97d900f0df4fe5fR38)? The backticks here combined with the implicit return initially confused me as I thought they were styled components.

src/components/StatusBar/styles/StatusBar.css.js
```diff
@@ -0,0 +1,100 @@
...
+        &:focus {
+        box-shadow: 0 0 0 2px ${rgba('white', 0.8)};
+        }
+    }
+    `
+
+const makeStatusStyles = () => {
+  return forEach(
+    statusStyles,
+    (name, value) => `
```
TerrenceLJones: The same question here as above.

src/components/StatusBar/styles/StatusBar.css.js
plbabin: I did update the code to do an explicit return here 0c7afaa5

src/components/ProgressBar/styles/ProgressBar.css.js
plbabin: all done in 0c7afaa5

src/components/StatusBar/styles/StatusBar.css.js
plbabin: I did update the code to do an explicit return here 0c7afaa5

plbabin:
I will go ahead and merge that in the v3 branch. There was two things in my original PR comment that I was listing as "todo". First was to move linkStyle() to the global stylesheet. I am not sure it's a good idea for now, it increase the risk of changing something somewhere in a component that we didn't want to

Second thing was to add a css reset for HSAPP. I will defer that for now. We need a way to increase specificity on all selectors otherwise our reset code will overwrite some component code. We also need to carefully select which style we want to reset without causing too much issue on our elements. I am open to ideas on how why should tackle those styles reset

PR #753, opened 2019-11-18
HSAPP-734 Create a Form component

cen10:
[JIRA](https://helpscout.atlassian.net/browse/HSAPP-734)

## Description
This PR adds a Form component with a primary Save action and an optional Cancel and Delete action. It is heavily inspired by the actions that can be used with the existing Page component. It is essentially a wrapper around whatever content you provide it, and always places the buttons below the content. The buttons can be displayed left to right or right to left.

Note: Typescript isn't used, as we're plannign on removing it, and this follows the new folder structure described by Maxi here: #749 . After speaking to Maxi, he suggested that sub-components like `Form.Action` could go into their own folders, so that's the format I've followed here.

Screen Shot 2019-11-21 at 1 36 35 PM

## Testing
Buttons will look like:
Screen Shot 2019-11-18 at 3 37 02 PM

https://deploy-preview-753--hsds-react.netlify.com/?path=/story/form--default

* [ ] Visit the Form story and test that you can see the console logged results when clicking each of the  buttons. 
* [ ] Using the Storybook knobs, you should be able to change the direction of the buttons from the default of right to left to the reverse by typing "left" in the `actionDirection` field.
* I couldn't figure out a way to use knobs to show that the save button will have the default text "Save" if no text is provided (it renders an empty string if you leave the field blank), but devs can remove that prop locally to see that it works.


nikkiathelpscout:
Just reviewing! This looks great, I like that you can flip it too. The only thing I'd want to confirm is hitting delete would have the option to throw a modal to delete the object, not just the text within the field. That's probably the case, just wanted to confirm though.

cen10:
@nikkiathelpscout Yep, that's just for the purpose of having something to do in Storybook. We could have it open a modal.

src/components/Form/Form.js
```diff
@@ -0,0 +1,82 @@
...
+        kind="primary"
+        size="lg"
+        version={2}
+        submit={true}
+      >
+        {saveText}
+      </Button>
+    )
+
+    const cancelButton = cancelText && (
```
knicklabs: Should we have saveText, cancelText, and deleteText have default props and instead only render these buttons if the actions (e.g. onSave, onCancel, and onDestroy respectively) are bound?

knicklabs: It might be worth keeping a default value for onSave just because the utility seems to arise from using a form with at least one of the actions.

cen10:
@knicklabs Updated!

cen10:
Merging as I got the OK to do so in Slack from Juan Pablo:
Screen Shot 2019-11-21 at 1 39 07 PM

PR #754, opened 2019-11-18
Upgrade node to v12.13

plbabin:
This update move the needle to 12.13.0 for the node version that is use to install packages.


plbabin:
@helpscout/javascripters-js Test are still not passing, it's related to the async way of loading svg from a babel plugin. The loading is timing out and causing an error to be thrown. Weird thing is that it's working totally fine when building or running storybook

I tried moving to https://github.com/airbnb/babel-plugin-inline-react-svg but we can't namespace our icon, and it requires refactoring how we inject icons. https://github.com/smooth-code/svgr could be a solution, but it would require running a script to build to icon, or force a svg loader in webpack

I am open to ideas on ways we could resolve the tests not working at all right now with the actual plugin

knicklabs:
🤔

plbabin:
Closing the PR. In v3 we did update to node 12.13 and refactors Icon to use https://github.com/airbnb/babel-plugin-inline-react-svg.

PR #756, opened 2019-11-19
Fix avatar selector border

tjbo:
Fixs the color and adjusts the border gap of the AvatarSelector


ryan-mulrooney:
@tjbo looks good to me!

PR #759, opened 2019-11-25
Update Dropdown Component `window` reference

TerrenceLJones:
This adds support for a `customWindow` prop in the Dropdown Component. 

This is needed to allow for use of the component in Beacon. Because Beacon is shown in an `iframe` element, we need to be able to override the `window` reference in the component to use the `iframe`'s `window` object and not the `host`'s.

### Before 
Screen Recording 2019-11-25 at 16 08

### After

Screen Recording 2019-11-25 at 15 14

Testing:

- That no regressions were introduced into the Dropdown V2 functionality
- That the scrolling issue displayed in the gifs above are fixed. You can check out this [PR](https://github.com/helpscout/beacon2/pull/771) which has a  beta release of the this work installed.

### DIFF
--- a/src/components/Dropdown/V2/Dropdown.Item.tsx
+++ b/src/components/Dropdown/V2/Dropdown.Item.tsx
@@ -28,6 +28,7 @@ import ItemSelectedCheck from './Dropdown.ItemSelectedCheck'
 export interface Props {
   actionId?: string
   className?: string
+  contentWindow: any
   disabled: boolean
   dropRight: boolean
   dropUp: boolean
@@ -54,6 +55,7 @@ export interface Props {
 
 export class Item extends React.PureComponent<Props> {
   static defaultProps = {
+    contentWindow: window,
     getState: noop,
     disabled: false,
     index: '0',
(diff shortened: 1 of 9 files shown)


src/components/Dropdown/V2/Dropdown.Item.tsx
```diff
@@ -54,6 +55,7 @@ export interface Props {
 
 export class Item extends React.PureComponent<Props> {
   static defaultProps = {
+    containerWindow: window,
```
TerrenceLJones: This default window prop is required for testing.

src/components/Dropdown/V2/Dropdown.MenuContainer.tsx
```diff
@@ -45,6 +45,7 @@ export class MenuContainer extends React.PureComponent<
     animationDuration: 80,
     animationSequence: 'fade down',
     closeDropdown: noop,
+    containerWindow: window,
```
TerrenceLJones: Required for testing.

PR #760, opened 2019-11-26
Table collapse and custom expander text

tinkertrain:
This PR adds the ability for the Table to be collapsed after it has been expanded and use custom text for the "expander" button.

Screen

📚 [Story](https://deploy-preview-760--hsds-react.netlify.com/?path=/story/table--expandable)

### DIFF
--- a/src/components/Table/Table.Body.tsx
+++ b/src/components/Table/Table.Body.tsx
@@ -10,7 +10,10 @@ export default class Body extends React.Component<BodyProps, BodyState> {
   columnsCache = this.getColumnsCache(this.props.columns)
 
   static getDerivedStateFromProps(nextProps, prevState) {
+    /* istanbul ignore else */
     if (!equal(nextProps.rows, prevState.rows)) {
+      return { rows: Body.getRows(nextProps) }
+    } else if (nextProps.isTableCollapsed) {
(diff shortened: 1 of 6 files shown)


src/components/Table/Table.Body.tsx
```diff
@@ -10,8 +10,11 @@ export default class Body extends React.Component<BodyProps, BodyState> {
   columnsCache = this.getColumnsCache(this.props.columns)
 
   static getDerivedStateFromProps(nextProps, prevState) {
+    /* istanbul ignore else */
     if (!equal(nextProps.rows, prevState.rows)) {
       return { rows: Body.getRows(nextProps) }
+    } else if (nextProps.isTableCollapsed) {
+      return { rows: Body.getRows(nextProps) }
     }
```
plbabin: since they both do the same actions, should we just have one `if` statement?

```suggestion
    const rowsAreNotEquals = !equal(nextProps.rows, prevState.rows) 
    if (rowsAreNotEquals || nextProps.isTableCollapsed) {
      return { rows: Body.getRows(nextProps) }
    }
```

tinkertrain: Ah yeah nice catch :)

PR #761, opened 2019-12-04
Optionally Prevent Focus on Form Actions

knicklabs:
[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-117)

This update adds a new `actionFocusable` property to the `Form` component. By default, this property is set to `true` and has no impact on the current functionality of the form. When set to `false` it adds `tabindex="-1"` to each of the form action buttons, preventing the user from being able to focus on them by pressing tab. This is sometimes required by consuming code if, for example, a form remains visible (for animation purposes) while being hidden, and thus a user cannot see the form and should not be able to interact with it.

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -7,6 +7,7 @@ export class Form extends React.PureComponent {
   static Actions = Actions
 
   static defaultProps = {
+    actionFocusable: true,
     cancelText: 'Cancel',
     destroyText: 'Delete',
     onSave: evt => {
@@ -18,6 +19,7 @@ export class Form extends React.PureComponent {
   render() {
     const {
       actionDirection,
+      actionFocusable,
       cancelText,
       children,
       className,
@@ -29,6 +31,7 @@ export class Form extends React.PureComponent {
     } = this.props
 
     const componentClassName = classNames('c-Form', className)
+    const commonButtonProps = actionFocusable ? {} : { tabIndex: -1 }
 
     const saveButton = (
       <Button
@@ -37,6 +40,7 @@ export class Form extends React.PureComponent {
         size="lg"
         version={2}
         submit={true}
+        {...commonButtonProps}
       >
         {saveText}
       </Button>
@@ -48,6 +52,7 @@ export class Form extends React.PureComponent {
         size="md"
         version={2}
         onClick={onCancel}
+        {...commonButtonProps}
       >
         {cancelText}
       </Button>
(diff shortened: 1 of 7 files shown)


src/components/Form/Form.js
```diff
@@ -70,6 +76,7 @@ export class Form extends React.PureComponent {
         {children}
         <Form.Actions
           direction={actionDirection}
+          focusable={actionFocusable}
```
cen10: Is this doing anything? Form.Actions isn't expecting a new prop, and I took it out when I was playing in Storybook and the buttons still became unfocused.

src/components/Form/Form.stories.js
```diff
@@ -206,3 +206,7 @@ stories.add('with only save button', () => {
 
   return <SampleForm />
 })
+
+stories.add('with unfocusable buttons', () => (
+  <Form actionFocusable={false} onCancel={() => {}} onDestroy={() => {}} />
```
cen10: 1. What do you think about reusing the previous story with a couple tweaks:
```
stories.add('with unfocusable buttons', () => {
  class SampleForm extends React.Component {
    state = {
      text: 'Sample Text',
    }

    handleChange = value => {
      this.setState({
        text: value,
      })
    }

    handleFormSubmit = evt => {
      evt.preventDefault()
      console.log(`Submitting text: "${this.state.text}"`)
      this.setState({ text: '' })
    }

    render() {
      return (
        <ContainerUI>
          <Form
            actionFocusable={boolean('actionFocusable', false)}
            onSave={this.handleFormSubmit}
            saveText="Save Entry"
          >
            <FormGroup>
              <FormLabel label="Site Name">
                <Input onChange={this.handleChange} value={this.state.text} />
              </FormLabel>
            </FormGroup>
          </Form>
        </ContainerUI>
      )
    }
  }

  return <SampleForm />
})
```

It's the same exact story as the previous one in the file, but minus the `actionDirection` prop and adding a knob. I think this makes it more easy to tell that focus is missing, because it clearly starts in the input.

2. In playing around with that story, it's clear that you can still click the Save button and console log some text. I think this is useful to include to drive home that we're not disabling the buttons entirely, just adding a tab index. Do you think this could lead to confusion though, and a name like "actionIsTabbable" while... interesting, would be more clear? I don't have a strong opinion here so happy to keep as is.

knicklabs:
Great feedback @cen10! I changed the prop name and updated the story.

knicklabs:
When it's green, I'll merge it and make a new release.

PR #763, opened 2019-12-16
Input Error and Tooltip Accessibility

knicklabs:
*[HSDS-114](https://helpscout.atlassian.net/browse/HSDS-114)*

This update improves the accessibility of errors on the Input component as well as generally improves the accessibility of the Tooltip component.

The Tooltip component was updated to:

1. When the Tooltip is set to open on hover and close on mouse leave, the Tooltip will also open on focus and close on blur.
2. When the Tooltip is set to open on click and close on click, the Tooltip will also open on enter press and close on enter press. Clicking or blurring will also close an open Tooltip.

Try the Tooltip story [https://deploy-preview-763--hsds-react.netlify.com/?path=/story/tooltip--default](https://deploy-preview-763--hsds-react.netlify.com/?path=/story/tooltip--default) and change the knob configurations to see the impact the various settings have.

Also try the Input (state:error) story [https://deploy-preview-763--hsds-react.netlify.com/?path=/story/input--state-error](https://deploy-preview-763--hsds-react.netlify.com/).

Finally, try the Input (suffix) story [https://deploy-preview-763--hsds-react.netlify.com/?path=/story/input--suffix](https://deploy-preview-763--hsds-react.netlify.com/?path=/story/input--suffix)

Note: Because we are making pop accessibility, popover also inherits these same features that tooltip is inheriting. I updated the popover story to demo it as well.

https://deploy-preview-763--hsds-react.netlify.com/?path=/story/popover--default

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.76.0",
+  "version": "2.76.1-1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.76.1-2",
+  "version": "2.76.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.76.1-1",
+  "version": "2.76.1-2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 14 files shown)


knicklabs:
@nikkiathelpscout In addition to the accessibility improvements, I've fixed the Input so that the error icon is displayed to the left of the suffix.

We provide a suffix to be rendered, so I was thinking we should rely on the user to handle providing the suffix with an error state as necessary. 

Unfortunately, I couldn't recreate your mockup with what we have. The secondary button has no error state. Furthermore, the disabled state of the button does not use the "not allowed" icon.

Do you think we should make further changes to the Button V2 in HSDS or should we handle these with custom styles in-app? 

See: http://c.hlp.sc/913a80149b5e

knicklabs:
WIP label on this as I'm discussing some visual changes with Nikki.

knicklabs:
@Charca I tagged you for review as I know you are using this in Beacon. Want to make sure this change isn't going to adversely impact you in anyway. The goal here is to make errors on inputs accessible.

knicklabs:
I carded up the Button focus issue here https://helpscout.atlassian.net/browse/HSDS-119

PR #764, opened 2019-12-17
Reaction Emoticons update

tinkertrain:
The reactions (happy/sad) emoticons have been updated by Jared:
screenshot

This PR updates the emoticons inside HSDS React.

While doing so, I noticed that we have `<Emoticon />` and `<RateAction />` components and their intended functionality was a bit mixed, specifically some stuff was on Emoticon that should have been on RateAction.

So I refactored both components so that:

- `Emoticon` is like `Icon`, but for svgs that have more than one colour (so shades do not apply for example)
- `RateAction` uses `Emoticon` under the hood, as before, but it takes care of the animation and interactions
- `RateAction`: added the "okay" animation
- `RateAction`: added all 3 sizes
- `RateAction`: I noticed that keyboard usage (tabbing in and out) of the action was weird: focusing on the action button would automatically select it, likewise blurring would deselect it, this behaviour is not a normal behaviour for a button. I fixed this by only selecting an action on click/enter, and having the styles adjusted for focus/blur.
- Removed TypeScript on both
- Added `prop-types`
- Uses JSDocs where appropriate
- Restructured folders to recently adopted pattern
- Added new stories for both
- Updated docs
- Maintains backwards compatibility
- Removes 'propConnect' from both

## Test

📚 [Emoticons](https://deploy-preview-764--hsds-react.netlify.com/?path=/story/emoticon--default)
📚 [RateAction](https://deploy-preview-764--hsds-react.netlify.com/?path=/story/rateaction--active)

🧪 On RateAction, try to use the dropdown I provided, this tests that the buttons are able to update from external props.
🧪 Also, use `tab` and `enter` to try the actions

### Emoticons
All

### RateAction

Reactions


src/components/Emoticon/Emoticon.js
```diff
@@ -0,0 +1,105 @@
...
+          className="c-Emoticon__icon"
+          dangerouslySetInnerHTML={src}
+          title={title}
+        />
+      </EmoticonUI>
+    )
+  }
+}
+
+Emoticon.propTypes = {
```
plbabin: any particular reason why the propTypes declaration is not inside the component?

src/components/RateAction/RateAction.css.js
```diff
@@ -0,0 +1,186 @@
...
+import { reactionEmoticonsColours } from '../Emoticon/Emoticon.css'
+
+export const config = {
+  size: {
+    default: '28px',
+    lg: '28px',
+    md: '24px',
+    sm: '20px',
+  },
+  boxShadowColor: '193, 203, 212', // grey.600
```
plbabin: there is a rgba function in the color utils. I think you could used that to keep rbg color out

it works like the sass method from what I recall `rgba(color, opacity)`

src/components/RateAction/RateAction.js
```diff
@@ -1,41 +1,21 @@
 import * as React from 'react'
```
plbabin: it feels good to cleanup the code right? 🥇

src/components/Emoticon/Emoticon.css.js
```diff
@@ -0,0 +1,81 @@
...
+  },
+  on: {
+    face: '#2E2D2E',
+    head: {
+      happy: '#FFDA6A',
+      okay: '#D2D2D2',
+      sad: '#BDD6FF',
+    },
+  },
+}
```
knicklabs: I see you are using colours outside of our palette here, except #A5B2BD which happens to be grey.800. I do wonder, if we should leave a comment here that we are using custom colours so no one comes along like I just did to check them against our color palette thinking we should have used getColor here to select these values from our palette. Thoughts?

jaredmcdaniel:
@plbabin not sure i follow your comment on the focus state issue? You dont think the green border is clear?

plbabin:
> @plbabin not sure i follow your comment on the focus state issue? You dont think the green border is clear?

green border is for the selected state. I thought the focus was only increasing the size, but it is the size, color and animation. I am fine with that https://c.hlp.sc/4guxJje2. 

Disregard my comment 😄

jaredmcdaniel:
@plbabin You're right ... my bad. But yea, think we're okay there. 👍

PR #766, opened 2020-01-08
Tweak copy button styles

tjbo:
This updates the copy button styles for HSAPP-995

### DIFF
--- a/src/components/CopyButton/CopyButton.tsx
+++ b/src/components/CopyButton/CopyButton.tsx
@@ -15,7 +15,6 @@ export interface Props {
   canRenderFocus: boolean
   children?: any
   className?: string
-  icon: string
   kind: string
   onClick: (event: Event) => void
   onReset: () => void
@@ -31,8 +30,6 @@ export interface State {
 class CopyButton extends React.PureComponent<Props, State> {
   static defaultProps = {
     canRenderFocus: false,
-    icon: 'copy',
-    iconShade: 'faint',
     onClick: noop,
     onReset: noop,
     kind: 'tertiary',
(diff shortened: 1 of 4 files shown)


tjbo:
@digitaldesigner , I enjoy your prototypes, I really do. 👍  

They just confuse the heck out of me, when combined with the mock ups in Figma and also HSDS-React (being different and having 10 options for something that we only use one version of and probably needs 0 options, but that's a different story). 🙃 

https://c.hlp.sc/4bdf269b9a0a

Hopefully you can help me get to the bottom of this, I've pushed the first commit and we can figure it out from here:

#1 - copy button hover color, is that new, should it apply to the ButtonV2 component default state as well, ie all Buttons with default state in HSAPP, or just specific to this component?

#2 - background color of input, to disable or not to disable?

#3 - focus border color, to use or not to use?

digitaldesigner:
Okay, I've confused you with my prototypes — that's my bad. So for the interactions, ignore the prototype completely, this is the reference: https://www.figma.com/file/WM0YJJK0Zbbby2Eu2domSC/Forms-Encapsulated-Buttons?node-id=1%3A302

So in summary, you're basically done — light blue background is the intended hover. Also if you could make the top-left and bottom-left have no rounded corners, that'd be great too (https://c.hlp.sc/o0uQRvB4)

tjbo:
@digitaldesigner , thanks for clearing up the confusion:

Updated here:
http://c.hlp.sc/c40462458553

And it updates the hover color for this one too (but not border radius):
http://c.hlp.sc/e56e1f0fa7b8

src/components/CopyButton/CopyButton.tsx
```diff
@@ -87,36 +84,19 @@ class CopyButton extends React.PureComponent<Props, State> {
     this.props.onClick(event)
   }
 
-  isIconOnly = () => {
```
tjbo: I removed some of this stuff as we only have one version of this button that we use everywhere.

src/components/CopyInput/CopyInput.tsx
```diff
@@ -68,6 +68,7 @@ class CopyInput extends React.PureComponent<Props> {
             onClick={this.handleCopyClick.bind(this)}
             size="lg"
             isLast
+            style={{ borderTopLeftRadius: 0, borderBottomLeftRadius: 0 }}
```
tjbo: These radii are only used in the Input, not in the TextArea version.

PR #767, opened 2020-01-09
Tweak page card bottom margin

tjbo:
Change bottom margin on page card to 20px. This is a global change.


tjbo:
Looks like this:
http://c.hlp.sc/16695c7a772a

PR #769, opened 2020-01-14
Removing Props Provider

plbabin:
This is happening! This update remove entirely Props Provider from HSDS react.

# Why

Long story short, Props Provider was a way to update props within a react context like API. It was meant to pass a value to a children to overwrite the actual component props value.

In theory it was amazing, in practice it was complex to maintain and implement. Sometime a prop was overwritten without really knowing where it was coming from, creating some confusion in the behavior of the component. And then sometime just implementing a new component with all the boilerplate to make them work was too much hassle for little gain.

Last point was performance wise. Since we allowed all components to be overwritten, it meant that all components was wrapped with a context API polyfill, create a huge component tree for again, little gain.

It was time to modernize the stack

# How
But, `I do want to modify some component without passing a prop` you asked! It's still possible, but in a limited manner. Instead of having all components updatable, we expose only the one that would really need to be update from outside HSDS. 

For example, you could need to make sure Tooltip has a z-index different everywhere in the app:

```javascript
import Tooltip, {TooltipContext} from './Tooltip'

<TooltipContext.Provider value={{zIndex: 100}}>
  {application}
</TooltipContext.Provider>
```

For internal usage, we could need to pass information down to children for some components. For example, TagList would need to pass down the `isRemovable` to all Tag children. Each `Tag` component use the TagListContext that will pull the `isRemovable` value.

Everytime we need to pass down information to unknown children, we will leverage and expose a react context (TagList, Modal, Card, Page etc).

# How does it impact me as a developer?

Primo, if your app is using PropProvider you will have to refactor your components to either use the new Context.Provider from the component, or rethink about the whys you were overwritting the prop

Secondo, when creating a new HSDS component, you don't need to add and wrap a COMPONENT_KEY anymore with propConnect. Isn't that cool?

# What's else?
Since some component had to be refactor (as a functionnal component to use context and hooks), we remove TypeScript while being there. We'll remove entirely TypeScript in the near future, but we thought we might as well do it right now for those component


plbabin:
We need to create a testing plan but it will be for the full v3. 

We’ll need to update some stuff in both hs-app and beacon

Charca:
Late to the party, but great work on this @plbabin! Excited to take v3 for a spin!

PR #773, opened 2020-01-24
Verification Code

tinkertrain:
New component!

https://www.figma.com/file/LM9TDdqD85RBXvz6EipJL6/Forms-Verification-Code?node-id=0%3A1

The verification code component is designed to be used for 2FA. The _feel_ of it is as if you are using a normal `<input />`, but the look is quite different.

For this component, some decisions of note I went with:
- Avoid usage of React State
- Make use of native browser functionality as much as possible (like using `<input />`) and some less known css features

Here's a video with more details that goes into how I went about it and why some things exist:

🎥 https://www.loom.com/share/a2a9e6aa44664445a3c922aa169daff2

## Todo
- [x] Complete test coverage
- [x] Documentation
- [x] Implement left and right arrow navigation (not in spec but I think it makes sense)

### DIFF
--- a/src/components/VerificationCode/VerificationCode.css.js
+++ b/src/components/VerificationCode/VerificationCode.css.js
@@ -1,5 +1,5 @@
 import styled from '../styled'
-import { getColor } from '../../styles/utilities/color'
+import { getColor, rgba } from '../../styles/utilities/color'
 import Icon from '../Icon'
 import Tooltip from '../Tooltip'
 
@@ -16,6 +16,8 @@ export const VerificationCodeFieldUI = styled('div')`
   border: 1px solid #c5ced6;
   border-radius: 3px;
   background-color: #fff;
+  box-shadow: 0 0 0 0 ${rgba(getColor('border'), 0)};
+  transition: box-shadow 100ms ease;
 
   *,
   *:before,
@@ -28,7 +30,7 @@ export const VerificationCodeFieldUI = styled('div')`
   }
 
   &:focus-within {
-    border: 2px solid ${getColor('blue.500')};
+    box-shadow: 0 0 0 2px ${getColor('blue.500')};
   }
 
   &.not-valid {
--- a/src/components/VerificationCode/VerificationCode.css.js
+++ b/src/components/VerificationCode/VerificationCode.css.js
@@ -33,7 +33,7 @@ export const VerificationCodeFieldUI = styled('div')`
   }
 `
 
-export const DigitInputUI = styled('div')`
+export const DigitInputWrapperUI = styled('div')`
   position: relative;
   margin: 0 15px 0 0;
 `
@@ -43,12 +43,12 @@ export const DigitMaskUI = styled('div')`
   top: 0;
   left: 0;
   width: 40px;
-  height: 35px;
-  padding: 0 0 5px 0;
+  height: 50px;
+  padding: 0 0 7px 0;
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial,
     sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
   font-size: 35px;
-  line-height: 27px;
+  line-height: 41px;
   text-align: center;
   color: ${getColor('charcoal.700')};
   border-bottom: 2px solid #d5dce1;
@@ -56,13 +56,17 @@ export const DigitMaskUI = styled('div')`
   &::selection {
     background-color: #b2d7ff;
   }
+
+  &.hidden {
+    opacity: 0;
+  }
 `
 
-export const InputUI = styled('input')`
+export const DigitInputUI = styled('input')`
   position: relative;
   width: 40px;
-  height: 35px;
-  padding: 0 0 5px 0;
+  height: 50px;
+  padding: 0 0 7px 0;
   margin: 0;
   border: 0;
   border-bottom: 2px solid #d5dce1;
@@ -85,4 +89,11 @@ export const InputUI = styled('input')`
   &.hidden {
     opacity: 0;
   }
+`
+
+export const ClipboardPlaceholderUI = styled('textarea')`
+  position: absolute;
+  top: 0px;
+  left: 700px;
+  pointer-events: none;
 `
--- a/src/components/VerificationCode/VerificationCode.css.js
+++ b/src/components/VerificationCode/VerificationCode.css.js
@@ -1,5 +1,7 @@
 import styled from '../styled'
 import { getColor } from '../../styles/utilities/color'
+import Icon from '../Icon'
+import Tooltip from '../Tooltip'
 
 export const VerificationCodeFieldUI = styled('div')`
   box-sizing: border-box;
@@ -21,10 +23,6 @@ export const VerificationCodeFieldUI = styled('div')`
     box-sizing: inherit;
   }
 
-  div:last-child {
-    margin-right: 0;
-  }
-
   &:focus {
     outline: none;
   }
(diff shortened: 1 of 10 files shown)


jaredmcdaniel:
Awesome work @tinkertrain! Got some polish items for ya so let me know if you've got any feedback on the below:

1. Can we make it where clicking _anywhere_ inside the lg "input" shows the focus states? Right now, I have to click directly into that first digit input to see the focus.

focus

2. Can the focus style follow the same patten as the default inputs by using an outer box shadow and transition? Tougher to pick up on the transition in a gif, but you can see the box-shadow focus gives the appearance of a slightly larger border radius than above.

focus-2

3. Can you give me a `-webkit-font-smoothing: antialiased;` on the text?

4. When deleting, if the current input is empty, hitting delete should go to the previous input. However, if it is filled, hitting delete should first clear the field but maintain focus. Hitting delete a second time should thenmove it. Hopefully that makes sense. 😅

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,119 @@
...
+  *:before,
+  *:after {
+    box-sizing: inherit;
+  }
+
+  &:focus {
+    outline: none;
+  }
+
+  &:focus-within {
```
cocoabythefire: this is awesome! i did not know about `focus-within`! It does look like this is not supported by IE though. Should make sure the component still looks/functions acceptably on IE.

tinkertrain: This is to be used in HS App if I'm not mistaken? We don't officially support IE there (https://docs.helpscout.com/article/1292-supported-browsers-and-system-requirements) only latest Edge, which happens to be the brand new chromium based one 🕺

And I would consider this as a "graceful degradation", so IE users still get an acceptable experience, but of course if there's a specific requirement for this component, pls let me know!

tinkertrain:
Thanks for the feedback @jaredmcdaniel!

I addressed all of it :), let me know your thoughts when you have a chance

jaredmcdaniel:
Looking great!! Last question ... possible to get the focus transition on the smaller digit input as well? I could be missing it, but seems like its only on the outer one right now.

tinkertrain:
> Looking great!! Last question ... possible to get the focus transition on the smaller digit input as well? I could be missing it, but seems like its only on the outer one right now.

Just added color change transition on focus 👍

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+    border-color: transparent;
+    box-shadow: 0 0 0 2px ${getColor('blue.500')};
+  }
+
+  &.not-valid {
+    border-color: transparent;
+    box-shadow: 0 0 0 2px #f23459;
```
plbabin: should the color be from our palette?

Based on colorway red.400 is the closest color: 
```
Finding the closest HSDS-React color for "f23459"
red.400: #f45b55
```

cc @jaredmcdaniel

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+`
+
+export const DigitMaskUI = styled('div')`
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 40px;
+  height: 50px;
+  padding: 0 0 7px 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial,
```
plbabin: mental note to update that in v3 to use `--HSDSGlobalFontFamilyMono`

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+    sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+  font-size: 35px;
+  line-height: 41px;
+  text-align: center;
+  color: ${getColor('charcoal.700')};
+  border-bottom: 2px solid #d5dce1;
```
plbabin:
```
Finding the closest HSDS-React color for "d5dce1"
grey.500: #d6dde3
```
cc @jaredmcdaniel

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+  -webkit-font-smoothing: antialiased;
+
+  &::selection {
+    background-color: #b2d7ff;
```
plbabin:
```
Finding the closest HSDS-React color for "b2d7ff"
indigo.300: #B2C8FB
```
cc @jaredmcdaniel

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+`
+
+export const DigitInputUI = styled('input')`
+  position: relative;
+  width: 40px;
+  height: 50px;
+  padding: 0 0 7px 0;
+  margin: 0;
+  border: 0;
+  border-bottom: 2px solid #d5dce1;
```
plbabin: color using our palette

src/components/VerificationCode/VerificationCode.css.js
```diff
@@ -0,0 +1,126 @@
...
+`
+
+export const ValidIconUI = styled(Tooltip)`
+  position: absolute;
+  right: 8px;
+  top: calc(50% - 24px / 2);
+`
+
+export const IconUI = styled(Icon)`
+  color: #f23459;
```
plbabin:
```
Finding the closest HSDS-React color for "f23459"
red.400: #f45b55
```

src/components/VerificationCode/VerificationCode.js
```diff
@@ -0,0 +1,292 @@
...
+    e.target.select()
+  }
+
+  render() {
+    const { numberOfChars, isValid } = this.props
+
+    return (
+      <VerificationCodeFieldUI
+        className={this.getClassName()}
+        innerRef={this.setVerificationCodeFieldNode}
```
plbabin: will need to be rename to `ref` once it's merged in v3. 

innerRef will be use for our component (otherwise it creates some conflicts with react 🤷‍♂), and ref is the keyword needed when set on html node

Also we could declare the ref here with React.createRef(). Then you would not need the `setVerificationCodeFieldNode` method.

src/components/VerificationCode/VerificationCode.js
```diff
@@ -0,0 +1,292 @@
...
+        onPaste={this.handlePaste}
+        onKeyDown={this.handleKeyDown}
+        onMouseDown={this.handleMouseDown}
+      >
+        <ClipboardPlaceholderUI
+          readOnly={true}
+          tabIndex="-1"
+          innerRef={this.setClipboardPlaceholderNode}
```
plbabin: same comment as earlier with the `ref` and v3

src/components/VerificationCode/VerificationCode.css.js
cocoabythefire: ok sounds good. I'm fine with it if it works on latest edge!

tinkertrain:
> There is one bug in the External value story. If the field is cleared the first character is still present in the input:
> Screen Recording 2020-01-27 at 11 31 AM

Oh, thanks for reporting, in fact I wanted to use knobs, and I think they're fixed on the latest release, so I'll make a note to switch it to that when we update.

src/components/VerificationCode/VerificationCode.js
tinkertrain: Yes, I started using `ref` because I remembered, but then the console reminded _me_ that fancy does not support them so I switched them

PR #774, opened 2020-01-24
Add isSortable Option to Accordion

knicklabs:
It should be simple to make an Accordion component sortable. This update adds an isSortable option to the Accordion to make that possible.

It is backed by the existing Sortable component and works with all variants of the Accordion component. You can use the knobs in Storybook to toggle isSortable on and off for every Storybook example of the Accordion component.


src/components/Sortable/Sortable.types.ts
```diff
@@ -1,22 +1,22 @@
...
+  onSortEnd?: (...args: any[]) => void
+  shouldCancelStart?: () => void
+  pressDelay?: number
+  useDragHandle?: boolean
+  useWindowAsScrollContainer?: boolean
+  hideSortableGhost?: boolean
+  lockToContainerEdges?: boolean
+  lockOffset?: number | string | Array<string | number>
+  getContainer?: () => void
+  getHelperDimensions?: () => void
```
knicklabs: None of these are required and create problems when trying to use the component internally the same way we use it externally. It is perfectly valid to make something sortable by wrapping it with `<Sortable></Sortable>` with no props.

Vesnaworx:
Tested and passed - checked every type, sorted with/without expansion, sorted with multiple expansions when available, moved expanded/not expanded to the very bottom, very top and middle.

PR #776, opened 2020-01-28
Increase Tooltip default zIndex for use in Modals

cocoabythefire:
The default zIndex for our Tooltip component should allow it to work within a Modal. Currently, you would need to manually override the z-index for a Tooltip appearing from within a Modal for it to appear.

**Before (default zIndex = 999)**
Screen Shot 2020-01-28 at 12 24 30 PM

**After (default zIndex = 9999)**
Screen Shot 2020-01-28 at 12 31 44 PM

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -43,7 +43,7 @@ export class Tooltip extends React.PureComponent<Props> {
     modifiers: {},
     placement: 'top',
     triggerOn: 'hover',
-    zIndex: 999,
+    zIndex: 9999,
   }
 
   static className = 'c-Tooltip'
--- a/stories/Modal.stories.js
+++ b/stories/Modal.stories.js
@@ -231,6 +231,19 @@ stories.add('header/footer with items', () => (
   </Modal>
 ))
 
+stories.add('with tooltip', () => (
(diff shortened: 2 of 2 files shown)


tinkertrain:
Pardon my intrusion 😁 

But is not possible to pass a `zIndex` as prop with a value that you need instead of updating the default? The prop is there already, we just need to update the documentation...

plbabin:
right that's a good point. I am fine with either way. @cocoabythefire you good update your code directly to save time, and we could make the default z-index higher in v3

cocoabythefire:
Thanks for looking at this with me! So the use case I need to solve for... is where I'm not explicitly rendering the Tooltip in a Modal. The Tooltip is rendered internally by the Input component. So I still think I would need to modify the `renderErrorMarkup` within the Input component to pass in an optional zIndex override. Unless I'm missing something...

As it stands now, no tooltip will work inside our Modal which to me seems like a bug. I'm fine with whatever is the best/easiest patch I just don't think in this case passing a zIndex prop to my input will have any effect on the tooltip it renders on error. Does that make sense @tinkertrain @plbabin

plbabin:
@cocoabythefire oh right, that totally make sense. I would go ahead with your PR, and we can still update the z-index directly on other component if needed. V3 has a React.Context that we can leverage as well to modify the z-index from anywhere

PR #777, opened 2020-01-29
Enhancement/folder structure

tinkertrain:
This PR updates the codebase to the following structure:

- All `.css.js` are located in the root component folder (previously in the `styles` folder)
- All `.test.js` are located in the root component folder (previously in the `__tests__` folder), except when there are more than **3 test files**, then we move them into the `__tests__` to reduce noise. (Not many cases when this happens)
- All `.stories.js` are located in the root component folder (previously in the `stories` folder outside the main components folder), except when there are more than **3 stories files**, then we move them into the `stories` to reduce noise. (Not many cases when this happens)
- I moved "specs" used in the stories into a `specs` folder inside the `utilities` folder
- I moved some storybook utilities into `storybook.js` inside the `utilities` folder
- Cleaned up a bit the stories
- Consolidated some tests into a single test file for some components
- Consolidated some css styles into a single css file for some components
- Tests pass
- Storybook runs


.vscode/extensions.json
```diff
@@ -2,7 +2,6 @@
   "recommendations": [
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
-    "jpoissonnier.vscode-styled-components",
-    "shinnn.stylelint"
```
plbabin: I guess we could remove styelint entirely from HSDS. From what I recall I had trouble with formatting styled-component file and I ended up with using stylehint to make it work correctly

We can always update the config too to make it better

PR #787, opened 2020-02-07
Prevent Data from Being Retained when Removing the Last Field Value in an EditableField

knicklabs:
When removing field values from an EditableField with multiple values, when the last field value is removed we need to reset it instead of deleting it because an editable field must always have at least one field value. 

There was an issue in the way were were resetting it. We were splatting the item and then overriding the `value` and `validated` properties to set them to `''` and `false`, respectively. The problem with this, is that items may have additional properties used in application logic, and by splatting the item on reset, those additional properties are retained.

An example of how this problem may manifest is if we had an `_id` property used to track the database id of a record and on commit, we used the presence of that property to determine if we would make a POST request to create a new record or a PUT request to update an existing record. If we delete the last item and then create a new one, the new item will have an `_id` property and thus the application will make a PUT request which will fail as that record no longer exists.

When we "remove" the last item, the `id` property is the only property we need to retain from the item being "removed". This is a unique key used for the HTML input id attribute and we want to recycle this. In total, the item we are left with should have three properties: `_id`, `value`, and `validated`.

Resolving this issue will resolve https://helpscout.atlassian.net/browse/HSAPP-1358.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.86.4-0",
+  "version": "2.86.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.86.3",
+  "version": "2.86.4-0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


knicklabs:
@tinkertrain Tagging you on this for review as the original author. I don't believe this change has any adverse affects. Will get QA on it nonetheless, but would appreciate any insights / concerns you have with this change.

Vesnaworx:
Tested and Passed in CI - @knicklabs I do not have access in order to remove the qa-review tag. Could you please remove it? Thank you!

knicklabs:
Soon as this is green I will merge and release.

PR #790, opened 2020-02-14
Modal V2

cocoabythefire:
This PR introduces versioning to the Modal component, and a new version 2 styling which can be seen in Figma [here](https://www.figma.com/file/qjxHKGok3AFDWJgcNFD3fv/Modals?node-id=200%3A704)

**Summary of changes in Modal V2** (these changes will only apply when passed `version = 2`):

- Introduces new `MODAL_KIND`. These are optionally passed into the Modal component via a `kind` prop. Styles are seen in the Figma: `Default`, `Branded`, `Alert`, and `Sequence`(shows a dot stepper). We also will want to support a `Custom` style where the header can be custom rendered but that is not included in this first set of changes.

- Modal optionally accepts a `state` prop. Currently, the only specially styled state is `danger` which turns the primary button and icon (if provided) red.

- A new `Modal.HeaderV2` component which is built-into the v2 of the component. Modal supports a few new optional props that are passed into the header: `kind`, ` title`, `description `, `icon`, `illo`, `illoSize`

- A new `Modal.ActionFooter` component which is optionally used to render an action toolbar along the bottom of v2 modals. It supports a default Cancel button (which may be hidden), Primary button, and optional Secondary button. The action footer also expects a modal style `kind` and renders accordingly.

- Slight tweaks to `Modal.Body` when inside a version 2 modal, most notably changes to padding and where the scrollbar appears if the content overflows.

-  A new open/close animation and darker overlay for the background behind the modal. Based off [this codepen](https://codepen.io/team/helpscout/pen/627b16d8a5d55747bae958c3dac0ad0e)

**Just a heads up re: fonts... The Modal `withAktiv` decorator isn't working (it is only applying to the screen behind the modal. So for me, everything is showing up in Storybook in Helvetica, although in reality these modals are all going to be showing in Aktiv Grotesk font. You can hack it in the inspector to see it with the correct font by adding the class `with-aktiv` in theory, although this does not work for me locally either.**

**Default**
Screen Shot 2020-02-14 at 9 34 06 AM

**Default in Danger state w/ Alert Icon**
Screen Shot 2020-02-14 at 9 34 14 AM

**Default, overflowing Body content**
Screen Shot 2020-02-14 at 9 37 02 AM

**Branded w/ example illo**
Screen Shot 2020-02-14 at 9 37 47 AM

**Alert**
Screen Shot 2020-02-14 at 9 38 41 AM

**Alert w/ danger state**
Screen Shot 2020-02-14 at 9 38 47 AM

**Sequence (step 2 of 5 example**
Screen Shot 2020-02-14 at 9 38 53 AM

**Storybook**
- View the new `Modal/V2` section for examples of usage


cocoabythefire:
@tinkertrain  can you give me a quick first pass assessment of code design/organization as this is my first real changes in HSDS, let me know what you think!

cocoabythefire:
TODO
- [ ] Add tests, make sure everything is passing

cocoabythefire:
fyi @cen10 ... getting a first pass of feedback right now, i haven't added tests yet but feel free to play around with this and let me know if you have any feedback!

cocoabythefire:
@jaredmcdaniel another overall pass on how all the stories look would be great. i also have a few specific questions for you:
- is the open/close animation looking as you expected? any tweaks still needed there?
- we talked about sometimes having a shadow on the footer toolbar but i forget under which conditions we wanted that and my notes aren't clear.
- i didn't get your hopes fully realized for the scrolling, the branded version does not currently scroll the header/description but given this hopefully won't happen often in real use, do you think it's sufficient as is for now when the content overflows?

jaredmcdaniel:
Nice @cocoabythefire!
- Super close on the close animation! Loading the modal looks great ... but Im wondering if when you close it, the modal can just fade out ... vs scale back down too?

- Let's remove the `default no cancel button` story. Id like to lean into keeping Cancel for most, if not all cases.

- for `default w/ very long content` ... can we customize the `c-Scrollable` gradient on top and bottom for this component? Instead of a white fader, I'd like to go with `background: linear-gradient( to bottom, rgba(0,0,0,.06), rgba(0,0,0,0));` or `to top` respectively. Would also give each a height of `4px`. Result:

image

- `branded w/ really long content` ... we can keep the top white fader, but incorporate the darker fader (mentioned above) for the bottom? Result:

image

plbabin:
Quick question to @jaredmcdaniel & @cocoabythefire: any reason we decided to go with having two version living side-by-side? It brings the total of different UI to three (no hs-app, hs-app and v2). Maybe it's easier for transitioning everything long term to this new design, but having gone through removing v2 for Button, Avatar & Dropdown I am kinda of afraid of introducing another versioned component. 

For the record,  I do think it would be nice to stick to only one version of Modal at least in HSDS 3.0.

@cocoabythefire I think that any new file could be written in plain JS since we have an open PR that removed TypeScript in HSDS 3.0

jaredmcdaniel:
Quick catch up with @plbabin and I believe we're on the same page. The goal is to ultimately have a single modal version (what is now v2), but regretfully the transition isn't going to happen overnight. That said, I'll leave it to you and @cocoabythefire as to when and how we get rid of the older versions and am happy to help in that process however I can. 👍

plbabin:
Synced up with Jared about Modal V2 and how we could integrate that with the upcoming `HSDS 3.0.0` release.

* The way it is setup right now is fine. I just wanted to point out that having 2 versions like that long term is hard to maintain. Quicker we can convert all modal to v2 the better it is.
* I will add (with your help @cocoabythefire ) a migration point in our Paper document so that when a team refactor a SPA within hs-app, Modal will be converted to V2.
* As soon as we are confortable with the release progress of `HSDS 3.0.0`, we'll convert the component to have V2 as the default one.

Does that seem to be a good plan?

BTW nice work on the component @cocoabythefire , I really like the direction here!

cocoabythefire:
> Synced up with Jared about Modal V2 and how we could integrate that with the upcoming `HSDS 3.0.0` release.
> 
> * The way it is setup right now is fine. I just wanted to point out that having 2 versions like that long term is hard to maintain. Quicker we can convert all modal to v2 the better it is.
> * I will add (with your help @cocoabythefire ) a migration point in our Paper document so that when a team refactor a SPA within hs-app, Modal will be converted to V2.
> * As soon as we are confortable with the release progress of `HSDS 3.0.0`, we'll convert the component to have V2 as the default one.
> 
> Does that seem to be a good plan?
> 
> BTW nice work on the component @cocoabythefire , I really like the direction here!

thanks PL! Yes that seems like a good plan. Before I started on the work I connected with JP and we given how complex the modal is and the time sensitivity of getting this changes in for onboarding, we didn't want to spiral the scope of this by having to upgrade and retest all uses of modal in HSApp if that makes sense. We need this to ship really soon for onboarding as its a prereq.

As for the future plans... yes I agree let's just make this the new default in HSDS 3.0 since we are already going to be QA'ing that... but because the usage is slightly different, we will have to update code in HSApp everywhere that we use modals to use it in the new style with the new prop s, etc.

I'm happy to help with updating docs wherever.

I figure this way we can roll out the v2 modal anytime and then keep iterating on it up until onboarding is live... we have opportunities to make it even better for HSDS v3 and beyond for sure. Thanks for all the feedback!

src/components/Modal/Modal.tsx
```diff
@@ -271,14 +351,39 @@ class Modal extends React.PureComponent<ModalProps> {
...
       className
     )
 
+    const innerWrapperClassName = classNames(
+      'c-Modal__innerWrapper',
+      v2 && 'v2',
+      kind === MODAL_STYLES.DEFAULT && 'is-default',
+      kind === MODAL_STYLES.ALERT && 'is-alert',
+      kind === MODAL_STYLES.BRANDED && 'is-branded',
+      kind === MODAL_STYLES.SEQUENCE && 'is-sequence'
```
plbabin: this pattern is repeated quite often. What do you think if we move that to a modal utility:

```javascript
export const getModalKindClassName = kind => {
  switch(kind){
    case MODAL_STYLES.ALERT: 
      return 'is-alert',
    case MODAL_STYLES.BRANDED: 
      return 'is-branded',
    case MODAL_STYLES.SEQUENCE: 
      return 'is-sequence',
    default:
      return 'is-default',
  }
}
```

src/components/Modal/Modal.utils.ts
```diff
@@ -1,8 +1,17 @@
...
   Header: 'ModalHeader',
+  HeaderV2: 'ModalHeaderV2',
   Body: 'ModalBody',
   Content: 'ModalContent',
+  ActionFooter: 'ModalActionFooter',
   Footer: 'ModalFooter',
   Overlay: 'ModalOverlay',
 }
+
+export const MODAL_STYLES = {
```
plbabin: since that represent a modal kind (base on the prop name) I would rename that constant to MODAL_KIND

src/components/Modal/Modal.tsx
cocoabythefire: done and done. great feedback.

cocoabythefire:
thanks @plbabin @tinkertrain @cen10  for looking this over. i've made all the code style changes recommended and updated the tests to be 100% coverage. i ran the beta release locally in my HSApp use case and it's working well so i'll probably merge this tomorrow.

Just have one CANI here I did not address for design changes @jaredmcdaniel the scrollable fader shadow adjustment is a bit messy to override, and will be better to solve this by enhancing our Scrollable component style options to allow for that shadow instead of the white one. I'm punting that and would like to address it separately as a Scrollable enhancement CANI.

Issue #775, opened 2020-01-28
Component: ButtonV2 can't be found using "getByText" in react-testing-library


cen10:
ButtonV2 nests the button text with a span:
```
<Button>
  <span>{text}</span> // simplified example
<Button>
```

When you use the react-testing-library `getByText('Save')` to find the Button component in a test, it ends up returning the span and not the button. This means that you can't click the button to submit a form and check that your spies were called.
```
const button = getByText('Save') // "button" is actually a <span>
button.click() 
expect(mySpy).toHaveBeenCalled() 
// ^ returns false because spans don't operate like buttons when clicked
```

Workaround:
```
const button = queryAllByRole('button')[2] 
// if there are multiple buttons on the page, eg Save, Cancel, Delete, you need 
// to get the array of buttons and pull out the one you need.
```

This is brittle as moving the buttons or adding a new one may end up resulting in the desired button having a different index in the returned array, but it will do the trick.

PR #793, opened 2020-02-20
Remove wrapping span around the button childrens only when it's not loading

plbabin:
[Jira](https://helpscout.atlassian.net/browse/JS-88)
Screen Recording 2020-02-20 at 10 00 AM

**What's the issue**
react-testing-library is not able to retrieve the Button label because our component wrap the label with a `span`.

**What's the new behavior?**
When the button is not loading, we don't wrap the button children with a span. We still need to wrap the children when it's loading to keep the button size the same for when the `isLoading` flag will be turned off.

### DIFF
--- a/src/components/Button/Button.css.js
+++ b/src/components/Button/Button.css.js
@@ -24,6 +24,14 @@ export const ButtonUI = styled.button`
   text-align: center;
   text-decoration: none;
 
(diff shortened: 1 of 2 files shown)


plbabin:
fixed #775

PR #795, opened 2020-02-24
HSDS 3.0.0 migration codemods

plbabin:
First let's just say that I am very excited by this PR. After creating the migration plan for HSDS 3.0.0, I started to think on how we could make it easier for dev to execute that migration.

I am very please to introduce HSDS 3.0.0 codemods.

## What's the new behavior?
This tooling will help any developper kickstart the process of migrating an SPA to HSDS 3.0. It will refactor some code for you or give you hints of place that would need to be refactor.

## How?
When releasing HSDS, we will add a `bin` that we could run on any app that installed HSDS:

```
npx hsds-codemod [transform] [path]
```

## List of all Codemods
Image 2020-02-24 at 5 01 10 PM

**All:** Run all codemod
**AutoDropdown:** Rename AutoDropdown to SearchableDropdown and add the autoInput prop
**AvatarGrid:** Rename AvatarGrid to AvatarList, and add props center & grid
**Button:** Remove the version={2} prop to Button. Rename some kind value
**ComboBox:** Rename to SearchableDropdown
**CssVariables:** Rename --BlueXXX to --HSDSXXX
**DeprecatedComponents:** Add a TODO comment before each deprecated component (before both the import and the component usage)
**DropdownV2:** Rename the import path to Dropdown, rename classNames when needed
**PropProvider:** Add a TODO comment before PropProvider import & component usage
**ReplaceImports:** Replace all hsds-react import with the next release (hsds-react-next)

## What's missing
There were some stuff that was too complicated to write in a codemod. For example `PropProvider` needs to be replace a on per component basis hence why we just add a TODO comment.

The setup of HSDS (adding global css + HSDS.Provider) needs to be place at a strategic place in our SPA. Each one is different so the developer should make the decision on where it should happen

## Warnings
Even though the tool will replace some code for you, it doesn't replace the fact that each migration would need to be QA. Same goes with validating and peer reviewing the code that was updated by this script.

## Up Next
With the v3 beta release, we'll be able to validate that the `bin` creating is working fine, otherwise we'll do some adjustment on the fly. All scripts have been tested with HsAPP code so far.

### DIFF
--- a/scripts/codemod/cli.js
+++ b/scripts/codemod/cli.js
@@ -1,0 +1,203 @@
+/**
+ * inspire by react-code-mod CLI: https://github.com/reactjs/react-codemod/blob/master/bin/cli.js
+ */
+
+// hsds-codemod optional-name-of-transform path/to/src [...options]
+
+const globby = require('globby')
+const inquirer = require('inquirer')
+const meow = require('meow')
+const path = require('path')
+const execa = require('execa')
+const chalk = require('chalk')
+const isGitClean = require('is-git-clean')
+
+const transformerDirectory = path.join(__dirname, 'transforms')
+console.log(__dirname)
+
+const jscodeshiftExecutable = require.resolve(
+  '../../node_modules/jscodeshift/bin/jscodeshift.sh'
+)
+
+function checkGitStatus(force) {
+  let clean = false
+  let errorMessage = 'Unable to determine if git directory is clean'
+  try {
+    clean = isGitClean.sync(process.cwd())
+    errorMessage = 'Git directory is not clean'
+  } catch (err) {
+    if (err && err.stderr && err.stderr.indexOf('Not a git repository') >= 0) {
+      clean = true
+    }
+  }
+
+  if (!clean) {
+    if (force) {
+      console.log(`WARNING: ${errorMessage}. Forcibly continuing.`)
+    } else {
+      console.log('Thank you for using hsds-react-codemods!')
+      console.log(
+        chalk.yellow(
+          '\nBut before we continue, please stash or commit your git changes.'
+        )
+      )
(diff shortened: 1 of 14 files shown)


cocoabythefire:
This was a great idea. Thanks for all this work!

scripts/codemod/cli.js
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+}
+
+function runTransform({ files, flags, parser, transformer }) {
+  const transformerPath = path.join(transformerDirectory, `${transformer}.js`)
+
+  let args = []
+
+  const { dry, print } = flags
```
cocoabythefire: +1 love doing a dry run first!

plbabin: verbose would need to be set to `2` to make it valuable. I guess I could rewrote the script to force the verbose value when a dry run is active

plbabin:
> The only thought I have... since codemods can be slightly dangerous, would it be worthwhile to write tests? So for instance check that a Button in a test file looks correct after running the transform against it? I'm not sure the lift would be worth it for some of these simple transforms but just a question i thought worth throwing out there. Nice work!

I am not oppose to write test on those codemods, I am just afraid that time available is rare these days :).

I didn't write any tests at first because we are a small team, and each update would need to be validate through commits and code reviews. I kinda trust everyone on this team to make the right call when updating. Goal #1 was to kickstart the refactor process, not necessarily have 
 code ready to go to production like the codemods from the react team. One option to reduce the risk is to run codemod one at the time, and validate that all changes are good to go. Another thing we can do is add me or Juan Pablo as reviewer for each SPA migration.

That being said, if anyone wants to add tests to our codemods I would be more than happy to add them to the PR 😄

PR #796, opened 2020-02-25
HSAPP-1451 Can Now Change Props on Children of EditableField.Composite

knicklabs:
It was not possible to pass new props to the EditableField components that are children of an EditableField.Composite component.

This was tricky to debug (thanks for looking at this with me @cen10), but we tracked the problem down. We were performing a deep equality comparison directly on the next and previous children, which included internal React properties that had circular dependencies. This resulted in infinite recursion with the equal method.

To resolve the issue, we compare only the props passed to the next and previous children to see if the props changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "2.87.1-1",
+  "version": "2.89.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


Vesnaworx:
Tested and Passed in storybook:

- tested clicking change to change the properties
- tested deleting field data and clicking change
- tested adding to field data and then clicking change

knicklabs:
I'm returning to this to fix up a couple more issues we've ran into. I'll bug you all for re-review when I'm done. Thanks!

knicklabs:
@cen10 @tinkertrain This is ready for re-review. I made a change to prevent the setTimeout callback from running in the case that the component has become unmounted. Reason being that it performances dom and state operations that will throw an error if executed after the component has been unmounted.

PR #799, opened 2020-03-02
Fix emoji-only UI in Message.Bubble component

TerrenceLJones:
In release https://github.com/helpscout/hsds-react/releases/tag/v2.88.0 the Chat UI was updated for when the message body included only Emoji characters. However this work did not include the scenario when message data was passed in the `body` prop. This PR addresses that missing support. 

### Before
Screen Shot 2020-03-02 at 16 07 55

### After
Screen Shot 2020-03-02 at 16 04 20

### DIFF
--- a/src/components/Message/Message.Bubble.tsx
+++ b/src/components/Message/Message.Bubble.tsx
@@ -131,7 +131,10 @@ export const Bubble = (props: Props, context: Context) => {
 
     if (showEmojiOnlyStyles) {
       return (
-        <MessageBubbleBody className="c-MessageBubble__body">
+        <MessageBubbleBody
+          className="c-MessageBubble__body"
+          showEmojiOnlyStyles={showEmojiOnlyStyles}
+        >
           <Text wordWrap lineHeightInherit size={48}>
             {body}
           </Text>
--- a/src/components/Message/Message.Bubble.tsx
+++ b/src/components/Message/Message.Bubble.tsx
@@ -70,27 +70,6 @@ export const Bubble = (props: Props, context: Context) => {
   )
 
   let showEmojiOnlyStyles = false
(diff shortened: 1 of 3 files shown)


src/components/Message/Message.Bubble.tsx
```diff
@@ -121,24 +100,61 @@ export const Bubble = (props: Props, context: Context) => {
...
-      className="c-MessageBubble__body"
-      isEmbed={isThemeEmbed}
-      dangerouslySetInnerHTML={{
-        __html: enhanceBody(body),
-      }}
-    />
-  ) : (
-    childrenMarkup
-  )
+  const hasOnlyOneChild = React.Children.count(children) === 1
```
TerrenceLJones: Nothing changed here with the block, expect that I moved it closer to where it is called.

src/components/Message/Message.Bubble.tsx
```diff
@@ -121,24 +100,61 @@ export const Bubble = (props: Props, context: Context) => {
...
+
+  const childrenMarkup = React.Children.map(children, child => {
```
TerrenceLJones: Nothing changed in the `childrenMarkup`, expect that I moved it closer to where it is called.

src/components/Message/Message.Bubble.tsx
```diff
@@ -121,24 +100,61 @@ export const Bubble = (props: Props, context: Context) => {
...
+        <Text wordWrap lineHeightInherit size={fontSize}>
+          {child}
+        </Text>
+      </MessageBubbleBody>
+    ) : (
+      child
+    )
+  })
+
+  const renderBody = () => {
```
TerrenceLJones: Changes begin here.

PR #800, opened 2020-03-03
Fix chat message body with emoji spacing issue

TerrenceLJones:
This address a spacing issues when Emoji only messages are sandwiched between other messages. 

## Before
Screen Shot 2020-03-03 at 12 54 29
Screen Shot 2020-03-03 at 15 42 21

## After
Screen Shot 2020-03-03 at 12 54 16

Screen Shot 2020-03-03 at 15 41 23

### DIFF
--- a/src/components/Message/Message.Bubble.tsx
+++ b/src/components/Message/Message.Bubble.tsx
@@ -55,19 +55,6 @@ export const Bubble = (props: Props, context: Context) => {
   const isThemeNotifications = theme === 'notifications'
   const isThemeEmbed = theme === 'embed'
   const fromName = from && typeof from === 'string' ? from : null
-  const componentClassName = classNames(
-    'c-MessageBubble',
-    from && 'is-from',
-    icon && 'withIcon',
-    isNote && 'is-note',
-    size && `is-${size}`,
-    ltr && !rtl && 'is-ltr',
-    !ltr && rtl && 'is-rtl',
-    theme && `is-theme-${theme}`,
-    to && 'is-to',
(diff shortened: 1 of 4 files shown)


src/components/Message/Message.Bubble.tsx
```diff
@@ -167,6 +154,21 @@ export const Bubble = (props: Props, context: Context) => {
     </div>
   )
 
+  const componentClassName = classNames(
```
TerrenceLJones: Moved this as I needed access to the latest version of `showEmojiOnlyStyles`

src/components/Message/styles/Message.css.js
```diff
@@ -60,6 +60,11 @@ const css = `
     margin-bottom: 0;
     min-width: initial;
   }
+
```
TerrenceLJones: This is an override for when a message only includes an emoji and no other related messages. 

Screen Shot 2020-03-03 at 12 58 58

src/components/Message/styles/Message.css.js
```diff
@@ -60,6 +60,15 @@ const css = `
...
+  .c-MessageChatBlock:only-child .c-MessageBubble.emoji-only {
+    padding-top: 0 !important;
+    padding-bottom: 0 !important;
+  }
+
+  .c-MessageChatBlock:last-child .c-MessageBubble.emoji-only {
```
TerrenceLJones: This is an override for when a message block ends with a message that includes an emoji.
Screen Shot 2020-03-03 at 15 41 23

PR #806, opened 2020-04-06
HSDS v3.0.0

plbabin:
This update bring HSDS to a new area! This is a sum of 6 months of work that improves numerous facet of HSDS that will make our life better.

You can always read the whole PR history to get a sense of that did change and understand the process of how we get to what we have in this PR. This is a summary of what did change:

https://github.com/helpscout/hsds-react/pulls?q=is%3Apr+label%3Av3+sort%3Acreated-asc

## Fancy -> styled-components
We rewrote all components to use styled-components as the main rendering engine. HSDS is also SCSS free, everything is render by styled-component.

## TypeScript is long gone
TS wasn't adding anything good to our daily coding session. We decided as a group to get rid of it.

##  Global Stylesheet
We introduce the concept of global stylesheet so we can set some base ground for any HSDS components. It will force the font-size & font-family while staying customizable by each application. 

We also create an `HSDS.Provider` that will scope all selectors + initialize the global stylesheet.

## Storybook as the source of truth
We completely refactored storybook to be the source of truth for our UI. We carefully went through each components to adjust the UI to make it pixel perfect

We are now leveraging the new storybook documentation API with MDX to create each components docs page.

## Migration Codemod
To update our applications to the new HSDS version, there is some code update that will need to happen. We create a codemod to kickstart the process and do all boring tasks for us (renaming import path, renaming variable, renaming component, etc).

## New Tooltip
The tooltip and popover component were refactor to use react-tipsy behind the scene instead of our own Pop/Popper implementation.


knicklabs:
@plbabin what are the release plans for this?

plbabin:
> @plbabin what are the release plans for this?

in the HSDS repo
* move v3 as the master branch
* move master as a v2 branch for hot fixes

in HsApp
* Create an alias for master (v3) which will be `@helpscout/hsds-react-next`
* Start refactoring per SPA (codemods will help kickstart that)

When is another question I don't have an answer for. It's up to each team. I do want to refactor one SPA as an example (and also see how long it takes to have a rough estimate for the whole migration).

cocoabythefire:
This looks great PL!! I reviewed the Modal v2 stuff specifically in addition to just playing around with this branch. The Storybook updates are verrrry nice.

I think a per-app rollout for HSApp makes sense... the QA effort to ensure nothing has broken will be quite a hefty undertaking i would think.

plbabin:
> This looks great PL!! I reviewed the Modal v2 stuff specifically in addition to just playing around with this branch. The Storybook updates are verrrry nice.
> 
> I think a per-app rollout for HSApp makes sense... the QA effort to ensure nothing has broken will be quite a hefty undertaking i would think.

yes QA will be very important but after 2-3 SPA migrated I think we'll be very confident in the new rendering engine and it will reduce the time it will take to QA for each new SPA

Charca:
This is really exciting! Do you think we can cut a Release Candidate from the `v3` branch before merging into master so we can start testing it? I want to upgrade HSDS in Beacon to get a sense of how much work it'll be to move everything to the new version.

I've also noticed that the Netlify deploy preview is not working on this PR. Is that something we'll fix after we merge this into master?

plbabin:
> This is really exciting! Do you think we can cut a Release Candidate from the `v3` branch before merging into master so we can start testing it? I want to upgrade HSDS in Beacon to get a sense of how much work it'll be to move everything to the new version.
> 
> I've also noticed that the Netlify deploy preview is not working on this PR. Is that something we'll fix after we merge this into master?

hum deploy preview seems to be working: https://deploy-preview-806--hsds-react.netlify.com/

We can definitely start with a release candidate. I like this idea

First thing for Beacon will be to be written in React 16.8+. Then the major thing that we need to verify is the Frame styling and the namespacing (to make sure we don't interfere with other styled-components website).

PR #807, opened 2020-04-15
Focus improvements

plbabin:
This update adds a couple of improvements to VerificationCode, Modal & CopyInput

### VerificationCode
We did add support to auto focus the input. It will be based on the actual code value. For example, if the code length is 3, the 3rd input will be focused.

Screen Recording 2020-04-15 at 04 39 PM

We also add the prop `autoSubmit`. When the input value reach the number of character, the `onEnter` prop will be executed. It will also work when pasting the code.

Screen Recording 2020-04-15 at 04 41 PM

### CopyInput
We add the ability to copy the input value when pressing <kbd>enter</kbd> while focusing the field

### Modal
We fix a bug where the modal title was always displayed as a tooltip when hovering a modal

### DIFF
--- a/src/components/CopyInput/CopyInput.jsx
+++ b/src/components/CopyInput/CopyInput.jsx
@@ -44,7 +44,6 @@ class CopyInput extends React.PureComponent {
   handleCopyClick = () => {
     this.selectText()
     this.copyToClipboard()
-    console.log('handleCopyClick', this.getInputValue())
     this.props.onCopy(this.getInputValue())
   }
 
--- a/src/components/CopyInput/CopyInput.jsx
+++ b/src/components/CopyInput/CopyInput.jsx
@@ -44,6 +44,7 @@ class CopyInput extends React.PureComponent {
   handleCopyClick = () => {
     this.selectText()
     this.copyToClipboard()
+    console.log('handleCopyClick', this.getInputValue())
     this.props.onCopy(this.getInputValue())
   }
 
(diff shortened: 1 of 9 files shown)


jaredmcdaniel:
@plbabin looks great! Couple questions on the VerificationCode component ... 

1. Select all seems to work fine in storybook. 🎉 Should I assume it will work the same in-app? Where `cmd-a` selects only the numbers in the input (vs all elements on the page)?

2. Can we do anything about skipped inputs when typing ... "fast"?

fast

plbabin:
> @plbabin looks great! Couple questions on the VerificationCode component ...
> 
> 1. Select all seems to work fine in storybook. 🎉 Should I assume it will work the same in-app? Where `cmd-a` selects only the numbers in the input (vs all elements on the page)?
> 2. Can we do anything about skipped inputs when typing ... "fast"?
> 
> fast

1. <kbd>cmd+a</kbd> will stay inside the component. In this story (https://deploy-preview-807--hsds-react.netlify.app/?path=/story/components-forms-verificationcode--auto-focus-with-half-code), there is a paragraph next to the input, you will see that kbd>cmd+a</kbd> doesn't select the text

2. I'll see what I can do. That is an edge case in the sense that I don't think someone will type that fast to match a code read from an email

src/components/CopyInput/CopyInput.jsx
```diff
@@ -43,6 +44,7 @@ class CopyInput extends React.PureComponent {
   handleCopyClick = () => {
     this.selectText()
     this.copyToClipboard()
+    console.log('handleCopyClick', this.getInputValue())
```
cocoabythefire, line 47: debugging

cocoabythefire:
LGTM thanks @plbabin for these awesome improvements!

PR #811, opened 2020-04-20
Tab focus cycling in Modal

plbabin:
This updates move the focus to the first or last in the list based on if the user reach an edge node inside a Modal

Screen Recording 2020-04-20 at 05 02 PM

_**edit**_

### Modal icon size
We also update the `Modal` component to add a new prop  `iconSize` to modify the size of the header icon

### VerificationCode autoSubmit
We split the autoSubmit prop into two different prop (`autoSubmitPaste` `autoSubmitKeyUp`), to allow for more customization. We can now trigger the auto submit on paste or keyup only, or for both.

### DIFF
--- a/src/components/Modal/Modal.jsx
+++ b/src/components/Modal/Modal.jsx
@@ -186,7 +186,7 @@ class Modal extends React.PureComponent {
 
     if (focusedNodeIndex === focusableNodes.length - 1) {
       event.preventDefault()
-      focusableNodes[0].focus()
+      if (focusableNodes && focusableNodes[0]) focusableNodes[0].focus()
     }
   }
 
@@ -199,7 +199,8 @@ class Modal extends React.PureComponent {
 
     if (focusedNodeIndex === 0) {
       event.preventDefault()
-      focusableNodes[focusableNodes.length - 1].focus()
+      const i = focusableNodes.length - 1
+      if (focusableNodes && focusableNodes[i]) focusableNodes[i].focus()
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/components/Modal/Modal.jsx
```diff
@@ -199,7 +199,8 @@ class Modal extends React.PureComponent {
 
     if (focusedNodeIndex === 0) {
       event.preventDefault()
-      focusableNodes[focusableNodes.length - 1].focus()
+      const i = focusableNodes.length - 1
```
cocoabythefire, line 202: should we defend for the possibility that focusableNodes.length could be 0 which make this try to access index of -1?

src/components/Modal/Modal.jsx
```diff
@@ -64,6 +64,7 @@ class Modal extends React.PureComponent {
     isHsApp: PropTypes.bool,
     isOpen: PropTypes.bool,
     icon: PropTypes.string,
+    iconSize: PropTypes.string,
```
cocoabythefire, line 67: can we accept string or number and be able to handle either? just a nice to have

src/components/Modal/Modal.jsx
plbabin: yes good catch, I did fixed it in 2d6cabbf

src/components/Modal/Modal.jsx
plbabin: 👍 all done in 2d6cabbf

PR #816, opened 2020-04-28
Add spacing to Chat message groups

TerrenceLJones:
[Jira Ticket](https://helpscout.atlassian.net/browse/HSAPP-1956)

Add more spacing to each Chat message group. 

### Before

Screen Shot 2020-04-28 at 16 30 18

### After (Actually screenshot)
Screen Shot 2020-04-28 at 16 29 14

### DIFF
--- a/src/components/Message/styles/Action.css.js
+++ b/src/components/Message/styles/Action.css.js
@@ -9,6 +9,7 @@ export const ActionUI = styled('div')`
 
   line-height: 22px;
   text-align: center;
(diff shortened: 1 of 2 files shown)


TerrenceLJones:
> This change is **only** intended to affect the status row (not all chat groups). So the extra margin on the status rows is exactly as intended, but everywhere else it's not.

@digitaldesigner Ahh sorry for my confusion...I'll fix that up right now.

PR #818, opened 2020-04-29
Refactor tooltip & popover to use Tippy

plbabin:
This update refactored entirely the Tooltip and Popover components

**What was the old behavior?**
Before we were using a fork of react-popper and the old popper.js function. That creates a lot of confusion in the team when it was time to modify or improve some functionalities in Tooltip/Popover. A lot of code was forked with the intent to support React 16 features in a react 15 codebase.

**What's new?**
We remove entirely Pop/Popper and Popper.js 1.x.x and replace them with modern library [Tippy](https://github.com/atomiks/tippyjs-react) (that use [react-popper/Popper V2](https://github.com/popperjs/react-popper))

The gain here is to simplify each team member life when working with Tooltip, with a bulletproof solution.

Both component are retro compatible with HSDS v2 prop, to facilitate the migration. We kept around only the props that were actively used in HSApp and Beacon, to keep each component as simple as possible.

## Tooltip
Screen Recording 2020-04-29 at 09 39 AM
Image 2020-04-23 at 2 48 28 PM
Image 2020-04-29 at 10 14 09 AM

The tooltip component has some improvements:

* Position awareness. If the tooltip is close to the edge of the screen (or the parent anchor), it will reverse the placement
* Multi line support
* bottom start/end and top start/end support

We map our internal prop name to match Tippy so we could stay retro compatible with the old version. For example, we use `triggerOn="hover"` but we convert that to `trigger="mouseenter focusin"` internally.

## Popover
Screen Recording 2020-04-29 at 09 38 AM

This component continue to use Tooltip internally. We simplify the renderHeader/Content and we overwrite the render function so we could restyle entirely the tooltip, without change the core API.


src/components/FilteredList/FilteredList.jsx
```diff
@@ -64,6 +64,7 @@ export class FilteredList extends React.Component {
 
   renderBadgeContent = () => {
     const { limit, items } = this.props
+    console.log(items)
```
tinkertrain, line 67: Stray console.log 🙂

PR #820, opened 2020-05-01
COMMS-613 Dropdown Improvements

TerrenceLJones:
Included in this change is:
- adding spacing between dropdown menu and the trigger
- fix issue where triggers via renderTrigger were not being informed on hover events

Screen Recording 2020-04-30 at 19 13

Screen Shot 2020-04-30 at 19 21 04
Screen Shot 2020-04-30 at 19 20 49

Screen Recording 2020-04-30 at 19 22


TerrenceLJones:
> great addition! I wonder if we should add the same spacing to AutoDropdown and ComboBox (SearchableDropdown in v3) ?

@plbabin I just checked each and they both inherited the spacing change in the core Dropdown component. So we're good to go.

PR #823, opened 2020-05-11
PROD-726 Fix Margin Bottom on HSDS Form

knicklabs:
The margin-bottom on the HSDS Form, applied via the Form Actions, was 100px, which was much too large. It should have been 20px, which is consistent with other forms in Help Scout.

In the following screenshot, we see the Form Actions used inside an Accordion Section. In total, there is 120px of space at the bottom, where there should be 40px (20px for the Form Action margin bottom and 20px for the Accordion Section padding.

Problem

In the next screenshot we see how after applying this change, we have the correct amount of space.

Solution

### DIFF
--- a/src/components/Form/Form.Actions.css.js
+++ b/src/components/Form/Form.Actions.css.js
@@ -1,7 +1,7 @@
 import styled from 'styled-components'
 
 export const config = {
-  marginBottom: 100,
+  marginBottom: 20,
   marginTop: 30,
   spacing: 10,
 }


knicklabs:
@plbabin This is for V3 so I targeted Master since V3 was merged. Is this correct? I see we still have a V3 branch on here. Should that be deleted?

plbabin:
Yes targeting master is what we want. V3 branch was not deleted because it was referenced in another PR

We should be able to delete it now

plbabin:
@knicklabs I am fine with grouping multiple small patches in the same release. I have some stuff I want to get out in a patch this morning as well (for v3).

knicklabs:
@plbabin Sounds good to me!

PR #824, opened 2020-05-11
PROD-726 Fix Margin Bottom on HSDS Form (v2)

knicklabs:
The margin-bottom on the HSDS Form, applied via the Form Actions, was 100px, which was much too large. It should have been 20px, which is consistent with other forms in Help Scout.

In the following screenshot, we see the Form Actions used inside an Accordion Section. In total, there is 120px of space at the bottom, where there should be 40px (20px for the Form Action margin bottom and 20px for the Accordion Section padding.

Problem

In the next screenshot we see how after applying this change, we have the correct amount of space.

Solution


knicklabs:
@plbabin This targets the v2 branch. Did I do this correctly?

PR #833, opened 2020-06-01
HSDS-149: New variant: Multi-line (X-Large) option for Accordion Row Link

Charca:
**[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-149)**

This PR adds a new size (`xl`) for the Accordion component, as well as a couple of new sub-components (Heading and Subheading) for creating a multi-line row.

I've also updated the drag handler icon to the new version and added the option to add a badge and a "status" color to each row.

With this new variant, we should be able to get rid of the MessageList component completely.

## Screenshot

Screen Shot 2020-06-01 at 5 28 33 PM

@digitaldesigner the hover style (grey background) doesn't look very good on the red/blue rows, do you think we need separate styles for these rows? Storybook [here](https://deploy-preview-833--hsds-react.netlify.app/?path=/story/components-structural-accordion--multi-line)


Charca:
> Mate this looks perfect, only one tiny tweak from me — could you make sure that the red and blue rows don't have a background color change? Normal rows are fine, but in those special ones the background should remain red and blue :)

@digitaldesigner done! https://deploy-preview-833--hsds-react.netlify.app/?path=/story/components-structural-accordion--multi-line

src/components/Accordion/Accordion.Heading.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { HeadingUI } from './Accordion.css'
+
+export default ({ children }) => (
+  <HeadingUI truncate weight={500} size={14} className="Accordion__Heading">
```
plbabin, line 5: I am fine leaving it like that (in parts because I think we should target component as selector over classNames), but the convention I believe is to prefix className with `c-`

src/components/Accordion/Accordion.Subheading.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { SubheadingUI } from './Accordion.css'
+
+export default ({ children }) => (
+  <SubheadingUI
+    truncate
+    weight={400}
+    size={13}
+    className="Accordion__Subheading"
```
plbabin, line 9: same comment as earlier about classNames prefix

src/components/Accordion/Accordion.css.js
```diff
@@ -144,7 +155,25 @@ export const TitleUI = styled('div')`
     .${titleClassNames.iconCaretClassName} {
       color: ${getColor('text.slightlyMuted')};
     }
+
+    .Accordion__Heading {
+      color: #253642;
```
plbabin, line 160: I think the color could be switch to `charcoal.700`. cc @digitaldesigner

src/components/Accordion/Accordion.css.js
```diff
@@ -144,7 +155,25 @@ export const TitleUI = styled('div')`
...
+    }
+
+    .Accordion__Subheading {
+      color: #556575;
```
plbabin, line 164: colorway is telling me that `charcoal.400` would be the closest color for that one. Would that work @digitaldesigner ?

src/components/Accordion/Accordion.css.js
```diff
@@ -220,11 +253,11 @@ export const TitleUI = styled('div')`
       left: -15px;
     }
     &.is-page .drag-handle {
-      left: 15px;
+      left: 39px;
     }
 
     .drag-handle {
-      color: ${getColor('grey.800')};
+      color: #405261;
```
plbabin, line 260: closest color for that hex code would be `charcoal.500`. cc @digitaldesigner

Charca:
Thanks for the review, friends!

@plbabin I've updated the class names to follow our convention, and I'll defer to @digitaldesigner on the questions about colors and the dragging behavior.

Charca:
I'll merge + release this one. @digitaldesigner please ping me if you want me to make the color changes and I'll open a new PR. Thanks! :)

PR #835, opened 2020-06-04
Adds Babel plugin to make HSDS imports nicer to write while tree-shakeable

tinkertrain:
What do we _don't_ want?

```js
import Page from 'hsds-react-next/components/Page'
import Button from 'hsds-react-next/components/Button'
import Select from 'hsds-react-next/components/Select'
```

What do we want?

```js
import { Page, Button, Select } from 'hsds-react-next'
```

Currently applications need to import components using the first mode above to allow tree-shaking, otherwise we can use `import { Page, Button, Select } from 'hsds-react-next/components/'` (notice the "/components" requirement) but it won't be tree-shakeable.

This PR adds a babel plugin that takes care of it for you!

So feel free to write the nice and concise form of importing, and get all the benefits. Who said you can't have your cake and eat it too?

Have a look at the test cases, in a nutshell:
- Allows you to use destructuring
- Allows you to remove "components" from the import path
- Accepts one option `packages` to instruct which hsds package you want this to have an effect on, say both of "helspcout-hsds-react-next" and "@helpscout/hsds-react" or just one

Besides unit testing included here, I made some tests with a brand new app, check the results for yourselves:

Destructuring without the plugin:

Screen Shot 2020-06-04 at 2 40 15 PM

With plugin:

Screen Shot 2020-06-04 at 2 40 47 PM

To use, add it to your babel conf file:

```
{
  "plugins": [
    [
      "@helpscout/hsds-react/src/utilities/babel-plugin-hsds-imports/",
      { "packages": ["helpscout-hsds-react-next"] }
    ]
  ]
}

```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5253,6 +5253,16 @@
       "integrity": "sha512-/+CRPXpBDpo2RK9C68N3b2cOvO0Cf5B9aPijHsoDQTHivnGSObdOF2BRQOYjojWTDy6nQvMjmqRXIxH55VjxxA==",
       "dev": true
     },
+    "@types/babel-plugin-tester": {
+      "version": "9.0.0",
+      "resolved": "https://registry.npmjs.org/@types/babel-plugin-tester/-/babel-plugin-tester-9.0.0.tgz",
+      "integrity": "sha512-+YTsSQgllJTOrvWT+DR34GQbLs/oKMODZV2WbypdoZTrZlr/+jkv4HK+E69ljoThECXErzQeItCNE/tZDOoZRA==",
+      "dev": true,
+      "requires": {
+        "@types/babel__core": "*",
+        "@types/prettier": "*"
+      }
+    },
     "@types/babel-types": {
       "version": "7.0.7",
       "resolved": "https://registry.npmjs.org/@types/babel-types/-/babel-types-7.0.7.tgz",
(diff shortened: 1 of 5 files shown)


Charca:
Great job @tinkertrain! Just a quick question, have you looked into the `babel-plugin-transform-imports` [plugin](https://www.npmjs.com/package/babel-plugin-transform-imports)? It's what we use in Beacon to allow this type of import syntax for hsds. Do you think we could use it here or extend it somehow?

plbabin:
we might be able to achieve the same thing with `babel-plugin-transform-imports`. I really like the current implementation in this PR though, but I don't know which direction we should take. `babel-plugin-transform-imports` was not updated in over a year, but maintaining or own babel plugin might not be ideal long term too. I don't know

I would love to have the input of everyone on the team on that situation!

tinkertrain:
@Charca @plbabin @knicklabs Thanks for looking at this!

Some reasons I went with in-house dev for this as opposed to 3rd party is that I wanted a super fast plugin that adds a negligible processing time, that does exactly the thing we need and nothing more and if we need to modify it down the line is so simple we don't have to wade through a more complex source code.

This is a _very_ simple plugin (and I stress the simpleness of it the plugin meat is <60 lines), definitely nowhere close to the maintenance burden in the style of Proxypack or Fancy or anything like that.

But as I stated in slack, I present this is as a proposal for _the Guild_ approval 👍

PR #837, opened 2020-06-09
Decreases babel plugin tester version for node 8

tinkertrain:
Pushed the last release using node 12, travis ci is failing tests, had to decrease the version of babel plugin tester as it doesn't support < 10


tinkertrain:
Not needed

PR #844, opened 2020-06-16
Change CopyInput to use copy-small icon

plbabin:
This update add an icon prop to the `CopyButton` component. We also add a label prop that can be turn off. The end result is a icon button that still has the confirm animation (triggered on click)

Screen Recording 2020-06-16 at 10 15 AM

This update also change the CopyInput component to leverage this new button style

Screen Recording 2020-06-16 at 10 16 AM

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -53,7 +53,8 @@ class Button extends React.PureComponent {
 
     return React.Children.map(children, (child, index) => {
       if (child && !child.hasOwnProperty('type')) return child
-      if (child.type !== Icon) return child
+      if (!child || (child && child.type !== Icon)) return child
+
       const len = React.Children.count(children)
       const isFirst = index === 0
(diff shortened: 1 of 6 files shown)


src/components/Button/Button.jsx
```diff
@@ -53,7 +53,8 @@ class Button extends React.PureComponent {
 
     return React.Children.map(children, (child, index) => {
       if (child && !child.hasOwnProperty('type')) return child
-      if (child.type !== Icon) return child
+      if (!child || (child && child.type !== Icon)) return child
```
tinkertrain, line 56: Not a big deal, I find these a bit hard to read, perhaps...

```
if (!child) return
if (!child.hasOwnProperty('type')) return child
if (child.type !== Icon) return child
```

Maybe is just me though 😅

plbabin: that's a great suggestion. 🙇

PR #845, opened 2020-06-16
Fix: Tooltips on Input Errors Do Not Display Correctly

knicklabs:
## Problem

The styles of the trigger are interfering with the styles of the tooltip, constraining it to the width of the trigger. This causes the text to wrap and due to the tall height of the tooltip, its placement will be forced to the bottom more easily than it should. The image below illustrates the problem.

Screenshot illustrating the Problem

## Solution

As [suggested in the documentation](https://atomiks.github.io/tippyjs/v5/faq/#my-tooltip-appears-cut-off-or-is-not-showing-at-all), the tooltip is being appended to the body to deal with the issue of the conflict between its display and the bounds of the trigger.

Screenshot illustrating the Solution

### DIFF
--- a/src/components/Input/Input.jsx
+++ b/src/components/Input/Input.jsx
@@ -433,6 +433,7 @@ export class Input extends React.PureComponent {
         <Tooltip
           animationDelay={0}
           animationDuration={0}
+          appendTo={document.body}
           display="block"
           placement="top-end"
           title={errorMessage}


knicklabs:
I've applied this fix to V3 only since it is not an issue in V2.

plbabin:
I am starting to think that the interactive option is causing more harm than good. Appending on the parent is to have the tooltip accessible, but with all the way we are using it's starting to be a problem. 

I am leaning toward removing the interactive option by default for tooltip (which will move back the append to be on the body by default). We could keep the popover to be interactive and change the appendTo when needed)

knicklabs:
@plbabin Have you run into other problems with the tooltip? If not, I'd lean towards overriding as needed (like was done here). So far, I've only encountered this problem on the Input error suffix.

plbabin:
> @plbabin Have you run into other problems with the tooltip? If not, I'd lean towards overriding as needed (like was done here). So far, I've only encountered this problem on the Input error suffix.

on EditableField and maybe somewhere directly in hs-app as well. I am fine keeping it like that for now, and if it's a trend we can revisit this issue

PR #847, opened 2020-06-16
Update modal v2 alert styles to match design specs

plbabin:
We update the modal alert component to match the current specs

* sit vertically centered on the page be 440px wide
* have Horizontal padding: 50px
* have Top padding: 50px
* have Bottom padding: 60px
* have min-height 180px (vs 179)
* Button group sits 24px below the copy

Image 2020-06-16 at 11 22 20 AM
Image 2020-06-16 at 11 23 05 AM
Image 2020-06-16 at 11 24 59 AM

Those changes will only affected `v3`

### DIFF
--- a/src/components/Modal/Modal.Body.jsx
+++ b/src/components/Modal/Modal.Body.jsx
@@ -57,7 +57,7 @@ class ModalBody extends React.PureComponent {
 
     const componentClassName = classNames(
       'c-ModalBody',
-      v2 && 'is-v2',
+      v2 && 'v2',
       seamless && 'is-seamless',
       scrollable ? 'is-scrollable' : 'is-not-scrollable',
       className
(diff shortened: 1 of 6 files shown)


src/components/Modal/Modal.Body.jsx
```diff
@@ -57,7 +57,7 @@ class ModalBody extends React.PureComponent {
 
     const componentClassName = classNames(
       'c-ModalBody',
-      v2 && 'is-v2',
```
cocoabythefire, line 60: this is fine since you are saying it will only apply to hsds v3... i just want to note that there are two instances of style overrides in hsapp targeting the '.is-v2' class, so if you end up merging this into v2 it will cause issues.

plbabin: I think alert is not even used anywhere right now (maybe in setup-guide, but it's in the migration process), hence why I was thinking of keeping it to v3 only

PR #848, opened 2020-06-16
Fix accordion useEffect to recalculate openSectionIds

plbabin:
This update fixed an issue with the way the openSectionIds were recalculate with useEffect

Now we make sure that if the array is empty we set the active section as empty. We also make sure to only set one sections as active if the `allowMultiple` prop is set to false


plbabin:
@knicklabs happy to deploy a preview so we can verify that everything is working fine in hs-app

plbabin:
All changes from this PR has been migrate to #840. Closing this PR so we can work on a single branch on this component

PR #851, opened 2020-06-24
Storybook figma fix, dependencies audit and ThemeProvider issue

tinkertrain:
## Problem (s)

😵 @digitaldesigner mentioned that the figma preview in stories was no longer working... and while only 1 story (Alert) was actually using it was indeed broken.

😵 While fixing this and updating that particular dependency, I found several **unused** dependencies and other **misplaced** (that should have been dev deps and not deps and/or viceversa).

😵 We have a "ThemeProvider" component besides the styled-components one, this is confusing and it confuses the HSDS imports babel plugin, this component is not used anywhere external as far as I can see (checked hsapp, beacon) and it's used on a couple of components inside HSDS.

😵 For some reason tests started failing complaining about missing `@helpscout/fancy` from baseIllo, I checked `helpscout/hsds-illos` dependencies, and indeed it is listed as a dev depency and peer dependency, not sure why it decided to complain about it now though...

## Solution

💃🏻 Fixes the figma addon issues
💃🏻 Adds figma links to the few stories we have with MDX:
|  🕺🏻 Accordion
|  🕺🏻 Alert
|  🕺🏻 Attachment
|  🕺🏻 Avatar
|  🕺🏻 AvatarList
💃🏻 Moves a bunch of dependencies to dev dependencies
💃🏻 Removes a bunch of dependencies
💃🏻 Brought back @helpscout/fancy (needs to be removed from hsds-illos first)
💃🏻 Fixes eslint configuration file, it had some invalid and repeated rules
💃🏻 Removes polyfill.js for Jest, as those are no longer needed
💃🏻 Renames ThemeProvider component to "SimpleThemeProvider" to avoid confusion between it and the styled components one, this fixes the issue with the babel plugin. Eventually we should remove this component and just use the styled components one.

## Testing

✨ Tests pass
✨ Storybook runs, figma links are displayed on the components mentioned above
✨ Cypress tests pass
✨ Builds fine


.eslintrc.js
```diff
@@ -9,7 +9,12 @@ module.exports = {
     jest: true,
     node: true,
   },
-  globals: ['cy', 'Cypress', 'expect', 'assert'],
+  globals: {
+    cy: 'readonly',
+    Cypress: 'readonly',
+    expect: 'readonly',
+    assert: 'readonly',
+  },
```
malikalim, line 17: @tinkertrain by moving these to readonly will they impact cypress tests at all?

cc: @ryan-mulrooney

malikalim:
also confirmed design tab looks like it's working now @digitaldesigner https://c.hlp.sc/12uN42l2

.eslintrc.js
tinkertrain: Hi @malikalim, at some point Eslint changed how these globals are declared, so my guess is this rule has not been taking any effect for a while.

What these declarations do is telling Eslint not to tell off the developer when they use these globals in their code (otherwise eslint tells you: "the `cy` variable has not been declared"), `readonly` means that a developer should not reassign their value (by doing something like `cy = 100` in their code), and if they do, Eslint will warn them about it.

Long story short, it doesn't impact the code 👍

PR #852, opened 2020-06-24
Fixed differences between v2 and v3 in Accordion component

Charca:
## Summary

@knicklabs spotted a few differences between the Accordion component implementation in v2 and v3. This PR solves these issues.

1. Sortable drag-handle updated to the right icon and size.
2. Fixed position of drag-handle while dragging an expanded item.
3. Fixed issue where chevron icons changed orientation when dragging them.

### DIFF
--- a/src/components/Accordion/Accordion.Title.jsx
+++ b/src/components/Accordion/Accordion.Title.jsx
@@ -103,7 +103,7 @@ const AccordionTitle = props => {
     isSortable,
   } = useContext(AccordionContext) || {}
 
-  const isLink = props.href || props.to || isSorting
+  const isLink = props.href || props.to
   const isIconOpen = isLink ? false : isOpen
 
   const componentClassName = getComponentClassName({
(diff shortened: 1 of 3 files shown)


src/components/Accordion/Accordion.Title.jsx
```diff
@@ -103,7 +103,7 @@ const AccordionTitle = props => {
     isSortable,
   } = useContext(AccordionContext) || {}
 
-  const isLink = props.href || props.to || isSorting
+  const isLink = props.href || props.to
```
Charca, line 106: @plbabin @tinkertrain this was one of the differences between v2 and v3, and I couldn't figure out why we added this condition here to set the `isLink` prop to true if the Accordion was sorting. Any of you remember why we added this? I've removed it because it was causing this issue: https://c.hlp.sc/DOuAD4wj (chevrons of non-sorted items change to being right-facing when an item is being sorted).

tinkertrain: Hmm I don't know but it seems like the right move is to remove it, I see no valid reason for it (if there is, maybe it belongs somewhere else)

plbabin: not so sure to be honest. Maybe we thought it was when the accordion was sortable, to apply a different caret (and mis read that it was during the sorting action) 🤷 .

let's keep your change, and adjust if something is not right later

PR #854, opened 2020-06-25
Fix how Tooltip renders custom title content

cocoabythefire:
**Original Issue:**
Updating to HSDS v3 in HSApp, I noticed that a Tooltip which is passed a custom element as the title property, was no longer rendering correctly (it was displaying as "Object object")

Screen Shot 2020-06-25 at 12 05 48 PM
Upon investigating, I determined that the Tooltip renders title via a span with dangerouslySetHTML which does work fine with Tippy's render when the title is in fact a string, but was completely breaking for any other type of custom title content.

**Solution:**
I added a check for whether the title attribute being passed into Tooltip is a string, and if so, go ahead with the implementation as is, but otherwise we pass through the custom content directly to the Tippy render function. This seems to work great but I'm new to Tippy so not sure if there's any downside to doing this.

Screen Shot 2020-06-25 at 12 03 05 PM

**Storybook:**

Two new stories
- Input Custom Error state
- Tooltip Custom content

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.2.10",
+  "version": "3.2.11",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.2.11",
+  "version": "3.2.13",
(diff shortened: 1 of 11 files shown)


cocoabythefire:
@plbabin so this works, but open to however way you want to solve it. Ultimately it wasn't our use of Tippy that was incorrect but more just that the way we passed the title into Tippy's render method, it seems to just assume it will always be a string but we have use cases in HSApp where we pass react elements as the title (errorMessage on Input -> Tooltip title)

PR #858, opened 2020-07-02
HSAPP-2429 Editable field with multi-values bug

cen10:
[JIRA](https://helpscout.atlassian.net/browse/HSAPP-2429)

## Problem
For multi-value editable fields like Email or Website on the Customer Property page, a couple of bugs existed. [See a video with verbal walk-through of both bugs here.](https://c.hlp.sc/geuwORXY)
1. If you have a single email on the page, highlight it, press the delete key, and then click outside of the field, you get an error message saying the field can't be empty. 
2. If you have multiple emails on the page, highlight one, press the delete key, and then click outside of the field, it appears that the email is deleted. However, on refresh, the email is restored.

The above problems also exist for the Website field in the customer profile, although it's not demoed in the video.

Solution:
For problem 1, we needed to update logic to remove a check for the length of the array stored in `removedEmptyFields`:
Including the check for "removedEmptyFields.length > 0" meant that when there was a single email visible and the user removed it via backspace and clicking outside of the input, we would never enter the code block that handled discarding empty fields. This is b/c the removedEmptyFields variable ends up being a list of the fields that remain after an empty field is removed, and in the case of only having one email in the list, there are no emails left once you remove it, so 'shouldDiscardEmpty' is always going to be false and we were never going to hit the DELETE api call, we would make a PUT instead.

Additionally for problem 1, we needed to remove the `_id` property from the data before it is passed along to the api call, again, so that a DELETE call can be made instead of a PUT. The logic for which call to make hinges at one point on the presence of that property.

For problem 2, we needed to update logic so that we were actually telling the DELETE request what field to delete. Previously, we were passing in an empty array, which meant that while the UX updated appropriately, the backend was never made aware of the request.
 
## Testing

This should be tested using the beta release - it's probably quickest to pull down the hs-app branch `editable-field-email` with the updated package-json pointing to the beta release and test using proxypack.

Note: I noticed that if you highlight an email, press the delete key, and then try pressing "enter" to save the changes, nothing happens. Initially I thought this was a bug, but [there's a comment explicitly saying that's what should happen](https://github.com/helpscout/hsds-react/blob/master/src/components/EditableField/EditableField.jsx#L510), so I'm leaving that alone.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.2.16-0",
+  "version": "3.2.15",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.2.15",
+  "version": "3.2.16-0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.2.15",
+  "version": "3.2.16-0",
   "private": false,
   "main": "dist/index.js",
   "module": "dist/index.es.js",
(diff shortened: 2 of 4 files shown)


cen10:
Reminder to me to delete beta release commit

PR #859, opened 2020-07-03
HSDS-158 Tooltip Fix: Text breaks mid-word

knicklabs:
[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-158)

# Problem

When tooltips have long lines of text, the text breaks at arbitrary points, including mid-word. See the screenshot below.

Screenshot of Problem

# Solution

I changed the value of the `word-break` style property applied to the tooltip to be `break-word` so that words are preserved if possible. 

Screenshot of Solution

# Testing

You can try it out with various text content using the knobs in the add-ons pane on the [Tooltip Story](https://deploy-preview-859--hsds-react.netlify.app/?path=/story/components-overlay-tooltipv2--default).


cen10:
Assigning myself so I can create a release for this when it's approved, since Nickolas is out all week.

PR #856, opened 2020-06-30
Upgrade React and React DOM peer dependencies to 16.13.1

cocoabythefire:
This PR will be held until after HSDS v3 migrations in HSAPP have all been merged for testing in HSApp

This is just a package change... and to limit scope it ONLY includes the React and React-DOM updates, no other packages are updated unless we identify mandatory updates for compatibility.

We will need to test and look for any components not functioning/displaying as expected or errors/warnings in the console, or lint errors newly thrown, but so far I'm not seeing anything unusual.

PR #860, opened 2020-07-08
Removes Chat components, upgrades dependencies to match HSApp

tinkertrain:
This PR:

- Removes ChatSidebar
- Removes ChatInbox
  - Both moved to HS App: https://github.com/helpscout/hs-app/pull/8613
- Removes ChatTranscript
  - Moved to HS App: https://github.com/helpscout/hs-app/pull/8754
- Cleans up dependencies:
  - Removes a couple of unused dependencies
  - Updates Prettier config file
- Match react and jest versions from: https://github.com/helpscout/hs-app/pull/8176/files#diff-2ac18b788648958ababf0b8169011dfd
  - Jest 26: fixes tests affected with the upgrade
    - Notes:
		- Use `jest.runAllTimers()` instead of `jest.runAllImmediates()` (https://jestjs.io/docs/en/jest-object.html#jestrunallimmediates)
        - `document.activeElement` doesn't work as it used to, so different strategies to test are required (check VerificationCode and Input tests for a couple of ways to test the same)
- Adjusts some `PropTypes` declarations based on HSApp needs

### Testing

- All Tests pass
- Storybook runs without issue
- Builds fine


package.json
```diff
@@ -80,12 +80,13 @@
     "make:version": "node ./scripts/make-version.js"
   },
   "peerDependencies": {
-    "react": "^16.8",
-    "react-dom": "^16.8",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
     "styled-components": "^5"
```
plbabin, line 85: maybe we should target 5.1 ?

plbabin:
We might need to sync with this PR #856 and see if we need to close it or not. Maybe we want to limit the scope to just a react version update first, and then migrate the rest with this PR. Since you did all the work @tinkertrain, I think you can decide what is best moving forward

cocoabythefire:
> We might need to sync with this PR #856 and see if we need to close it or not. Maybe we want to limit the scope to just a react version update first, and then migrate the rest with this PR. Since you did all the work @tinkertrain, I think you can decide what is best moving forward

yeah feel free to close mine - i just threw that up to make sure we wouldn't forget, but it's redundant to these changes here

package.json
tinkertrain: agree 👍

tinkertrain:
@plbabin @cocoabythefire Woops, I didn't see that PR... React was on `^16.8` which means that it was already installing >16.13.1, so there should not be any impact, I'll close the other PR

PR #862, opened 2020-07-10
Add disabled support to SelectDropdown component

TerrenceLJones:
This add disabled styling support to the `SelectDropdown` component. 

### Before (label text)
Screen Shot 2020-07-10 at 17 20 03

### After
Screen Shot 2020-07-13 at 18 09 31


TerrenceLJones:
> Looks good. Not a blocker, but I do wonder if we should dim the icon as well when the select is disabled?

@plbabin ahh, great catch!

PR #866, opened 2020-07-21
Portal / React 16 API

plbabin:
### Issue
Portal were introduce in HSDS while we were supporting React 15. We had a lot of workaround to make portal work correctly, but now we start to hit some issue with the new Frame component & styled-components. 

Because the solution was custom, each Portal were rendered outside the react DOM tree, causing issue where the portaled component was not accessing the right Stylesheet Manager. The side effects of that was that no Portal components rendered within an iframe was styled with a scope, and all styles were injected at the root document head.

### Solution
We start `Portal.jsx` from the ground up, and we are now leveraging [createPortal](https://reactjs.org/docs/portals.html) to handle the Portal injection in the dom. We also spend some time to integrate the HSDS GlobalContext, so we could automatically wrap the Portal with the actual scope, resolving an issue that we had often while migrating to v3.

We also improve the anchor selection to pick the right body based on the Portal context (renderTo prop, portal.container, document.body or frameContext).

The last thing we did was to reorganize the `Frame`, just to make it a bit more clearer what was going on with the Stylesheet manager (and to use `useContext` instead of a render prop).

#### PortalWrapper
This is an HOC that wrap any component we want to prepare to use Portal. It will merge some props from Portal and from the composed component together, and it will pass down some state and callbacks too. It would be nice long term to move that to be a hook but we would have to change the fundamentals of how we integrate Portal within a component:

```javascript
const Modal = () => {
  const [isOpen, closePortal, openPortal] = usePortal()
  return <ModalUI {...props} isOpen={isOpen} />
}

```

But that one It will be for another time


plbabin:
> Looks great! I just found one issue but I'm sure it's a Storybook thing. Sometimes when you move from the Frame story to a different one (like the Modal story), when you come back to the Frame story all styles are missing. Here's an example: https://c.hlp.sc/YEupxZYN
> 
> It really looks like a Storybook thing though, so probably it's nothing to worry about, right?

that looks like a storybook issue. I think to be safe , when we'll upgrade in hs-app, we should do a smoke test to validate that Portal is working correctly in multiple place (and there is no UI issue like you just found)

PR #869, opened 2020-07-23
Dev docs

tinkertrain:
# Feature

Following the release of the documentation for v3 [here](https://github.com/helpscout/hsds-react/releases/tag/v3.4.0) this PR updates the developer documentation with the most up-to-date information.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '8.14.0'
+  - '12.13.0'
```
tinkertrain, line 3: @plbabin Will this be the reason for the travis' failures? is there a reason not to upgrade it?

plbabin: yeah that's probably the reason, dunno why it was not updated 🤷

PR #872, opened 2020-07-29
Fixing ChoiceGroup, RadioCard, Choice in preparation for Checkmark

tinkertrain:
# Feature (and problems)

**Feature** 👉 We need a new component, `Checkmark`!

The solution is very similar to `RadioCard`, so much this component could be named `CheckCard`...

`RadioCard` normally gets used with `ChoiceGroup` like so:

```jsx
<ChoiceGroup>
      <RadioCard icon="fab-chat" value="chat" />
      <RadioCard icon="fab-antenna" value="antenna" />
      <RadioCard icon="fab-buoy" value="buoy" />
      <RadioCard icon="fab-search" value="search" />
      <RadioCard icon="fab-question" value="question" />
</ChoiceGroup>
```

We'd like to continue using this pattern

### Problem 1: multiSelect

`ChoiceGroup` has a `multiSelect` option, that like the name indicates allows you to select more than one choice in the group, usually with radio inputs you turn this off and with checkbox you turn it on.

A while back, @cocoabythefire mentioned she was having some issues with this and when we looked at the code we saw that indeed something a little strange was happening here, she managed to fix somehow in place in HS App.

This PR fixes the issue:

- Allows the correct use of `multiSelect` via a prop
- `multiSelect` defaults to `true`, the reason I chose `true` is because in most uses in HS App it was declared as false where needed. I have another PR for HS App (https://github.com/helpscout/hs-app/pull/8824) that adds the prop in the rest of the cases where necessary.
- ChoiceGroup was "magically" setting this for you if it detected a `Radio` or `RadioCard` setting it to `false` in those cases, I removed this because the implementation was brittle and making some assumptions, this can easily be handled by the user by just using the prop manually.

## Problem 2: RadioCard focus

The focus state on RadioCard is broken at the moment in multiple scenarios, and the styles do not match what we have in figma https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=2582%3A9509

I removed the `FocusUI` way of rendering the focus state (using a `div` and `setState` under the hood) in favour of a css solution that correctly renders a focus state (try using tab now 😉)

## Problem 3: No keyboard interaction

Now that focus is working properly, pressing Space or Enter on a Choice (RadioCard, Radio, Checkbox) should interact (select/deselect) just like you would expect.

This PR implements this functionality.

## Problem 4: Styles

Besides the focus styles, we had a slight mismatch with styles and animations based on the codepen and figma links below

## Testing

Check the following stories:

- Checkbox: https://deploy-preview-872--hsds-react.netlify.app/?path=/story/%F0%9F%A7%AC-components-forms-checkbox--default-story
- Radio: https://deploy-preview-872--hsds-react.netlify.app/?path=/story/%F0%9F%A7%AC-components-forms-radio--in-a-group
- RadioCard: https://deploy-preview-872--hsds-react.netlify.app/?path=/story/%F0%9F%A7%AC-components-forms-radiocard--with-headings

Test using the keyboard

## Relevant Links
Jira: https://helpscout.atlassian.net/browse/HSDS-157
Spec: https://paper.dropbox.com/doc/Spec-Custom-Avatars-for-Beacon--A4mkPLCiSo7B4lkN3xccvN8rAg-c2gwAHcAr6dvs2amX6qE6
Figma: https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=5015%3A49
Codepen prototype: https://codepen.io/team/helpscout/full/5d458fb128e71e612ef1ad7202b178cd

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -192,22 +192,20 @@ Button.propTypes = {
   isSuffix: PropTypes.bool,
   /** Applies the specified style to the button.
    * 'primary': Blue button. Used for primary actions.
-   * 'primaryAlt': Purple button. Used for primary actions.
    * 'secondary': White button with a border. Used for secondary actions.
-   * 'secondaryAlt': White button with a green border. Used for secondary actions.
+   * 'tertiary': White button with a green border. Used for secondary actions.
    * 'default': Borderless button. Used for subtle/tertiary actions.
    * 'link': Button that looks like a `Link`. Used for subtle/tertiary actions.
    */
   kind: PropTypes.oneOf([
     'primary',
-    'primaryAlt',
     'secondary',
-    'secondaryAlt',
+    'tertiary',
     'default',
     'link',
   ]),
-  /** Sets the size of the button. Can be one of "sm", "md" or "lg". */
-  size: PropTypes.string,
+  /** Sets the size of the button. */
+  size: PropTypes.oneOf(['sm', 'md', 'lg']),
   /** A special property that... spins the button if `isLoading`. */
   spinButtonOnLoading: PropTypes.bool,
   /** Applies state styles to the button.
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -205,7 +205,7 @@ Button.propTypes = {
     'link',
   ]),
   /** Sets the size of the button. */
-  size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'xs', 'xss']),
+  size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'lgxl', 'xs', 'xxs']),
   /** A special property that... spins the button if `isLoading`. */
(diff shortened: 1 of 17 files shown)


tinkertrain:
Thanks @digitaldesigner !

I fixed all of the things you mentioned.

On the last one, I removed the knob for "isFocused" because for some reason it wasn't working as it should be, might be a "knobs addon" bug, in the story code you can see that I added the 'isFocused' prop to the buoy and selecting and deselecting other options work as it should be:

Screen Shot 2020-07-30 at 1 18 35 PM

Let me know about the alignment, I think it looks better now, but you tell me if that is enough.

And one last thing, about the green background in the choice, you might want to update the figma file for the RadioCard (https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=2582%3A9509)

Thanks again!

PR #874, opened 2020-08-03
CheckMarkCard

tinkertrain:
# Feature

New component: `CheckMarkCard` 🥳 

Similar to `RadioCard` but with a "checkbox" style functionality.

## Solution

1. Adds a new prop to `ChoiceGroup` => `multiSelectLimit` which allows you to pass a number and limit the number of items selected, once the limit is reached the rest get disabled; deselecting an item and thus being under the limit, re-enables the choices.
2. Adds new component `CheckMarkCard`, which is used in a similar manner to `RadioCard`
- The main difference with this is that the user should provide the content as opposed to using multiple props for rendering:

```jsx
// RadioCard

<RadioCard
      icon="fab-chat"
      content="some content"
      heading="Chat"
      value="chat"
      maxWidth="214px"
/>


// CheckMarkCard
<CheckMarkCard label="Derek" value="derek">
        <Avatar size="xl" image={AvatarSpec.generate().image} name="Derek" />
		<div> Name </div>
		<div> Location </div>
</CheckMarkCard>

/*
	It's up to the user to decide what to render and style accordingly
*/
```

4. Use it inside of `ChoiceGroup` in the same way as `RadioCard`:

```jsx
<ChoiceGroup multiSelectLimit={3}>
	<CheckMarkCard label="Derek" value="derek" />
	<CheckMarkCard label="Hansel" value="hansel" />
	<CheckMarkCard label="Matilda" value="matilda" />
</ChoiceGroup>
```

5. While using the component generator I found some mistakes in the templates, they're fixed here.

### Depends on:
- https://github.com/helpscout/hsds-react/pull/872
- https://github.com/helpscout/hs-app/pull/8824

### Relevant Links:
- [Jira](https://helpscout.atlassian.net/browse/HSDS-157)
- [Spec](https://paper.dropbox.com/doc/Spec-Custom-Avatars-for-Beacon--A4mkPLCiSo7B4lkN3xccvN8rAg-c2gwAHcAr6dvs2amX6qE6)
- [Figma](https://figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=5015%3A49)
- [Codepen prototype](https://codepen.io/team/helpscout/full/5d458fb128e71e612ef1ad7202b178cd)

### Screenshots

> A little update on the CheckMarkCard components not reflected on these screenshots is the extra grey border when checked that has now been removed.

### CheckMarkCards with content, inside a ChoiceGroup with a multi-select limit:

> Not sure why in this screenshot items seem to flash on selection, try the story to see it properly


src/components/CheckMarkCard/CheckMarkCard.jsx
```diff
@@ -0,0 +1,163 @@
...
+      'c-CheckMarkCard',
+      className,
+      checked && 'is-checked',
+      disabled && 'is-disabled'
+    )
+  }
+
+  setInputNodeRef = node => {
+    this.inputNode = node
+    this.props.inputRef(node)
```
TerrenceLJones, line 45: @tinkertrain Just so I can understand this pattern...why would we need both a `inputRef` and an `innerRef` here?

src/components/Checkbox/Checkbox.jsx
```diff
@@ -11,7 +10,7 @@ class Checkbox extends React.PureComponent {
 
     return (
       <Choice
-        {...getValidProps(rest)}
+        {...rest}
```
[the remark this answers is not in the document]
tinkertrain: @Charca Yes, good call! I had to remove it from `Checkbox` as ti had some issues, I think that is a great rule of thumb. Thanks!

src/components/CheckMarkCard/CheckMarkCard.jsx
tinkertrain: No reason! I will remove one, not sure why RadioCard has it this way, after refactoring and removing several thing from there this one escaped me and I brought it over, I'll fix it in both components

tinkertrain:
> Looking great functionally, and I only have a couple of tweaks to the visual design:
> 
>     1. The default (inactive) state of the cards should have a border, currently they don't (see [c.hlp.sc/mXuynEG1](https://c.hlp.sc/mXuynEG1))
> 
>     2. The selected state adds a blue border, however it should replace the existing one. Currently the grey stroke is still visible (see [c.hlp.sc/P8ubpWKx](https://c.hlp.sc/P8ubpWKx))

@digitaldesigner Thanks for looking at it!

I made the updates, regarding the borders and shadows I noticed they are slightly different from the RadioCard (I copied some styles from it thinking they were the same):



Do you think they should be the same?

digitaldesigner:
Great catch @tinkertrain — I think they should be consistent. See attached image and [here's the corresponding Figma file](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=6103%3A13568)
Radio Card

I guess the larger question is how we move towards a more consistent set. In the Figma files we use one of 7 different depths, each of which has a matching hover state. See [HSDS Depth Guide](https://cdpn.io/team/helpscout/debug/jOWBppq)

Depth Examples

tinkertrain:
> Great catch @tinkertrain — I think they should be consistent. See attached image and [here's the corresponding Figma file](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=6103%3A13568)
> Radio Card
> 
> I guess the larger question is how we move towards a more consistent set. In the Figma files we use one of 7 different depths, each of which has a matching hover state. See [HSDS Depth Guide](https://cdpn.io/team/helpscout/debug/jOWBppq)
> 
> Depth Examples

That looks great, one thing we could do is having the "depth set" as UI components or set of styles inside HSDS, that then we can just use all over HSDS to ensure consistency and make it easier fo everyone, then in figma you can leave the little comments like you do with colours `/* depth 300 */`.

Let's continue this conversation off here, I'll release this and then have another PR dedicated to **Depth**

PR #876, opened 2020-08-10
Updates for beacon

Charca:
## Features

This PR updates a few components based on a few issues I've found while upgrading Beacon to hsds-react v3:

- Updated styling of `ArticleCard` and `MessageCard` components.
- Fixes icon names.
- Adds an optional `tooltipAppendTo` prop to the `Input` component to control where the tooltip is rendered.
- Fix an issue within the FrameComponent where the cached selectors are passed to the new sheet in StyleSheetManager. A fix already exists in styled-component, but it's not released yet (https://github.com/styled-components/styled-components/pull/3159)
- Fix an issue with the mountSelector in Portal. It will only fallback on the frame document only if there is no existing body from  the FrameContext


plbabin:
I added some new fixes that are required in beacon. I updated the PR description with the changes

PR #877, opened 2020-08-13
ChoiceGroup: onChange and onEnter callbacks always receive value wrapped in array but it only should be for multiSelect

knicklabs:
**[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-164)**

##Problem/Feature

A regression was introduced in https://github.com/helpscout/hsds-react/pull/872 which was released in [v3.5.2]( https://github.com/helpscout/hsds-react/releases/tag/v3.5.2) leaving us unable to upgrade to 3.5.2 or later. The regression was that ChoiceGroup components, when used as a single-select, would receive the single value wrapped in an array via the `onEnter` and `onChange` callbacks. Existing projects consuming this code expected the values to NOT be wrapped in an array.

##Solution

Update the `handleOnChange` and `handleOnEnter` component methods to invoke `onChange` and `onEnter` respectively with the value NOT wrapped in an array when the `multiSelect` prop is set to false. The corresponding test was also updated.

### DIFF
--- a/src/components/ChoiceGroup/ChoiceGroup.test.js
+++ b/src/components/ChoiceGroup/ChoiceGroup.test.js
@@ -54,7 +54,10 @@ describe('ChoiceGroup', () => {
           <Radio />
         </ChoiceGroup>
       )
-      const input = wrapper.find('.c-Radio').first().find('input')
+      const input = wrapper
+        .find('.c-Radio')
+        .first()
+        .find('input')
 
       input.simulate('blur')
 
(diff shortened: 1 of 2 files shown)


src/components/ChoiceGroup/ChoiceGroup.test.js
```diff
@@ -185,18 +209,27 @@ describe('ChoiceGroup', () => {
...
-      expect(spy).toHaveBeenCalledWith(['derek'])
+      expect(spy).toHaveBeenCalledWith('derek')
 
       input2.simulate('change', { target: { checked: true } })
-      expect(spy).toHaveBeenCalledWith(['hansel'])
+      expect(spy).toHaveBeenCalledWith('hansel')
 
       input3.simulate('change', { target: { checked: true } })
-      expect(spy).toHaveBeenCalledWith(['mugatu'])
+      expect(spy).toHaveBeenCalledWith('mugatu')
```
knicklabs, line 232: Sorry for the formatting noise. Lines 226, 229, and 232 are the only changes in this file.

PR #883, opened 2020-09-09
Datepicker component v1

tinkertrain:
#  📅 Datepicker

[Figma](https://www.figma.com/file/TgNl8UMoxkDTpCfhdg77YX/Date-Picker?node-id=0%3A1)
[Paper](https://paper.dropbox.com/doc/HSDS-Date-Picker--A7O_8gGGkZIc1Sf8AkYT7gZwAg-KN1p5cY7zSDN0oK4mOCgc)

New component!

For version 1, we support:

- 1 calendar view
- Choose a single date
- Navigate month by month
- Navigate by period (months in a year, a range of years)
- Highlight current day
- Ability to prevent navigating to the future
- Use keyboard throughout

> Note: Uses [@datepicker-react/hooks](https://www.npmjs.com/package/@datepicker-react/hooks) to handle most of the calendar logic (not UI)

The usage is up to the user (use input, editable fields, other...)

[Story](https://deploy-preview-883--hsds-react.netlify.app/?path=/story/🧬-components-dropdowns-datepicker--default-story)

### Regular usage and navigation
datepicker

### Choose start of week
datepicker - start of week

### Prevent selecting dates in the future
datepicker - no future dates

## Testing

This PR introduces the react testing library to HSDS, as this is what we use in HS App.


digitaldesigner (changes requested):
Stunning work as usual! I had a few tweaks to make now I'm able to see (and feel) it in code... I actually ended up coding a prototype to get my head around some of the finer details. [Here's the link](https://cdpn.io/team/helpscout/debug/02495b058c4ab913343200a1666037e1) if the below doesn't make sense (especially around focus/active states):

## Visual
- Switch the font family to system (system is better for representing numbers IMO)
```
font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
- Apply anti-aliasing to the numbers (this gives a better color differentiation and helps visually distinguish the bolder characters). See prototype for type of aliasing
- Day titles and arrows to `charcoal.500` (currently black)
- No hover state to disabled numbers (previous/next month overlap). I deliberated over this, and although they should remain selectable, I think it's visually a lot less cluttered when they don't have a blue hover state

## Structural
- If an entire row is disabled, could we remove that row? This wouldn't apply when there is a range since we'd want both months to be equal heights
- Trivial detail, but the entity between year ranges should be an ["en dash"](https://www.thepunctuationguide.com/en-dash.html). E.g. [2016–2027](https://apps.timwhitlock.info/unicode/inspect?s=2016%E2%80%932027)

## Interaction
- Can we prevent the focus state on click? Only blue outline on tab focus
- The currently selected month in the month viewer (same for the year viewer) should have an active state (i.e. blue background)
- Hover state on selected date/year etc should be the selected state (not lighter blue)
- `cursor: pointer;` for anything clickable

tinkertrain:
@digitaldesigner Thanks a lot for looking at it! I pushed an update, I think I addressed all of your points, let me know 😀

digitaldesigner (changes requested):
I'm only nit-picking now because you're so incredibly close to perfection. If you don't have time for the below let me know and I'll hit the :shipit: button:

- Remove font-smoothing on today and selected items (will make the text a little bolder)
- For the month titles "September 2020", could we use the app font ("Aktiv") or fall-back to Helvetica? Currently it's Arial, which I dislike
- Arrows, month title and month/year navigation still have the focus states on click (and shouldn't). How you've amended them to work on the date numbers is perfect though
- Feels like in the interest of consistency, the arrows should have the pale blue background and also be blue on hover

digitaldesigner (changes requested):
Visually all looking really great, thanks @tinkertrain — looks like there might have been a small regression on the year/month selected state though. In this view (below) the currently active month and year should be the blue selected state (it was definitely correct at some point recently). You're a legend

Screen Shot 2020-09-14 at 8 27 36 AM

tinkertrain:
@digitaldesigner Thanks! Not sure what happened but it should be working again now...

digitaldesigner (commented):
Not seeing any difference — here's a little video: https://c.hlp.sc/o0u8dgnY

tinkertrain:
@digitaldesigner Ahhh, I see what you mean. The reason is that the month/year range selection is not technically a "selection" just navigation, without actually selecting a day there is no selection to show.

I can implement a "temporal" state where I save those month/year with no day selections but I think we might end up in a confusing state, consider:

1. No date selected
2. Use the "deep nav" button and select the month april 2020
3. You get taken to that month in the calendar
4. Still no day selection, repeat 2 and select another month (may)
No problem so far
5. Now select a day (may 5th, 2020)
6. Use the deep nav, you see the month selected (and you can verify the year is also selected by clicking the deep nav button again)
7. Now, if I click another month or year:
  a) As it is now: you will be taken to that month/year in the calendar, but your original selection (may 5th 2020) is still "active", you only change a date by selecting a specific day.
  b) If we were to store the month/year as selections, we'd have to clear the original selection, otherwise switching between the calendar view and the "period" view won't make sense. Clearing the selection might be risky (perhaps leading to unintended changes in the date selection in a form for example).

The way I see it (and of course I might be wrong) is that the "period view" is there to allow navigating the calendar faster, functionality wise I don't see it different to using the `<` & `>` buttons to go the next or previous month.

Let me know what you think!

digitaldesigner (approved):
Good points about the "deep nav" — I trust your judgement there. Looks awesome to me

src/components/Datepicker/Datepicker.PeriodCalendar.js
```diff
@@ -0,0 +1,157 @@
...
+      <PeriodUI
+        role="grid"
+        className={classNames(
+          'c-Period',
+          mode === NAVIGATION_LEVELS.YEAR_BY_YEAR
+            ? 'is-mode-months'
+            : 'is-mode-years'
+        )}
+      >
+        {mode === NAVIGATION_LEVELS.YEAR_BY_YEAR
```
plbabin, line 78: the ternary condition is a bit confusing here, just because it's on multiple line. maybe we could extract that to a function to render the right content?

src/components/Datepicker/Datepicker.js
```diff
@@ -0,0 +1,159 @@
...
+    exactMinBookingDays: minBookingDays === 1,
+    maxBookingDate: !allowFutureDatePick ? new Date() : undefined,
+  })
+
+  /**
+   * This section performs a `getDerivedStateFromProps` style logic where we
+   * store the prop in state (as prevDate) so we can compare in the next render
+   * and only if the prop changes this gets run again
+   */
+  if (
```
plbabin, line 56: useEffect is here for you, for exactly that reason. It will execute code based on a variable, when those change

you might have tried it, and determine that it was not doing what you need, and that's fine.

a possibility would be to separate the state for each key:

```javascript
const [startDate, setStartDate] = useState(getJSDateFromString(startDateProp))
const [endDate, setEndDate] = useState(getJSDateFromString(endDateProp) || startDate)
```
qq: Is there a moment where focusedInput could change? if not you might just use the constant in useDatepicker. 

then useEffect for updating the date
```javascript
useEffect( () => {
  setStartDate(getJSDateFromString(startDateProp))
  setEndDate(getJSDateFromString(endDateProp))
}, [getValidDateTimeString (startDateProp), getValidDateTimeString (endDateProp)])
```

and finally in handleDateChange you could set each date state:

```javascript
function handleDateChange(data) {
    setStartDate(data.startDate)
    setEndDate(data.endDate)
    onDateChange(data)
  }
```

I did not test that, it could be different dependencies for useEffect. You know the code better than me, so I will let you judge if that update would be beneficial in the end 🎉

plbabin (approved):
incredible work here! That component looks soooooo good 🎉 🌮 .

Left some suggestion, but overall nothing major!

src/components/Datepicker/Datepicker.js
tinkertrain: Using `useEffect` is how I first implemented this part, and it worked ok-ish: I didn't like the fact that it would always trigger a re-render on the component initial mounting.

There are 2 things that need to happen when the prop changes: 
1. Select the date (this is what the `setDates` call is responsible for)
2. Navigate the calendar to that date (done by the function `goToDate` provided by datepicker-hooks, I believe this uses setState under the hood to set some "navigation" state)

Number 2 is an unnecessary call most of the time except for this particular case (it gets called for you in most scenarios, like navigating month by month), without it the new date would be selected but you would have to navigate manually to the correct year/month to see it.

So what ends up happening with the use of `useEffect` is that you set the dates using `setState` and then after the render (or mounting) happens, it runs the `setState` and `goToDate`, triggering a re-render.

The code here ensures not having to do something more complicated just to be able to use `useEffect` (given that `goToDate` is a dependency for example, which means something weird like having to use `useCallback` for it, or having to add extra conditionals to check the "navigation" state of the calendar, etc), and it performs as expected with no re-renders and no effect other than if a different date is passed as a prop, and my guess is that changing the date via prop in something other than the initial render rarely happens.

Hope that clarifies things a bit, thanks for the notes!

src/components/Datepicker/Datepicker.PeriodCalendar.js
tinkertrain: No problem, done!

src/components/Datepicker/Datepicker.Day.js
```diff
@@ -0,0 +1,100 @@
...
+      })}
+    >
+      <time dateTime={getValidDateTimeString(date)}>{dayLabel}</time>
+    </DayUI>
+  )
+}
+
+Day.defaultProps = {
+  leading: false,
+  trailing: false,
```
knicklabs, line 97: Small nitpick, but I think -- especially with function components -- that we should not use `defaultProps` and use the default param values, which are now native to JavaScript. It keeps the values closer to their use and I'd prefer native solutions over library-based one.

knicklabs (approved):
This looks great @tinkertrain! Well done 👏

src/components/Datepicker/Datepicker.Day.js
tinkertrain: Didn't really thought about this before, makes total sense! Thanks for mentioning it

knicklabs: @tinkertrain it's in our JS Standards document in the Best Practices section :)

PR #887, opened 2020-09-29
Remove cyan

tinkertrain:
# Feature

In this PR we remove the library `@helpscout/cyan` and use [React Testing Library](https://testing-library.com/) instead (this library got introduced in a previous PR).

- Rewrites most cyan tests to testing library (just a couple to enzyme)
- Removes tests and/or assertions made obsolete by testing library or that were pointless
- Rewrites a few enzyme tests to testing libraries (not all of course, that would be madness!)
- Updates testing configuration
- Updates the generator test template to encourage the use of testing library instead of enzyme
- Updates testing docs
- Updates illos to latest and removes fancy from deps


plbabin (approved):
tests are passing, the house is in order, what else there is left to say?

happynew-5kFW1M5uFrMU8RDq14

PR #890, opened 2020-10-13
Datepicker bug fixing

tinkertrain:
This PR fixes a few issues we've found on the `Datepicker` component:

## Issue 1

Weird and buggy behaviour when on different timezones, it seemed to only work well in Barcelona's timezone  ¯\_(ツ)_/¯

### Solution

Refactor and fixed how the active month was calculated and consumed by `Month`

## Issue 2

Calculating "Leading Days" was not working on multiple scenarios and Datepicker did not support start days other than Sunday and Monday almost at all

### Solution

Changed how the amount of leading days are calculated.

## Issue 3

When the month was december, clicking on a trailing day (meaning a day belonging to January in the next year) would select January 1st of the same year, instead of the next year.

### Solution

Make sure we take into account if we are in this situation and increase the year accordingly.

--

Added and updated tests for each case


cen10 (approved):
Looks great -- I'm also not seeing the NaN bug so 🤷‍♀️ maybe it's a non-issue now.

PR #891, opened 2020-11-02
Add Depth to components

tinkertrain:
# Feature

Use the new Depth mixins in HSDS components:

d200 - FluffyCard
d300 - Page
d300 - RadioCard
d400 - CheckmarkCard
d400 - ArticleCard
d400 - Attachment with preview
d600 - Notification
d600 - MessageCard
d700 - Modal

### DIFF
--- a/src/components/ArticleCard/ArticleCard.css.js
+++ b/src/components/ArticleCard/ArticleCard.css.js
@@ -1,17 +1,7 @@
-import Card from '../Card'
 import styled from 'styled-components'
-import { getColor, rgba } from '../../styles/utilities/color'
-
-export const config = {
-  hover: {
-    baseShadow: `
-      0 1px 3px 0 rgba(0, 0, 0, 0.1),
-      inset 0 0 0 1px ${rgba(getColor('grey.600'), 0.7)},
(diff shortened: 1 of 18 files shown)


tinkertrain:
Hi @digitaldesigner, a couple of questions for ya!

## CheckmarkCard 
* I put the same depth as -> RadioCard, is that ok? should I roll back those?
* Just noticed that the focus style on a checked card is invisible as the border and the focus style are the same

## Other components
Should we apply depth to any of these?
* AttachmentList with preview
* MessageCard

digitaldesigner (changes requested):
Ah sorry man, CheckmarkCard should be `d400`

If you want to apply depth to AttachmentList with preview then it'd be `d400`, and MessageCard is `d600`

tinkertrain:
> Ah sorry man, CheckmarkCard should be `d400`
> 
> If you want to apply depth to AttachmentList with preview then it'd be `d400`, and MessageCard is `d600`

Thanks! I made the updates

digitaldesigner (approved):
Looks great, thanks @tinkertrain

plbabin (approved):
looks good 🚀

PR #892, opened 2020-11-18
Datepicker range

tinkertrain:
# Feature

In this PR we add the ability to select a range of dates in the Datepicker.

Storybook 2020-11-18 15-31-45

✅ Use the new `enableRangeSelection` prop (default false)
✅ Restrict the selection to a minimum number of days with `minBookingDays`
✅ After completing a selection range, starting a new one clears the existing one
✅ `onDateChange` returns both `startDate` and `endDate`
✅ Adds `type=button` to days to avoid triggering a form submit event in case you use the datepicker inside one.
✅ Adds a performance improvement with `useMemo`
✅ Adds an outer container div (to allow the use of styled components)
✅ Adds `innerRef` to obtain the DOM node

📹 https://www.loom.com/share/871fd4a5bd6747fc86fd8f8b66cc09d1

### DIFF
--- a/src/components/Datepicker/Datepicker.js
+++ b/src/components/Datepicker/Datepicker.js
@@ -10,6 +10,7 @@ import DailyCalendar from './Datepicker.DailyCalendar'
 
 function Datepicker({
   allowFutureDatePick = true,
+  enableRangeSelection = false,
   endDate = null,
   firstDayOfWeek = 1,
   minBookingDays = 1,
@@ -46,7 +47,7 @@ function Datepicker({
     onDatesChange: handleDateChange,
     numberOfMonths,
     minBookingDays,
-    exactMinBookingDays: minBookingDays === 1,
+    exactMinBookingDays: !enableRangeSelection,
     maxBookingDate: !allowFutureDatePick ? new Date() : undefined,
   })
 
@@ -76,7 +77,7 @@ function Datepicker({
     if (!data.focusedInput) {
       setDates({ ...data, focusedInput: START_DATE })
     } else {
-      if (minBookingDays > 1 && data.endDate) {
+      if (enableRangeSelection && data.endDate) {
         setDates({ ...data, endDate: null })
       } else {
         setDates(data)
@@ -99,7 +100,7 @@ function Datepicker({
       onDateHover,
       startDate: dates.startDate,
       endDate: dates.endDate,
-      minBookingDays,
+      enableRangeSelection,
     }),
     [
       isDateBlocked,
@@ -111,7 +112,7 @@ function Datepicker({
       onDateHover,
       dates.startDate,
       dates.endDate,
-      minBookingDays,
+      enableRangeSelection,
     ]
   )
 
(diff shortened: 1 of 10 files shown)


cen10:
@tinkertrain Not a blocker, but was curious if this would be a good PR to include the two changes we talked about when I was replacing the DatePicker in the Customers app - 
Inner ref: https://github.com/helpscout/hs-app/pull/9112/files#r515216437
Wrapper div: https://github.com/helpscout/hs-app/pull/9112#discussion_r512691125

tinkertrain:
> @tinkertrain Not a blocker, but was curious if this would be a good PR to include the two changes we talked about when I was replacing the DatePicker in the Customers app -
> Inner ref: [helpscout/hs-app#9112 (files)](https://github.com/helpscout/hs-app/pull/9112/files#r515216437)
> Wrapper div: [helpscout/hs-app#9112 (comment)](https://github.com/helpscout/hs-app/pull/9112#discussion_r512691125)

@cen10 Good call, thanks for reminding me 👍

digitaldesigner (changes requested):
Amazing work! A couple of very minor UI updates required, but functionally working super well and exactly as expected (thanks for the video).

The bulk of my feedback is covered in **[ this loom video](https://www.loom.com/share/871fd4a5bd6747fc86fd8f8b66cc09d1)** — but also listed below:

- If "today" is within the selection, style it the same as other days (i.e. blue background)
- On hover of a 1st column range, the left side of the cell should be rounded (currently square)
- If the 1st column is the end range, there should be no light-blue square edge
- There's a weird bug where if you hover the bottom corner of a middle range (before end date is selected), the cell will oscillate between square and round (should remain rounded, see gif below)
- The grid for the day names is slightly different widths and x-pos offset to the date cells, which causes a visual mis-alignment for the day names. See below. Not a blocker, but once you see it, it cannot be unseen

### Weird mouse interaction
ezgif com-gif-maker
### Cell alignment on days vs dates
Frame 3

digitaldesigner:
On the above comment about the cell sizes, changing the day name cells to `36px` wide magically fixes everything. I might be over-simplifying things, but hopefully it's not a major structural change

Screen Shot 2020-11-19 at 8 28 26 AM

tinkertrain:
@cen10 I have added the wrapper and the innerRef 👍 
@digitaldesigner Thanks for reviewing! I restructured some stuff and I think most if not all your points have been addressed... let me know!

digitaldesigner:
Thanks @tinkertrain — a few new bugs now:

- Hover states are missing for all cells, unless making a selection (should have a hover)
- On mouse down there's a focus border and the text moves a couple pixels (and shouldn't)
- If you select today as the start or end date the text is no longer visible
- Hovering over the left column mid-selection makes the right side of the cell flat and the left rounded (should be entirely rounded)

tinkertrain:
@digitaldesigner Thanks for taking the time to review!

1. Yikes I accidentally deleted those lines 🤦 
2...4 Fixed! 🤞

digitaldesigner (approved):
Looks awesome, thanks for your patience @tinkertrain

src/components/Datepicker/Datepicker.js
```diff
@@ -25,6 +28,14 @@ function Datepicker({
   const [prevDates, setPrevDates] = useState({})
   const [deepNavVisible, setDeepNavVisibility] = useState(false)
 
+  useEffect(() => {
+    if (datePickerRef && datePickerRef.current) {
+      innerRef(datePickerRef.current)
+    }
+    // No need to add innerRef prop to the deps list
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [datePickerRef])
```
plbabin, line 37: I think that `datePickerRef.current` would trigger the useEffect callback correctly here (not 100% sure though, maybe just `datePickerRef` is fine)

plbabin (approved):
thank you so much for taking care of that feature! The UX is spot on! nice work 🥇

PR #893, opened 2020-11-26
Upgrade green & purple color

plbabin:
# Feature
This is the last color change to be up-to-date with our color palette.

## Green

Overall, the green color update is straightforward. The only color change that would have been tricky is the `green.800` & `green.900` (they are a bit more darker). Fortunately, we don't use that color anywhere within HSDS.

One component that would need a design review is `Button`. Just to make sure the color change match our design vision. It looks almost the same as before, except the active color:

Screen Recording 2020-11-26 at 11 02 43 AM

## Purple
Purple color is migrating to `indigo`.

Three HSDS components are affected by this change:

### OptionTile
Image 2020-11-26 at 11 08 15 AM

### StatusDot
The custom status, which we believe is used within beacon:
Image 2020-11-26 at 11 09 28 AM

### Tag
Image 2020-11-26 at 11 09 49 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -314,6 +314,34 @@
         "@babel/types": "^7.10.1"
       }
     },
+    "@babel/helper-skip-transparent-expression-wrappers": {
+      "version": "7.12.1",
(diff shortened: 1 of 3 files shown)


digitaldesigner (approved):

tinkertrain (approved):
:shipit:

PR #894, opened 2020-12-07
Added onBodyClick prop to MessageCard component

Charca:
# Problem/Feature

This PR adds an `onBodyClick` prop to the `MessageCard` component. We need this in order to implement a fix for [a bug](https://helpscout.atlassian.net/browse/BEMBED-228) in Beacon, where links in a Message's body open inside the Beacon iframe.

## How to test

Pass a function as the `onBodyClick` prop to a `MessageCard`. The function should be called when you click on any element of the Message's body, but not when clicking on other parts of the Message (Title, Subtitle, Action CTA, etc).

### DIFF
--- a/src/components/MessageCard/MessageCard.jsx
+++ b/src/components/MessageCard/MessageCard.jsx
@@ -133,7 +133,7 @@ MessageCard.defaultProps = {
   in: true,
   isMobile: false,
   isWithBoxShadow: true,
-  onBodyClick: () => {},
+  onBodyClick: noop,
 }
 
 MessageCard.propTypes = {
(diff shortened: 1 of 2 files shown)


src/components/MessageCard/MessageCard.jsx
```diff
@@ -129,6 +133,7 @@ MessageCard.defaultProps = {
   in: true,
   isMobile: false,
   isWithBoxShadow: true,
+  onBodyClick: () => {},
```
plbabin, line 136:
```suggestion
  onBodyClick: noop,
```

I would replace the arrow function with the noop variable (it look better in the docs)

plbabin (approved):
looks good. I propose a small change for the default value, but it's not a big deal :shipit:

Charca:
Thanks @plbabin! Done!

PR #896, opened 2020-12-16
Removed MessageList component

Charca:
# Problem/Feature

This PR removes the now unused `MessageList` component. This is one of the components we had previously identified as not belonging in hsds-react, and since we have just removed its only usage from hs-app, it's a great opportunity to remove the code from this repo.

Bye-bye MessageList! 👋


plbabin (approved):
RgbPdsKuNkWbDCScZF

PR #897, opened 2020-12-18
Use effect only shadow for MessageCard component

Charca:
# Problem/Feature

We've recently found an issue on Beacon/Messages where the `MessageCard` component wasn't resizing correctly in some cases. After tracking down the bug, we found that it was coming from the HSDS [v3.9.3 release](https://github.com/helpscout/hsds-react/releases/tag/v3.9.3), where we added the depth component and refactored `MessageCard` to use the depth mixins.

In particular, the issue comes from the `will-change: box-shadow;` property that seems to be affecting the order in which property changes are calculated. Since we have a dynamic-height iframe that cares about when the `height` property changes, we were running into a conflict there.

My fix was to add a `d600Effect` depth constant that only applies the `box-shadow` styles.

# Question

I can't say I fully understand the `will-change` property, but I've noticed we're using it in a few places within HSDS, and since [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) suggests that we should only use this as a last resort, I was wondering if there was a particular reason why we sometimes use this optimization? Also, when we do need it, should we be applying it and removing it with JS as MDN suggests instead of doing it with CSS?

### DIFF
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -2,7 +2,7 @@ import styled from 'styled-components'
 import { makeFontFamily, setFontSize } from '../../styles/utilities/font'
 import { getColor } from '../../styles/utilities/color'
 import { FONT_FAMILY } from '../../styles/configs/constants'
-import { d600 } from '../../styles/mixins/depth.css'
+import { d600Effect } from '../../styles/mixins/depth.css'
 import Card from '../Card'
 import Button from '../Button'
 import Heading from '../Heading'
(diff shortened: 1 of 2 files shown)


Charca:
@plbabin assigning this one to you since JP is out of the office I believe.

plbabin (approved):
Looks good

as for `will-change` my comprehension is a bit like you. I think we started using it as a pre-optimization step when we know that a property will be animated through css (like box-shadow in the `Depth` component. 

We could do the update with JS, but I am not sure how we would do that through styled-component. also it's the first time we had issue with `will-change`. If it happen more often we could revisit that optimization step.

PR #900, opened 2021-01-06
Fix: EditableFields retain focus on enter/escape keypress events (Safari ONLY)

knicklabs:
# Problem/Feature

This update fixes an issue in Safari where enter/escape keypress events received on EditableField inputs result in the input retaining focus. Since we rely on the input losing focus in order to apply the mask and create a visual cue that a change was backed out or committed, this issue is impeding the usability of EditableFields for Safari users.

See this recording:

Demo of Problem

When enter is pressed the cursor jumps to the start of the input but the input does not blur and it is not hidden and replaced with the mask. You can try for yourself here: https://hsds-react.netlify.app/?path=/story/components-forms-editablefield--default-story - make sure that you use Safari!

# Solution

It's unfortunate to have to create a browser-specific fix however I believe that's the best course of action here since the problem is related to an unusual quirk in Safari that is present in the current version and the next version. There is also a history of focus issues with the browser; see https://github.com/facebook/react/issues/10871 and https://bugs.webkit.org/show_bug.cgi?id=179990.

To fix this issue, in the `handleInputFocus` handler we check if the related target is set on the event and if it is the field mask value because that is the related target when Safari sends the extra focus event. In this case, we then blur the input again otherwise we won't be able to click on it to reveal the input from behind the mask without first clicking outside the element. Normally the related target is null; however it may be set when using tab navigation so it is important to check for this specific target. Moreover, the fix is guarded by a browser check so it is only applied when the browser is Safari.

See this recording of the working solution 

Demo of Solution

You can try for yourself here: https://deploy-preview-900--hsds-react.netlify.app/?path=/story/components-forms-editablefield--default-story - make sure that you use Safari!

### DIFF
--- a/src/components/EditableField/EditableField.jsx
+++ b/src/components/EditableField/EditableField.jsx
@@ -181,17 +181,19 @@ export class EditableField extends React.Component {
     // See: https://github.com/facebook/react/issues/10871
     // See: https://bugs.webkit.org/show_bug.cgi?id=179990
 
-    // In the case where this happens, the second event will have a
-    // relatedTarget. The relatedTarget may also be set when tabbing
-    // from one input to another, so we need to specifically check that the
-    // relatedTarget is the field mask otherwise we will break tab navigation
-    // in Safari.
-    if (
-      navigator.userAgent.indexOf('Safari') !== -1 && // Apply only to Safari.
-      navigator.userAgent.indexOf('Chrome') === -1 && // Careful. "Safari" is also in Chrome's user agent.
+    // In the case where this happens, the second focus event will have a
+    // secondary target. If we have an EditableField, the secondary target
+    // will be the field mask. If we have an EditableFieldComposite, the
+    // secondary target will be the composed mask. Otherwise the secondary
+    // target will be null or another element in the case of tab navigation.
+    const isSafari =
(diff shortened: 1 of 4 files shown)


knicklabs:
There is a beta release for this: v3.10.4-0.

src/components/EditableField/EditableField.jsx
```diff
@@ -174,6 +174,30 @@ export class EditableField extends React.Component {
...
+    // In the case where this happens, the second focus event will have a
+    // secondary target. If we have an EditableField, the secondary target
+    // will be the field mask. If we have an EditableFieldComposite, the
+    // secondary target will be the composed mask. Otherwise the secondary
+    // target will be null or another element in the case of tab navigation.
+    const isSafari =
+      navigator.userAgent.indexOf('Safari') !== -1 &&
+      navigator.userAgent.indexOf('Chrome') === -1
+    const maskIsSecondaryTarget =
+      !!event.relatedTarget &&
```
tinkertrain, line 193: Oh, I've never seen `relatedTarget` before!

tinkertrain (approved):
That's an annoying 🐛, thanks for finding and fixing. Works 👌 

:shipit:

PR #901, opened 2021-01-11
Fix: isMonthInThePast returns false for past dates if the month is greater than this month

knicklabs:
[Jira Issue](https://helpscout.atlassian.net/browse/HSDS-180)

# Problem
There is a bug in the isMonthInThePast function the Datepicker utils that has surfaced and was caught be the 'isMonthInThePast' test in the Datepicker utils tests. The bug surfaced when we rolled over into the new year and since the test is failing we can not make new releases.

This blocked me in trying to release the next patch version 3.10.4.

The issue is with this logic:

```
return (
  someDate.getFullYear() <= today.getFullYear() &&
  someDate.getMonth() < today.getMonth()
)
```

The date used in the test was “May 29, 2020” and because we rolled over into 2021 this condition returned false when it should have returned true. May 29, 2020 is in the past but May is greater than January. We should only check if the month is less than this month if the year is this year.

 # Solution

I updated the test to use the following cases:

- "date" compared to "past date", where the month of the "date" is greater than the month of the "past date"
- "date" compared to "past date", where the month of the "date" is less than the month of the "past date". This condition is what caused tests to break when we rolled over into the new year.
- "date" compared to "same date"
- "date" compared to "future date"

I then updated the util method to get these tests to past. This involved checking if the month was less then this month if the year is the same on both dates being compared. Otherwise, we can simply compare the years.

### DIFF
--- a/src/components/Datepicker/Datepicker.utils.js
+++ b/src/components/Datepicker/Datepicker.utils.js
@@ -47,11 +47,14 @@ export function isToday(someDate) {
  */
 export function isMonthInThePast(someDate) {
   const today = new Date()
+  const thisYear = today.getFullYear()
+  const someYear = someDate.getFullYear()
 
-  return (
-    someDate.getFullYear() <= today.getFullYear() &&
-    someDate.getMonth() < today.getMonth()
-  )
+  if (someYear === thisYear) {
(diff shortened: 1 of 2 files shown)


tinkertrain (approved):
Thanks! 💙

PR #906, opened 2021-02-19
DropList (new dropdown component)

tinkertrain:
# Old Dropdown, meet DropList

Excited

This a completely new component to replace our current Dropdown, a difficult to tame beast 🦖.

DropList was designed to cover our current needs and nothing more, while remaining flexible if such needs change in the future.

DropList uses two open source libraries under the hood to do the heavy lifting: **downshift** for dropdowny stuff and **tippy** for positioning and opening and closing state, including _where_ to render. DropList tries to abstract as much as possible so you as a user don't have think too much about the internals, but if you need to you should be able to read those libraries docs and tweak to your heart content!

DropList has a batteries included mindset, and so it includes a collection of what we'll call "togglers" to cover all current HSApp needs, from a regular ol' button, a select style design, split button (that will replace our current SplitButton component) and more! DropList handles the position and direction of them for you, one thing less to think about.

DropList is extensively documented and "storied"! Check it [here](https://deploy-preview-906--hsds-react.netlify.app/iframe.html?id=components-dropdowns-droplist--default-story&viewMode=docs#droplist)

And finally, DropList's interactions and behaviours are well tested:

Screen Shot 2021-03-02 at 5 04 38 PM
 
## ✋ Note
Remember those nested dropdowns? They are a thing of the past, so they were dropped from DropList (bad pun intended): they are bad UX and inaccessible, if you encounter them you have a license to push back and ask to use a different pattern 💪.


digitaldesigner:
Mate stunning work as always! Below is some broad feedback that applies to all versions of the dropdown. Also see attached image for # reference:

## Visual QA
1. The spacing top and bottom has changed in Figma since the original designs (I'm so sorry). So it's basically the same as the side spacing
2. Border color should be `grey.600` but appears a little darker in code
3. There should be `2px` space between rows
4. For searchable dropdowns, could we go a little wider? I.e. `220px` instead of `200px`?
5. The searchable input height should be `2px` smaller in height
6. Would it be possible to contradict the spacing from point number 1 if the first-child is a heading? In the attached image what you've coded _looks_ better because of the additional space. No dramas if not

01-QA_Searchable
02-QA_Groups

## Functional QA
7. The cursor is often not the hand (and should be). See video from number 8
8. [Multiple Selection](https://deploy-preview-906--hsds-react.netlify.app/?path=/story/components-dropdowns-droplist--multiple-selection) is a bit borked, see video: https://c.hlp.sc/P8uk1ZYK
9. There looks to be some visual QA necessary on the multi-select checkmark icon (should be a little smaller and fatter)

04-QA_Multi-Select

digitaldesigner (changes requested):
See above comment

tinkertrain:
@digitaldesigner Thanks for looking at it and the feedback!

1. ✅
2. I changed it to `grey.600` (from `charcoal.600`) and it was still quite dark, I found there's a mistake in the colorway template https://github.com/helpscout/colorway/blob/f816d80946c0a92f1d6e7c74dfaac27ecf9bf20f/manifests/hsds-react#L42, the 600 value is the same as the 800... can you confirm that the value should be `#C5CED6` for `grey.600` so I can change it there?
3. ✅
4. ✅
5. ✅
6. ✅
7. ✅
8. I couldn't reproduce this on my mac (firefox, safari or chrome) and I clicked like a madman, weird! Does it happen all the time with you? What browser?
9. Does the "check" icon svg needs to be updated to get it "fatter"? (I reduced the size)

ItsJonQ:
Haiiiiiiiiii ❤️ !!!

I hope you folks are doing well 🤗 .
I just wanted to leave some notes as I had to work with `downshift` for custom dropdown stuff not too long ago.

@tinkertrain A slight challenge we had was with downshift + Popper.js + **Portal**.
(Downshift was quite unhappy with Portal with this stack)

Based on the Storybook, it looks like Popper.js is something y'all are using for positioning (via Tippy).

We eventually got things working though.

As you know, Dropdowns are beasts :lolcry:.

I have the source code for my solutions here, in case it comes in handy:
https://github.com/ItsJonQ/g2/tree/main/packages/components/src/select-dropdown

Good luck + have fun friends! 🎉

digitaldesigner (changes requested):
Awesome work man, all looking good on the Default and Groups. Multiple selection is still causing some problems for me, both in Chrome (v88) and Safari (v14) on Mac (Big Sur, 11.2.1) — missing hover and active states. If it's any help, [this prototype](https://cdpn.io/buzzusborne/debug/4150653a2cc483e6dfa439375a942078) demonstrates the intended functionality for hover (i.e. blue check) and active (white check, blue circle). Described in this video: https://www.loom.com/share/86b30468c12b4853b64b054f7bd167a2

Also yes, `grey.600` **should** be `#C5CED6`.

tinkertrain:
@ItsJonQ Helloo! Hoping you are also doing great, long time no see! Thanks for the tips 🥇 

I hadn't really looked at the portaling yet, and I just added support for it. I used [tippy's appendTo](https://atomiks.github.io/tippyjs/v6/all-props/#appendto) to handle it, all I had to do to make it work properly is add the scope's classname (hsds-react by default) to the instance to get the styles working. 

Then you can do something like:

```jsx
<DropList
  items={regularItems}
  toggler={<Button text="Click" />}
  tippyOptions={{
    appendTo: () => document.body,
  }}
/>
```

Changes here: [`d3e353d` (#906)](https://github.com/helpscout/hsds-react/pull/906/commits/d3e353de5218453125ead5969e9fcd143b7d3a71)

I need to do more testing, but I think it works?....

tinkertrain:
@digitaldesigner Thanks for the detailed feedback and 🎥 !

Multiple selection 🐞 ... with PL's help I managed to reproduce it in my machine, when running it locally it all was working but the netlify hosted version wasn't and it didn't occurred to me to try it there (almost never do, but from now I will 😆).

Anyways to not make the history long I managed to find the cause (it had to do with how downshift gets compiled for production... talk about a random cause 😅) and managed to fix it so now you should be able to experience the thing as intended (this includes those hover states too).

Some other things regarding your other notes:
- `tab`: According to the [wai aria guidelines](https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-6) for select and combobox, the only way to move through the list is using up and down arrows (and `home` and `end`, but I guess those are less common). These guidelines are what the library I'm using here (downshift) follows.
- Which means `tab` will move the focus to the next _focusable_ element on the page (I made an adjustment to close the list when this happens)
- Related to that, the blue border on the list items: at the moment the highlight (light blue) happens both on hover and when moving with arrows as they are essentially the same action (highlight an item for selection), not sure we need 2 different visual effects for this.
- Fixed the input being auto-filled bug you mentioned.


Thanks!

digitaldesigner (changes requested):
Amazing work, thanks @tinkertrain! Wild about the difference in environments, but glad you found the cause!! Yeah the multiselect is looking really great, only a few tiny requests below:

1. Could the hover of an un-selected multiselect item include a white background to the checkmark? See image
2. New checkmark icon, see [checkmark.zip](https://github.com/helpscout/hsds-react/files/6066839/checkmark.zip)
3. Totally fine with the tab improvements, all works as I would expect. As for the following item, I couldn't repro the blue outline, but you're right it's not necessary so probably worth removing as the blue background should be enough
> The blue border on the list items: at the moment the highlight (light blue) happens both on hover and when moving with arrows as they are essentially the same action (highlight an item for selection), not sure we need 2 different visual effects for this.

cen10:
@tinkertrain Amazing work!! I haven't had time to look at the code in detail yet, but I did play with everything in Storybook. I wanted to run these by you in case they're unwanted behavior. Maybe it's just something having to do with the storybook environment.

When I first load [Storybook](https://deploy-preview-906--hsds-react.netlify.app/iframe.html?id=components-dropdowns-droplist--default-story&viewMode=docs#droplist), the very first example has a blue focus outline around the carat/button part. This focus outline goes away if I alt+tab away from the browser and come back, or if I click somewhere else on the page. So it seems to be starting out with focus on it already?

Screen Shot 2021-03-03 at 2 43 14 PM

When I click on the dropdown, I see the blue focus outline, but I thought that was only supposed to show up on tab? 
https://c.hlp.sc/2NuE8x16

tinkertrain:
@cen10 Thanks for looking at it!

The reason it was focused on load is that I added the `initialIsOpen` prop on the default story, which happens to be a combobox with an input the get focused when you open the dropdown... in a nutshell is not a good thing to do on storybook because by "stealing" the focus it messes up the docs, I've removed the prop.

The buttons that make of both of the togglers `Button` and `SplitButton` are HSDS buttons, so the focus, hover, active, etc behaviours are the same as those, I didn't really tweaked them.

digitaldesigner (approved):
Nailed it

src/components/DropList/DropList.jsx
```diff
@@ -0,0 +1,230 @@
...
+    if (toggler.type === SelectTag) {
+      const { text } = toggler.props
+
+      if (text == null) {
+        togglerProps.text = itemToString(selectedItem)
+      }
+    }
+
+    const { placement, offset } = getTogglerPlacementProps(toggler)
+    tippyProps.placement = placement
```
plbabin, line 105: In case we create a custom toggler, it might be nice to let that component pass down a placement/offset. I guess we could skip that step if tippyProps/tippyOptions already contain a placement/offset

any thoughts on that?

src/components/DropList/DropList.css.js
```diff
@@ -0,0 +1,180 @@
...
+
+export const DropListWrapperUI = styled('div')`
+  box-sizing: border-box;
+  width: ${({ variant }) => (variant === 'combobox' ? '220px' : '200px')};
+  padding: 0;
+  background-color: white;
+  border: 1px solid ${getColor('grey.600')};
+  border-radius: 4px;
+  box-shadow: 0px 1px 7px rgba(0, 0, 0, 0.08);
+  font-family: ${AKTIV_FONT_FAMILY};
```
plbabin, line 15: instead of forcing Aktiv here, maybe it should it be `var(--HSDSGlobalFontFamily)` . Just in case the font is overwritten on the global stylesheet. 

same with the font-size : `var(--HSDSGlobalFontSize)`.

src/components/DropList/DropList.css.js
```diff
@@ -0,0 +1,180 @@
...
+export const InputSearchHolderUI = styled('div')`
+  width: calc(100% - 8px);
+  margin: 4px 4px 5px 4px;
+  display: ${({ show }) => (show ? 'block' : 'none')};
+
+  input {
+    width: 100%;
+    height: 38px;
+    padding: 0 15px;
+    font-family: ${AKTIV_FONT_FAMILY};
```
plbabin, line 50: same as previous comment about using css variable `HSDSGlobalFontFamily`

src/components/DropList/DropList.togglers.jsx
```diff
@@ -0,0 +1,396 @@
...
+
+const NavLinkTogglerUI = styled('button')`
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  height: 55px;
+  padding: 0 18px;
+  background: transparent;
+  border: 0;
+  font-family: ${AKTIV_FONT_FAMILY};
```
plbabin, line 92: another spot where we could use the variable `--HSDSGlobalFontFamily`

src/components/DropList/DropList.togglers.jsx
```diff
@@ -0,0 +1,396 @@
...
+  width: 150px;
+  height: 38px;
+  padding: 0 15px;
+  margin: 0;
+  background: white;
+  border: 0;
+  box-shadow: inset 0 0 0 1px ${getColor('grey.800')};
+  box-sizing: border-box;
+  border-radius: 3px;
+  font-family: ${AKTIV_FONT_FAMILY};
```
plbabin, line 222: +1 for `--HSDSGlobalFontFamily`

plbabin (approved):
with all latest changes, I really like this new component. Thank you!!

I left a couple of comments on using the css variable for font instead of the Aktiv constant. However this is not a blocker from me. :shipit:

src/components/DropList/DropList.css.js
tinkertrain: Always forget... thanks!

PR #908, opened 2021-03-23
HSDS-178: Update RadioCard component

jakubjanczyk:
[JIRA issue](https://helpscout.atlassian.net/browse/HSDS-178)

# Feature

Purpose of this PR is to update RadioCard component to the newest design. The use of it can be found here: https://www.figma.com/file/JwgxQmv2KyQTKwoiuyxHGq/Spec%2F-Messages-1.5

The demo implementation: https://codepen.io/team/helpscout/pen/5d458fb128e71e612ef1ad7202b178cd

The main changes are in "checked" state, the whole card gets larger. Additional there's some opacity added for "non-checked" cards.

# Solution

The implementation isn't precisely as in codepen, as there are used images and fixed sizes to achieve the result. However it looks and behaves very similarly. The only thing not implemented as part of this story is change of the font color - that's because it collides with the existing usage of this component.

The solution is to use `scale` transform function to make the card larger. The card itself doesn't have a fixed size, so there's no easy way to know how many pixels the component should be larger with. Also, using `scale` makes the implementation much simpler, without the need to use absolute position and wrap into some additional HTML elements.

As part of this PR I've also added additional available Icon size, as it would be needed in the place where it would be used (Messages builder). Additionally, IconWrapper component has possiblity to adjust better to the size of Icon. This change doesn't break anything, but adds another possibility to the usage. 

There's new story added to the storybook of this component that shows an example of usage with hsds-illos.

https://c.hlp.sc/YEuRgQmD

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1964,9 +1964,9 @@
       }
     },
     "@helpscout/hsds-illos": {
-      "version": "1.8.0",
-      "resolved": "https://registry.npmjs.org/@helpscout/hsds-illos/-/hsds-illos-1.8.0.tgz",
-      "integrity": "sha512-4Q++pUlCBG45COw2RmeQO5vNRV/7k13pUHL0UPTD5BQCmpobKZLGG/xKoU4HeayfQL3x92hCjul8QgLRqrLNCA=="
+      "version": "1.9.0",
+      "resolved": "https://registry.npmjs.org/@helpscout/hsds-illos/-/hsds-illos-1.9.0.tgz",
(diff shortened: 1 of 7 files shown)


plbabin:
@jakubjanczyk one thing I noticed is that the `RadioCard` active state overwrite the selected one so when the mouse is down on the card, it will animate even though it's already selected. I think it would look better without the animation in that case.

Screen Recording 2021-03-24 at 09 30 18 AM

jakubjanczyk:
@plbabin I haven't noticed that case, thanks!

tinkertrain (approved):
Thanks @jakubjanczyk looks good to me! 💯

jakubjanczyk:
@tinkertrain thanks! Is there someone responsible for releasing as in other projects or should I do it?

PR #909, opened 2021-03-23
Add withTriggerWrapper prop on Tooltip

plbabin:
# Problem/Feature

Screen Recording 2021-03-23 at 02 14 30 PM

This prop will allow us to remove the wrapping span around the tooltip trigger. That span exists to ensure a valid trigger for Tippy. It needs a forwardable component to attach any events on the trigger. In some scenario, that span become an issue:

- Focusable button, we need the button to have the focus state, not the wrapping span
- Group of components were we style based on the children position (first child, last child, etc) 

To make that new prop work, the children needs to be a unique and valid child:

```javascript
<Tooltip title="Hello" withTriggerWrapper={false}>
  <button>No wrapping span Trigger</button>
</Tooltip>
```

If we pass a non-valid react element, it will be wrap with the span even though the flag is turned off

```javascript
<Tooltip title="Hello" withTriggerWrapper={false}>
  No wrapping span Trigger
</Tooltip>
```

We did decide to have the wrapping `span` on by default, to stay compatible with the way the tooltip was implemented in hs-app

### DIFF
--- a/src/components/Tooltip/Tooltip.jsx
+++ b/src/components/Tooltip/Tooltip.jsx
@@ -57,6 +57,7 @@ const Tooltip = props => {
     renderContent,
     title,
     triggerOn,
+    withTriggerWrapper,
     zIndex: zIndexProp,
     ...rest
   } = props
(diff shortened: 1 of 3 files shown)


tinkertrain (approved):
:shipit:

PR #910, opened 2021-03-30
Allow to use data-bypass for Nav.Item

jakubjanczyk:
# Problem

This PR adds possibility to provide data-bypass attribute for a Nav.Item component and it would be set on actual link html element.
Context for the change is this story: https://helpscout.atlassian.net/browse/HSAPP-3810. In this case, we are blocking route change on React router, however there's Backbone router on top of this all that still changes the route. `data-bypass` is a special attribute to tell Backbone app to not use its router. 

Similar functionality is already present for a `Button` component

### DIFF
--- a/src/components/Nav/Nav.Item.jsx
+++ b/src/components/Nav/Nav.Item.jsx
@@ -35,6 +35,7 @@ export class NavItem extends React.Component {
       location,
       strict,
       to,
+      'data-bypass': this.props['data-bypass'],
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/Nav/Nav.Item.jsx
```diff
@@ -35,6 +35,7 @@ export class NavItem extends React.Component {
       location,
       strict,
       to,
+      'data-bypass': this.props['data-bypass'],
```
tinkertrain, line 38: If you want to follow the destructuring above, you can rename it in place:

```
const { disabled, exact, isActive, location, strict, to, 'data-bypass': dataBypass } = this.props


//...
to,
'data-bypass': dataBypass,
}
```

This is how we do it with `data-cy`

tinkertrain (approved):
Just left a little comment, not very important!

🚢

PR #911, opened 2021-04-01
Add support for image below Message body

jakubjanczyk:
# Feature

Related JIRA ticket: https://helpscout.atlassian.net/browse/BEMBED-235
Related figma: https://www.figma.com/file/JwgxQmv2KyQTKwoiuyxHGq/Spec%2F-Messages-1.5?node-id=0%3A1

We need to be able to display Image on the Message Card, below the body. There could be only one image per message at the moment. Image would be resized to the original size that it was uploaded with, and centered when width of image is smaller than width of message card.

### DIFF
--- a/src/components/MessageCard/MessageCard.jsx
+++ b/src/components/MessageCard/MessageCard.jsx
@@ -11,6 +11,8 @@ import {
   SubtitleUI,
   BodyUI,
   ActionUI,
+  ImageUI,
+  ImageContainerUI,
 } from './MessageCard.css'
 import Truncate from '../Truncate'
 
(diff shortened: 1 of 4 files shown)


src/components/MessageCard/MessageCard.jsx
```diff
@@ -81,6 +83,21 @@ export class MessageCard extends React.PureComponent {
...
   }
 
+  renderImage() {
+    const { image } = this.props
+
+    return image ? (
+      <ImageContainerUI>
+        <ImageUI
+          src={image.url}
+          alt={'Message image'}
```
tinkertrain, line 93: For accessibility reasons, maybe we can make allow for this to be passed, as `image.altText` or something like that and make the default 'Message Image' as you have it now.

src/components/MessageCard/MessageCard.jsx
```diff
@@ -164,6 +182,12 @@ MessageCard.propTypes = {
   subtitle: PropTypes.string,
   /** Title of the Message. */
   title: PropTypes.string,
+  /** Definition of the Message image */
+  image: PropTypes.shape({
+    url: PropTypes.string,
```
tinkertrain, line 187: Perhaps the `url` should be required?

tinkertrain (approved):
I left a couple of comments, just little things. Good work! 💯

Charca (approved):
LGTM 👍

PR #912, opened 2021-04-12
Enhance conditions components for a new Messages builder

jakubjanczyk:
# Feature

This PR updates Multiple components around Conditions functionality to adjust to the newest design of a Message builder.

Here's the figma file: https://www.figma.com/file/wTmMX4SE6Jy1kTyU55K6UG/WIP---Spec%2F-Properties-in-Messages?node-id=139%3A0
JIRA ticket: https://helpscout.atlassian.net/browse/HSAPP-3292

The main changes are:
- Possibility to switch between AND/OR conjunction - this is optional and can be enabled or disabled. New DropList component is used here for the implementation. 
- Possibility to change conjunction even after it's not possible to add next conditions any more (because of reaching a limit, for example.
- Possibility to mark Condition, so there's no dropdown to select type - optional

There are no breaking changes - every new feature is optional.

This PR includes updated Story with example of usage.

Screenshot from 2021-04-12 19-51-21
Screenshot from 2021-04-12 19-51-04


src/components/ConditionField/ConditionField.test.js
```diff
@@ -109,3 +107,128 @@ describe('Tooltip', () => {
...
+  test('should display AND conjunction selected', () => {
+    render(
+      <ConditionField.Group canChangeConjunction conjunction={'and'}>
+        <ConditionField />
+      </ConditionField.Group>
+    )
+
+    user.click(dropdownTrigger())
+    expect(screen.getByRole('button', { name: 'and' })).toBeInTheDocument()
+    // expect(getSelectedConjunctionOption()).toHaveTextContent('AND')
```
Charca, line 152: Can these commented out expect calls be removed?

Charca (commented):
Code looks good to me, great job @jakubjanczyk!

Just two things I've noticed:

1. The Figma file shows that the chevron icon should point up when the dropdown is open: https://c.hlp.sc/ApuYLdE9
2. There's a weird behavior with the conjunction selector that I've noticed testing this in Storybook. Everything works great, but if you change to a different story, and then come back to the Condition story and open the dropdown, it looks like there's a z-index issue that prevents you to select an option. Here's a screencast showing this behavior: https://c.hlp.sc/6qu8PnQ8. I'm not sure if this is a Storybook thing, a DropList thing, or something about the Condition component. Any clues?

src/components/ConditionField/ConditionField.test.js
jakubjanczyk: Yeah, I wanted to test this, but the component doesn't support it yet. I'll remove it.

jakubjanczyk:
@Charca Thanks for the observations!
1. :heavy_check_mark: 
2. This one is tricky, but it mostly seems to be just StoryBook issue with iframes. We are trying other options to not have this problem, but none is working perfectly. Worst case, we would, need to keep it as it is - when using on hs-app it seems to be working fine.

tinkertrain (approved):
Good work 💯  and as discussed earlier 🚢 !

PR #914, opened 2021-04-14
DropList: Adds menuCSS prop to allow custom css on the menu when portaling

tinkertrain:
# Feature

Sometimes when portaling you might need to apply some custom css to the menu but find that adding styles using Styled Components doesn't work (this happens because the styled component can't target the menu as it's no longer in its scope). In these rare ocasions you can use the `menuCSS` prop to add your styles, it accepts the returned value from the [`css` helper from Styled Components](https://styled-components.com/docs/api#css)

```jsx
// import { css } from 'styled-components'

<DropList
  items={regularItems}
  toggler={<SimpleButton text="Click" />}
  tippyOptions={{
    appendTo: () => document.body,
  }}
  menuCSS={css`
    width: 134px;

    &:hover {
      color: salmon;
    }
  `}
/>
```

> Note: `menuCSS` should only be used in conjuction with `tippyOptions.appendTo`, if not you will see errors in the console. Use regular styled components in any other instance.

## Other

- Fixes a bug that was happening in the condition story

### DIFF
--- a/src/components/ConditionField/ConditionField.AddButton.jsx
+++ b/src/components/ConditionField/ConditionField.AddButton.jsx
@@ -75,6 +75,7 @@ const AnimatedComponent = withMotion({
     const height = node.clientHeight
     node.style.height = 0
     node.style.overflow = 'hidden'
+    node.style.willChange = ''
 
     return animate({
       keyframes: [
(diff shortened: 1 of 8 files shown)


jakubjanczyk (approved):
Looks great! Thank you for those changes @tinkertrain

PR #916, opened 2021-04-20
Prevent MessageCard image from resizing

jakubjanczyk:
# Problem

The size of the MessageCard image is now set to be the size of original image. This approach has an issue that when the original image is really big, it would cause the Message UI to jump after loading image to adjust to the actually displayed size..

Because of that, it's necessary to recalculate the dimensions of an image so that it fits the given boundaries. Boundaries has been defined as a 278x278 square (based on the available width within a MessageCard). The image would keep aspect ration, independent if it wider or taller, adjusting to the longer side. The image container would be visible before image renders, as a transparent space, preventing the card to resize afterwards.

How it looked before: https://c.hlp.sc/YEuRPR4B (there's a short, but visible moment when the whole card gets bigger when image is loaded)
How it looks now: https://c.hlp.sc/YEuRPRYj (notice that the card has the same size for the whole time)

Related JIRA ticket: https://helpscout.atlassian.net/browse/BEMBED-266

### DIFF
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -7,6 +7,9 @@ import Card from '../Card'
 import Button from '../Button'
 import Heading from '../Heading'
 import Image from '../Image'
+
+export const MAX_IMAGE_SIZE = 278
+
 const fontFamily = makeFontFamily('Barlow')
 
 export const MessageCardUI = styled(Card)`
@@ -272,7 +275,7 @@ export const ImageUI = styled(Image)`
   border-radius: 3px;
   height: ${({ height }) => height};
   width: ${({ width }) => width};
-  max-height: 278px;
+  max-height: ${MAX_IMAGE_SIZE}px;
 `
 
 export const ImageContainerUI = styled('div')`
@@ -281,5 +284,5 @@ export const ImageContainerUI = styled('div')`
   width: 100%;
   margin-top: 20px;
   padding: 0 10px;
-  max-height: 278px;
+  max-height: ${MAX_IMAGE_SIZE}px;
 `
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -271,6 +271,9 @@ export const ActionButtonUI = styled(Button)`
 
 export const ImageUI = styled(Image)`
   border-radius: 3px;
+  height: ${({ height }) => height || 'auto'};
+  width: ${({ width }) => width || '100%'};
+  max-height: 278px;
 `
 
 export const ImageContainerUI = styled('div')`
@@ -279,4 +282,5 @@ export const ImageContainerUI = styled('div')`
   width: 100%;
   margin-top: 20px;
   padding: 0 10px;
+  max-height: 278px;
 `
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -271,8 +271,8 @@ export const ActionButtonUI = styled(Button)`
 
 export const ImageUI = styled(Image)`
   border-radius: 3px;
-  height: ${({ height }) => height || 'auto'};
-  width: ${({ width }) => width || '100%'};
+  height: ${({ height }) => `${height}px` || 'auto'};
+  width: ${({ width }) => `${width}px` || '100%'};
   max-height: 278px;
 `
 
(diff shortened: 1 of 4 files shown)


src/components/MessageCard/MessageCard.jsx
```diff
@@ -86,16 +99,45 @@ export class MessageCard extends React.PureComponent {
...
+  // Calculate size of image to keep the original aspect ratio, but fit within 278x278 square for image
+  calculateSize = image => {
+    if (!image.width || !image.height) {
+      return {}
+    }
+    const width = parseInt(image.width)
+    const height = parseInt(image.height)
+
+    // Not necessary to recalculate if it fits within boundaries
+    if (width < MAX_IMAGE_SIZE && height < MAX_IMAGE_SIZE) {
```
plbabin, line 129: I do wonder if `MAX_IMAGE_SIZE` should be moved as a prop so we can change it per usage. Default could be 278

cc @tinkertrain

src/components/MessageCard/MessageCard.jsx
```diff
@@ -86,16 +99,45 @@ export class MessageCard extends React.PureComponent {
...
+
+    return (
       <ImageContainerUI>
         <ImageUI
           src={image.url}
           alt={image.altText || 'Message image'}
-          width={image.width || '100%'}
-          height={image.height || 'auto'}
+          width={width || '100%'}
+          height={height || 'auto'}
```
Charca, line 114: Should we _not_ pass the default values `100%` and `auto` here, since we're considering the null scenario in the styled component? Otherwise, I think we could end up with a value like `100%px`.

Charca (commented):
Nice job @jakubjanczyk, I just left a couple of comments for ya.

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: It is a good idea @plbabin, but the width of message itself is not customizable, so bigger size might not make sense in this case. Probably only to make it smaller. Do you think maybe make it customizable, but limit to 278 maximum either way?

src/components/MessageCard/MessageCard.css.js
```diff
@@ -271,6 +270,9 @@ export const ActionButtonUI = styled(Button)`
 
 export const ImageUI = styled(Image)`
   border-radius: 3px;
+  height: ${({ height }) => `${height}px` || 'auto'};
+  width: ${({ width }) => `${width}px` || '100%'};
```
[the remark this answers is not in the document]
jakubjanczyk: Yes :facepalm: I've been changing it as a fix and forgot about this. Thanks for spotting it!

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: yeah - same as above :facepalm:

src/components/MessageCard/MessageCard.jsx
tinkertrain: IMO if the need to make it customizable has not been needed until now, I'd leave it as is until it is. It's a simple enough change though, so if any of you know that the use case will be indeed needed feel free to add it!

src/components/MessageCard/MessageCard.css.js
```diff
@@ -271,6 +270,9 @@ export const ActionButtonUI = styled(Button)`
 
 export const ImageUI = styled(Image)`
   border-radius: 3px;
+  height: ${({ height }) => height};
+  width: ${({ width }) => width};
+  max-height: 278px;
```
tinkertrain, line 275: Perhaps you can import the constant from `src/components/MessageCard/MessageCard.jsx:19` and use it here and in line 284, to mitigate the feeling of a "magic number" and if the need to change it arises you only do it on one place.

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: I think we don't have any other cases at the moment that would require us to make it customizable. @Charca maybe you know about any?

src/components/MessageCard/MessageCard.css.js
jakubjanczyk: Good catch! thanks :)

src/components/MessageCard/MessageCard.jsx
plbabin: if it's a fixed width then it's all good with me. I wasn't sure if this component has to be fluid or not. We can always revisit if we need it to have different width

Charca (approved):

PR #917, opened 2021-04-21
Refactors EmojiPicker to use DropList instead of Dropdown

tinkertrain:
Step 1 for removing Dropdown is to remove it from HSDS internal usages.

This PR refactors EmojiPicker to use DropList under the hood instead of Dropdown.

Additionally takes `EmojiPicker.Emoji` to a standalone component `<Emoji />`, this is then used by the new `EmojiPicker` and `site/js/apps/mailbox/chat/ChatApp/src/components/EmojiCommand/EmojiListItem.js` in HS App.


TerrenceLJones (approved):
This looks great! 🚢

plbabin (approved):
looks really good! 🎉 🎉🎉

PR #919, opened 2021-04-28
[HSDS-185] Datepicker component is borked in Storybook

tinkertrain:
# Problem

https://helpscout.atlassian.net/browse/HSDS-185

The problem appears to be happening only on Storybook and I can't figure out why, I did have a similar issue when working on DropList. They both use an external library but not the same so not sure what the deal is with that.

The issue is with styled components, when styling a component using props like this the usage of props to set some CSS prop is not being applied correctly:

```
<DayUI bgColor={isSelected ? 'blue' : 'green'} />

DayUI = styled('button')`
   background-color: $(props.bgColor => props.bgColor);
`
```

Using class names instead works fine:

```
<DayUI className={isSelected ? 'is-selected' : ''} />

DayUI = styled('button')`
  background-color: green;

  &.is-selected {
    background-color: blue;
  }
```

This PR refactors the DatePicker css to rely on classnames instead of props since it works everywhere (it's the same solution I used in DropList eventually).

I still don't know why the combination of Storybook, styled components and the prop pattern is not working correctly lately.

### Other

Upgraded storybook to latest, adjusted configuration and removed the webpack external file and instead put it in the storybook configuration, like is recommended in the docs.


jaredmcdaniel (changes requested):
Looking good @tinkertrain. Only thing I'm seeing now is the `selected` state for the [**default** variation](https://deploy-preview-919--hsds-react.netlify.app/?path=/story/components-dropdowns-datepicker--default-story). Currently it shows as light blue (hover style) but I believe it should be blue-500 with white text [per the docs](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=7453%3A17827).

image

plbabin (approved):
looks good, in a way it's easier to understand the colors now, and we can use styled-component directly to overwrite some colors if needed!

tinkertrain:
@jaredmcdaniel Thanks for looking at it! I have pushed an update

jaredmcdaniel (approved):
Looks great! 🚢

PR #920, opened 2021-04-29
Improve accessibility of Attachment remove button

jakubjanczyk:
# Problem/Feature

Currently it's not possible to navigate to remove button of an Attachment preview with keyboard. This PR adds a functionality to show remove button not only on hover, but also on focus. Additionally there's a change on how hiding of a button is done. Previously it was by setting `display:none;`, but this breaks the accessibility. Instead, the button is now clipped to no size and the clip is removed on hover/focus. 

The best way to test this change is to check `Attachment -> Preview theme` story and try tabbing to the attachment and a remove button.

See this recording: https://c.hlp.sc/ApuYg8ez

### DIFF
--- a/src/components/Attachment/Attachment.css.js
+++ b/src/components/Attachment/Attachment.css.js
@@ -110,7 +110,8 @@ export const AttachmentUI = styled.a`
 
   ${bem.element('closeButton')} {
     ${cardStyles()};
-    display: none;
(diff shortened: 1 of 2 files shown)


tinkertrain (approved):
Nice! I haven't thought of using `clip` to hide elements 🚢

PR #921, opened 2021-05-04
[Refactor] Replaces usage of Dropdown in EditableField with DropList

tinkertrain:
In this PR we refactor EditableField (the options variant) to use DropList instead of Dropdown.

This does not change external behaviours so there's no need to update wherever is used (this variant of EditableField is only used on one place in HSApp by the way).

[Story](https://deploy-preview-921--hsds-react.netlify.app/?path=/story/components-forms-editablefield--with-options)

I also fixed a couple of EditableFields minor issues that I noticed:
- Propagate the disabled option all the way
- On multiple value fields with options, adding a new one and opening the options dropdown with the input empty would remove the newly added field.

While doing the refactor I noticed DropList has `onMenuBlur` but not focus, so I added `onMenuFocus` which I ended up not needing for the refactor but kept it for consistency.

## ESC handling

https://helpscout.atlassian.net/browse/HSDS-181

When an editable field resided inside a modal, activating the field and then hitting ESC would deactivate the field and also close the modal. In that particular case, we don't want to close the modal, only deactivate the field.

But this would happen also with: a DropList being opened and hitting ESC (which will normally just close the DropList), editable textarea...

I tried to stop the propagation of the event so that it wouldn't reach the modal listener but due to the way it was handled with `keyup` instead of `keydown`, it wouldn't catch it. And when I changed it to `keydown` the modal event happens first, so again no way to stop the bubbling.

I ended up adding an option to PortalWrapper (the handler of this event on the modal) to avoid closing the modal if the target originating the event is coming from an opened component.

I added a story on the Modal to test this

### DIFF
--- a/src/components/EditableField/EditableField.jsx
+++ b/src/components/EditableField/EditableField.jsx
@@ -211,17 +211,9 @@ export class EditableField extends React.Component {
 
   handleInputBlur = payload => {
     const { name, event } = payload
-    const { activeField, multipleValuesEnabled } = this.state
+    const { activeField, multipleValuesEnabled, valueOptions } = this.state
     const { validate, onCommit, onDiscard, onInputBlur } = this.props
-
-    if (equal(this.state.initialFieldValue, this.state.fieldValue)) {
-      this.setState({ activeField: EMPTY_VALUE }, () => {
-        onInputBlur({ name, value: this.state.fieldValue, event })
-      })
-
-      return
-    }
-
+    const hasOptions = valueOptions != null
     const changedField =
       this.state.fieldValue.length === 1
         ? this.state.fieldValue[0]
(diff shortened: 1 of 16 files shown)


src/components/EditableField/EditableField.jsx
```diff
@@ -956,7 +958,9 @@ export class EditableField extends React.Component {
         {fieldValue.map((val, index) => {
           const isActive = activeField === val.id
           const isDisabled =
-            val.disabled || this.state.disabledItem.indexOf(val.id) !== -1
+            disabled ||
+            val.disabled ||
+            this.state.disabledItem.indexOf(val.id) !== -1
```
plbabin, line 963: we could use includes instead of indexOf here

```suggestion
            this.state.disabledItem.includes(val.id)
```

plbabin (commented):
code looks really good. I really like extracting the toggler to another component 🎉 

I left one minor code comment.

Also I did found a bug on the netlify storybook. When opening the dropdown on an empty field, the dropdown is set to hidden. 

Screen Recording 2021-05-04 at 09 47 07 AM



Last thing, I think on the toggler we could add a `cursor: pointer` to let the user know it's clickable

src/components/EditableField/EditableField.jsx
tinkertrain: Cool, done!

tinkertrain:
@plbabin Fixed, thanks for taking a look at it!

I added a fix for https://helpscout.atlassian.net/browse/HSDS-181, more details on the description :)

plbabin (approved):
🎉 🎉 🎉

PR #922, opened 2021-05-05
Update message card padding according to the newest designs

jakubjanczyk:
# Problem/Feature

After a review of HSAPP task, we need to make some adjustment to the MessageCard visual, as we've missed some cases.

This PR for now includes increasing padding around the image to align with other elements. See details here: https://github.com/helpscout/hs-app/pull/9906


alexestrada (approved):

tinkertrain (approved):
💯

PR #924, opened 2021-05-10
Add error state to a DropList's SelectTag toggler

jakubjanczyk:
# Problem/Feature

This PR adds possibility to provide an error to a SelectTag toggler. The toggler would be styled with red border and `alert` icon is added with a tooltip containing provided error message. This functionality is needed for a Messages builder, but seems to be generic enough to do it here, instead of creating a customized component elsewhere.

There has been added Story presenting this state: `DropList` -> `Toggler: SelectTag with error`

Screenshot from 2021-05-10 12-47-07


plbabin (approved):
looks good, thank you for making this change!

PR #926, opened 2021-05-10
BEMBED-268: Updates to MessageCard component

Charca:
**[Jira Issue](https://helpscout.atlassian.net/browse/BEMBED-268)**

# Problem/Feature

In order to provide a better user experience when showing Messages with images in them, we're updating the logic so that the Message only becomes visible after the image has finished loading.

With this update we now have three scenarios to test:

1. A Message without an image should become visible right away.
2. A Message with an image should become visible after the image has finished loading.
3. A Message with an image that fails to load (with a 404 error for example) should become visible without rendering the image.

### DIFF
--- a/src/components/MessageCard/MessageCard.jsx
+++ b/src/components/MessageCard/MessageCard.jsx
@@ -10,6 +10,7 @@ import { MessageCardSubtitle } from './components/MessageCard.Subtitle'
 import { MessageCardImage } from './components/MessageCard.Image'
 import { MessageCardAction } from './components/MessageCard.Action'
 import { MessageCardBody } from './components/MessageCard.Body'
+import { MessageCardContent } from './components/MessageCard.Content'
 
 export const MessageCard = React.memo(
   ({
@@ -80,13 +81,14 @@ export const MessageCard = React.memo(
         >
           <MessageCardTitle title={title} />
           <MessageCardSubtitle subtitle={subtitle} />
-          <MessageCardBody
+          <MessageCardContent
             withMargin={title || subtitle}
-            body={body}
-            onClick={onBodyClick}
-          />
-          <MessageCardImage image={image} onLoad={makeMessageVisible} />
-          {children}
+            render={body || image || children}
+          >
+            <MessageCardBody body={body} onClick={onBodyClick} />
+            <MessageCardImage image={image} onLoad={makeMessageVisible} />
+            {children}
+          </MessageCardContent>
           <MessageCardAction action={action} />
         </MessageCardUI>
       </MessageCardWrapperUI>
--- a/src/components/MessageCard/MessageCard.jsx
+++ b/src/components/MessageCard/MessageCard.jsx
@@ -4,59 +4,12 @@ import getValidProps from '@helpscout/react-utils/dist/getValidProps'
 import MessageCardButton from './MessageCard.Button'
 import { classNames } from '../../utilities/classNames'
 import { noop } from '../../utilities/other'
-import {
-  ActionUI,
-  BodyUI,
-  ImageContainerUI,
-  ImageUI,
-  MAX_IMAGE_SIZE,
-  MessageCardUI,
-  MessageCardWrapperUI,
-  SubtitleUI,
-  TitleUI,
-} from './MessageCard.css'
-import Truncate from '../Truncate'
-
-const sizeWithRatio = (recalculatedSide, otherSide, defaultValue) =>
-  // Check if other side is smaller than max size to not recalculate unnecessarily this side as it doesn't need any scaling
-  // other condition checks that the image fits the boundaries
-  otherSide < MAX_IMAGE_SIZE
-    ? defaultValue
-    : (recalculatedSide / otherSide) * MAX_IMAGE_SIZE
-
-const getTruncatedText = (text, limit) => {
-  return (
-    <Truncate limit={limit} type="end">
-      {text}
-    </Truncate>
-  )
-}
-
-// Calculate size of image to keep the original aspect ratio, but fit within 278x278 square for image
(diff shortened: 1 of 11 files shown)


src/components/MessageCard/MessageCard.jsx
```diff
@@ -91,10 +122,20 @@ export class MessageCard extends React.PureComponent {
...
+    this.setState({ imageError: true })
+    this.makeMessageVisible()
+  }
+
   renderImage() {
     const { image } = this.props
+    const { imageError } = this.state
 
-    if (!image) {
+    if (!image || imageError) {
```
jakubjanczyk, line 138: Good catch with the image error case!

src/components/MessageCard/MessageCard.test.js
```diff
@@ -61,26 +60,88 @@ describe('Align', () => {
...
-    const wrapper = mount(<MessageCard animationSequence="scale" />)
-    const o = wrapper.find(Animate)
+describe('Visibility', () => {
+  jest.useFakeTimers()
+
+  test('Should be visible by default if there is no image', () => {
+    const onShowSpy = jest.fn()
+    const wrapper = mount(<MessageCard onShow={onShowSpy} />)
+
+    expect(wrapper.state('visible')).toEqual(false)
```
jakubjanczyk, line 70: It's not anything major and I am okay with pushing it out as it is. However, I don't like accessing the internal state in tests. I would rather check for the prop on a component that we are passing this state to (which is not perfect either, but feels a bit better), than accessing the state directly. Ideally, we could just check for set style on the element, but not sure if this is possible with styled-components and enzyme. What do you think?

jakubjanczyk (changes requested):
Looks great in general! Just left one comment :)

src/components/MessageCard/MessageCard.jsx
```diff
@@ -28,6 +28,37 @@ export class MessageCard extends React.PureComponent {
...
+  hasImage() {
+    return this.props.image && this.props.image.url
+  }
+
+  makeMessageVisible = () => {
+    setTimeout(() => {
+      this.setState({ visible: true })
+    }, 0)
+  }
+
```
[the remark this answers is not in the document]
jakubjanczyk: Sound good! (but I might be biased, as I don't think class components have much use nowadays :sweat_smile:) I can give it a try if Maxi wouldn't be able to take a look before Sabbatical. What do you think @Charca?

src/components/MessageCard/MessageCard.test.js
Charca: @jakubjanczyk that's a great point, and I agree that testing that the underlying styled-component has the correct styles would be a better test, since that's what we're trying to accomplish in the end.

There's a jest plugin that would allow us to do this: https://github.com/styled-components/jest-styled-components, which we'll need to install in this repo if we want to go that route. I would leave that up for you and @plbabin to discuss next week!

src/components/MessageCard/MessageCard.jsx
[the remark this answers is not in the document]
Charca: That sounds good to me! Would you mind making that update sometime next week @jakubjanczyk? I appreciate it!

Charca:
@jakubjanczyk thank you for offering to address the comments in this PR for me, I appreciate it! Once this is released you update hsds-react in Beacon, keep in mind that you'll need to:

- Pass the `withAnimation={true}` prop.
- Pass the `onShow` prop to fire the right events after the Message is shown.

Thanks again!

jakubjanczyk (approved):

plbabin (commented):
What do you think if we move all utility functions to a new file `MessageCard.utils` and adds some JSDocs for each function?

that would help cleanup the main file a bit, it's a bit long to read. Even splitting each render in a smaller component (and let those component decide if they should render or not), but doing that might makes it hard to find information quickly on what is render within the card, so not 100% sold on this suggestion. Any thoughts on that?

jakubjanczyk:
@plbabin Hmm... yes, moving utils makes sense. I'll do that. With the components. I agree that this function is now quite long, I am only hesitant to extract them and make it possible to import them directly, as they should not be used separated from MessageCard itself. I am not too worried about the conditional render logic being moved there, as here you either way have to go to the function to see when some part is rendered. I'll change it and we can always revert the commit.

jakubjanczyk:
@plbabin pushed - let me know what you think :)

src/components/MessageCard/MessageCard.jsx
```diff
@@ -1,187 +1,100 @@
...
-    const { height, width } = this.calculateSize(image)
+import { MessageCardUI, MessageCardWrapperUI } from './MessageCard.css'
+import { MessageCardTitle } from './components/MessageCard.Title'
+import { MessageCardSubtitle } from './components/MessageCard.Subtitle'
+import { MessageCardImage } from './components/MessageCard.Image'
+import { MessageCardAction } from './components/MessageCard.Action'
+import { MessageCardBody } from './components/MessageCard.Body'
+import { MessageCardContent } from './components/MessageCard.Content'
+
+export const MessageCard = React.memo(
```
plbabin, line 15: I did take a look at the implementation in beacon and hs-app, `innerRef` is not used at all. That's a good occasion to migrate this component to forwardRef and remove the `innerRef` prop entirely.

After that I am good to go!

jakubjanczyk: Good catch, changing it now :)

plbabin: awesome thank you!!!

plbabin (approved):
amazing work!!! 🎉

PR #929, opened 2021-05-18
HSDS-189 Update helix to 0.2.0

plbabin:
Helix was updated to the latest version of all packages (https://github.com/helpscout/helix/pull/31). The package was also updated to use a plain babel transpile instead of having both commonjs and es module

This update bring the latest from helix to HSDS. All tests are still passing, and the avatar 404ing is also fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1774,13 +1774,13 @@
       }
     },
     "@helpscout/helix": {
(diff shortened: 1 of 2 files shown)


jakubjanczyk (approved):
Looks good! I have not found anything breaking in Storybook :)

PR #930, opened 2021-05-19
BEMBED-275: Helpscout chat modal color contrast

maria-p:
# Problem/Feature
In an effort to make Beacon more accessible we are changing the way text color is calculated based on the pre-set background color to ensure better contrast.

There's a [formula](https://24ways.org/2010/calculating-color-contrast/) that we used to calculate if the text should be light or dark based on the given background color. Originally we tweaked the RGB constants from that formula to have less contrast and 'nicer' looking designs, but it rarely produced the results aligning with accessibility guidelines (especially for lighter Beacon colors). I've reset the constants to their original values and added some logic to actually evaluate the resulting color contrast instead of just relying on background shade luminosity only. 

Additionally, changing `backgroundColorInteractive` and `textColorInteractive` formulas used to highlight active beacon tabs - also for better contrast.

[Figma with beacon changes](https://www.figma.com/file/RkHMrOkWhmfytX5dkntBbO/Discovery%2F-Beacon-Embed-Accessibility?node-id=1097%3A5739)

Example of how it would look in beacon now (before/after):

More examples here: https://github.com/helpscout/beacon2/pull/921

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.14.0",
+  "version": "3.14.1-color-contrast-1",
```
luketlancaster, line 3: is this an acceptable version name for the HSDS project?

maria-p: I used it for local testing with beacon & hsapp, according to [this doc](https://helpscout.gitbook.io/hsds-react/testing/local-integration-testing). Should I roll it back now that I feel like PR is ready? Not sure what's the process here 😅

luketlancaster: lol you and me both 😁 . Wonder if @plbabin or @tinkertrain can chime in with direction?

tinkertrain: Hi! @maria-p , @luketlancaster

I think that at the moment of release the tool we use (np) sets the correct value there automatically, so there's no need to worry about as far as I know (I've never set my own name but something similar happens when releasing a beta version).

luketlancaster (approved):
This all looks great to me!

tinkertrain (approved):
very nice, thank you!

ps let us know if you need help releasing it

plbabin (approved):

PR #932, opened 2021-05-20
HSDS-186 new CheckmarkCard UI

plbabin:
# Feature
Screen Recording 2021-05-20 at 12 28 37 PM

This updates the CheckmarkCard component with the latest UI:

* Component was refactor to hooks and functional component
* Border is now used only for the focus state
* There is a light background on the selected state
* Focus outline + the mark will fade in/out on state change
* with status mark has a focus border matching the parent component
* d400 shadow + hover translate persist across state, but is removed on checked/with-status state
* keyboard enter is now working (making it even more accessible)

## Breaking change
Instead of passing a color, we decided to limit the range of modifications that could be done in that regard. Now it will be the color family and the change will be applied with classname. `blue` & `lavander` are the only valid option for now

Change will need to be applied on hs-app within the mailbox UserList component


nikkiathelpscout:
Just one change - let's remove the purple border on the locked card. Here's a screenshot: 
image
Then the focus border on the line icon doesn't overlap that purple border. Thanks!

plbabin:
@nikkiathelpscout I pushed an update on the component : 
Image 2021-05-20 at 2 31 11 PM

nikkiathelpscout (approved):
The screenshots look good to me. Would like to play with this in browser.

src/components/CheckMarkCard/CheckMarkCard.jsx
```diff
@@ -1,178 +1,171 @@
...
-    }
-    // If the card has a status provided, it should take precedence even if the card
-    // is checked from external props for some reason
-    if (withStatus) {
-      iconName = withStatus.iconName
-      iconSize = withStatus.iconSize || '20'
-      color = withStatus.color
-      tooltip = withStatus.tooltipText
+  useEffect(() => {
+    // Update the document title using the browser API
```
TerrenceLJones, line 119: Am I missing where we updating the document title?

TerrenceLJones (approved):
Just one tiny question, otherwise...LGTM!

src/components/CheckMarkCard/CheckMarkCard.jsx
plbabin: I bet I copied that useEffect from somewhere else 🤦

I will remove that comment before merging this PR

PR #937, opened 2021-05-28
CheckMarkCard improvements

plbabin:
# Problem
When implementing #932, we realized that a lot of code was duplicated across all instances of the CheckMarkCard. It was always the same code. We decided to bring all that code inside the HSDS instance to make it easier to implement when in hs-app. 

Image 2021-05-28 at 3 46 27 PM

- We removed the `withStatus` prop (better to have direct props instead of an object). It was replace with `markColor`, `tooltipText`, `status`, `iconName` and `iconSize`.
- We implemented the duplicate styling code from hs-app to the component. Default sizing is 170x160.
- We also did add the heading, subtitle and avatar that were present in hs-app. It's always possible to turn off all of them by omiting a value in the prop (or in case of the heading, passing false to `showHeading`. If there is no heading present, it will fall back to the label.
- Lastly, we create a new styled component for the grid, adding it as a static variable to the Checkmark component. The goal was once again to share duplicate code in a centralize place


jakubjanczyk (approved):
Looks good! :shipit:

tinkertrain (approved):
💯 Nice!

PR #938, opened 2021-06-03
Controlled Tooltip

plbabin:
This update adds the `visible` props to the tooltip in case we need to entirely control the tooltip from the outside (that's the case with the Editor).

When this prop exists (just declared on the props object is enough) it will make sure to remove the `showOnCreated` and `trigger` list of props that will be passed down to Tippy.

### Checkmark
Also, at the same time, we fixed a small color issue with the CheckmarkCard component.  We increase the color contrast for the subtitle.


tinkertrain (approved):
🚢

PR #939, opened 2021-06-04
[DropList] Improves up down navigation with all type of items present in the list

tinkertrain:
## Problem

When navigating with up/down keys:

- It skips disabled items correctly
- Internally it "highlights" group labels and dividers, an action that doesn't show up because those items are unselectable, so you had to press the key again to actually go to the next selectable item.
- When going one up from the first selectable item, downshift correctly goes to the last item in the list (effectively looping), but going down from the last one was not taking you to the top of the list.

## Solution

Adjust the `getEnabledItemIndex` function to take into consideration items that are group or dividers and ignore them, and also use a little bit of recursion in when we are the bottom of the list and press DOWN.


plbabin (approved):
:tada: well done!

PR #940, opened 2021-06-08
DropList UI tweaks

tinkertrain:
# Feature

While using DropList in the _real world_ we found a couple of enhancements:

- Don't always show the scrollbar (should depend on the number of items)
- Long text items should be truncated in one line with ellipsis
- Laser Targeting: if you need to target a specific List Item for styling (or else), you can add the property `className` to each item in the array, the value will be added to the existing css class names on the `DropListItem`:

```js
const items = [
  { label: 'a', className: 'paintItBlue' },
  { label: 'ab' },
  { type: 'divider' },
  { label: 'b', className: 'paintItGreen' },
]
```

will result in something like:

```html
<ul>
	<li class="DropListItem paintItBlue">a</li>
	<li class="DropListItem">ab</li>
	<li class="DropListItem DropListItem--divider"></li>
	<li class="DropListItem paintItGreen">b</li>
```

- `onSelect` callback got a bit more useful specially with multiple selection enabled:

```js
// before it only passed the "selection" argument, this one does not change

/**
 *
 * @param {string | object | array} selection The current selection, if multiple selection enabled, then it is an array
 * @param {string | object} clickedItem The item that was clicked, useful in multiple selection cases
 */
function onSelect(selection, clickedItem) {}
```

### DIFF
--- a/src/components/DropList/DropList.ListItem.jsx
+++ b/src/components/DropList/DropList.ListItem.jsx
@@ -5,11 +5,13 @@ import {
   getItemContentKeyName,
   isItemADivider,
   isItemAGroupLabel,
+  objectHasKey,
 } from './DropList.utils'
 import {
   DividerUI,
   GroupLabelUI,
   ListItemUI,
+  ListItemTextUI,
   SelectedBadge,
 } from './DropList.css'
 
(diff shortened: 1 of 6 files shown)


plbabin (approved):
so cool, thank you for jumping on that so quickly. I really love the `className` idea, well done!

PR #942, opened 2021-06-11
Adds some enhancements to DropList

tinkertrain:
# Feature

Continuing the replacement of Dropdown, found a few more things to improve:

- Cleaned up togglers file (moved all styled components into a css file)
- Adds 'is-active' class to all preset togglers
- Adds some options to `IconBtn` toggler (`withCaret`, `iconSize`, `caretSize`)
- Adds `a11yLabel` to togglers that are just icons so we can pass text that will not be rendered but readable for screen readers
- Adds the ability to enable Tooltip on icon-only togglers with the flag `withTooltip` using the `a11yLabel` text
- Adds `closeOnBlur` flag to DropList, mainly because is useful when debugging (so that the menu DOM Node can be inspectable on devtools)
- Improves the padding/margin of menu when there is only one item in the DropList

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -21,6 +21,7 @@ import ListItem, { generateListItemKey } from './DropList.ListItem'
 import { VARIANTS } from './DropList.constants'
 
 function Combobox({
+  closeOnBlur = true,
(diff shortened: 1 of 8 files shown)


plbabin (approved):
looks good! that `closeOnBlur` prop is really clever 🙌

PR #944, opened 2021-06-14
Fixes selection and beefs up test for ActionSelect

tinkertrain:
# Problem

The selection from props and state in `ActionSelect` was out of sync and used in a mixed manner:
- The state value was not set on mount (so it was being ignored)
- The state was being in one place and the prop value in another (causing the sync issue)
- The corresponding test while existing it was incomplete and so giving false positive

Also in the SplitButton toggler from Droplist, moved the classname from props to be applied to the root instead of the actual toggler button as it is a little more intuitive (you can target the toggler button with its own classname 'SplitButton__Toggler`)


jakubjanczyk (approved):
Looks good!

PR #945, opened 2021-06-15
Avoid setting state on unmounted Tooltip component

jakubjanczyk:
# Problem/Feature

In Tooltip component, there's `onShow` function that calls `setTimeout` with setting state inside. In most situations it works fine, but in case of ConditionField component, there's a tooltip used on `Remove` button. If you click it fast enough (before animation duration passed), then the component would be unmounted, but it would still try to set state after timeout, which results in a console warning about setting state on unmounted component, like this:
Screenshot from 2021-06-15 11-03-44

The solution is to save flag saying that component is mounted and do not set the state when this flag is false.

No new story has been added, but it can be checked on `Condition` story. Click `Remove` button next to `Specific URL` condition. It has to be done fast, before the tooltip has a chance to be displayed - just hovering over the button and immediately clicking it would be enough.

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/Tooltip/Tooltip.jsx
+++ b/src/components/Tooltip/Tooltip.jsx
@@ -1,4 +1,10 @@
-import React, { createContext, useContext, useState } from 'react'
+import React, {
+  createContext,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


tinkertrain (approved):
:shipit:

PR #946, opened 2021-06-18
[Bug] Fixes EditableFields bugs when blurring the inputs

tinkertrain:
## Problem

Reported: Not committing email addresses when clicking away from email field if there is already 2+ emails attached

The issue happened on multiple value EditableFields, when more than one value alreaady present, adding a new one and then blurring would break.

Multiple Values Before

While looking into it I found a similar but cause by different reasons bug in the multiple value with options fields, again blurring was broken:

With options before

## Solution

- Check if the EditableFields is a multiple option field, and put a guard if not.
- Persist the `Event` coming from blur

Multiple Values after

With options after


plbabin (approved):
👍 all good

PR #947, opened 2021-06-23
[Fix] DropList tweaks

tinkertrain:
# Problem/Feature

1. When a DropList lives inside a Modal, pressing ESC should only close the DropList  and not the Modal, this was already handled for the Select variant, here we add the Combobox variant to the list
2. Reported by @plbabin:
> So if I tab to the trigger press enter then tab again, the dropdown closed. That’s fine, but then I need to press enter twice to get it to open once again
so the sequence I did:
tab (trigger is focused ) + enter (opened)+ tab (closed, trigger is focused) + enter (nothing happened) + enter (opened) 

That was happening due to some complex event sequences, here we add a keyDown handler besides the onClick handler to deal with them separately.
3. DropList  togglers now will get the `DropListToggler` classname applied to them automatically and merge the user toggler classname from props.
4. In some scenarios using up/down arrows was still haveing some issues, sometimes skipping an item or making it "un-highlightable" with the keyboard, here we simplify the logic of `getEnabledItemIndex` significantly to fix these issues.

Try it in the stories, specially the most complex one that includes groups, dividers and disabled items: http://localhost:8900/iframe.html?id=components-dropdowns-droplist--items-groups-and-dividers&args=&viewMode=story, going up and down should highlight items skipping disabled items, group labels and dividers.

### DIFF
--- a/src/components/DropList/DropList.utils.js
+++ b/src/components/DropList/DropList.utils.js
@@ -201,64 +201,36 @@ export function getEnabledItemIndex({
 }) {
   let enabledItemIndex = 0
 
-  if (arrowKey === 'UP') {
-    for (let index = items.length - 1; index >= 0; index--) {
-      const isIndexItemHighlightable =
-        !checkIfGroupOrDividerItem(items[index]) && !items[index].isDisabled
-      const isNextIndexItemHighlightable =
-        items[nextHighlightedIndex] &&
-        !checkIfGroupOrDividerItem(items[nextHighlightedIndex]) &&
-        !items[nextHighlightedIndex].isDisabled
+  const isNextIndexItemHighlightable =
+    !checkIfGroupOrDividerItem(items[nextHighlightedIndex]) &&
+    !items[nextHighlightedIndex].isDisabled
 
-      if (
-        isIndexItemHighlightable &&
-        (index <= nextHighlightedIndex ||
-          currentHighlightedIndex === nextHighlightedIndex)
-      ) {
-        enabledItemIndex = index
-        break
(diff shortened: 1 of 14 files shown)


plbabin (approved):

src/components/DropList/DropList.utils.js
```diff
@@ -201,43 +201,36 @@ export function getEnabledItemIndex({
...
-        currentHighlightedIndex,
-        nextHighlightedIndex: 0,
-        items,
-        arrowKey,
-      })
+  if (
+    isNextIndexItemHighlightable &&
+    currentHighlightedIndex !== nextHighlightedIndex
+  ) {
+    enabledItemIndex = nextHighlightedIndex
```
jakubjanczyk, line 212: What about returning here early and do not add `else` on the next line. Would make the code a bit more clearer I think (at least there would be one less indentation)

src/components/DropList/DropList.utils.js
```diff
@@ -201,43 +201,36 @@ export function getEnabledItemIndex({
...
+  } else {
+    let newNextHighlightedIndex
+
+    if (arrowKey === 'UP') {
+      newNextHighlightedIndex =
+        nextHighlightedIndex - 1 >= 0
```
jakubjanczyk, line 218: I think this is something like `isNotLast` - could we maybe extract it to variable saying what exactly is checked here? Would be easier to get what is going on here :)
And the same on line 223

jakubjanczyk (approved):
Looks good! Just left a few comments to improve code a bit, but nothing blocking

tinkertrain:
@jakubjanczyk Thanks for the suggestions! I have made some changes and it does read much better

PR #948, opened 2021-06-29
Updated babel's preset-env loose mode to reduce component bundle size

Charca:
# Problem/Feature

While working on some performance improvements on the Beacon embed-side, we discovered that version v3.12.0 of hsds-react increased our overall bundle size by about 10% for no apparent reason. After some investigation, we found out that changing the `@babel/preset-env` `loose` setting to false, caused built assets to include a bit more code than before, adding some extra functions and variables during build time.

I don't think this `loose` setting has any impact on functionality, so this should be a pretty low-risk change. Also, I didn't change the setting in the Babel config used by Storybook because it didn't seem to have an effect, and leaving it as `true` gets rid of some warnings caused by Storybook plugins.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -48,7 +48,7 @@ module.exports = api => {
             safari: '12',
             node: '12.16.3',
           },
-          loose: false,
+          loose: true,
           shippedProposals: true,
         },
       ],


tinkertrain:
@Charca So sorry I forgot about this! The change is fine I think, maybe we should do the same on the storybook babel file? (`.storybook/.babelrc`)

tinkertrain (approved):
Just one little comment

plbabin (approved):

plbabin:
Since all our app (beacon, hs-app, hs-app-ui), I do wonder if we should let those app transpile HSDS as well. We could deploy all pure file to npm, nothing would be lose, and I guess it might be better for IDE autocomplete. Just thinking out loud

tinkertrain:
@plbabin I can explore this as part of trying to make npm link work

plbabin:
> @plbabin I can explore this as part of trying to make npm link work

I bet that would help with npm link too

Charca:
@tinkertrain I didn't change `.storybook/.babelrc` because setting the config to `true` in this file was causing a few warnings when running Storybook, which I think are mostly just noise. Since this flag doesn't make any functional difference, I thought I'd leave it as `false` in Storybook to get rid of the warnings, but let me know if you think we should make it both the same :)

PR #950, opened 2021-07-05
Update react-router to newest version and change not compatible components

jakubjanczyk:
# Problem/Feature

This PR updates react-router to the newest version and adds peerDependency to the library. The reason for that is currently used version has old Context API used and this might prevent us from updating React further in the future. Additionally, we would like to use newest version in the new has-app-ui code (for now - Standalone Messages app).

Few things has been done as part of this PR:
- `Route` component has been removed. It was just a copy of react-router's component of the same name and it was not really used anywhere outside of this repo. There seems to be no point in keeping this copy here
- `RouteWrapper` and `PortalWrapper` has been updated to not rely on old Context API for router. Instead, new `WithRouterCheck` HoC has been created that checks for existence of Context and wraps component with `withRouter` HoC that provides necessary props. This was necessary (instead of just wrapping with `withRouter` by default), because some components (like Button) might be used without Router Context, so this preserves backward compatibility. This was really the easiest way - one other solution would be to not mix regular Button with "Link" Button, but this would require a lot of changes in other repositories
- Some tests had to be fixed (some even were not really working before - not checking the right thing), due to introducing new Router Context


jakubjanczyk:
@plbabin @tinkertrain I am still fighting with myself if `WithRouterCheck` HoC would be the right way here, but any other option I though about is either very complicated or might break compatibility for old code. I'd like to have your opinion on this :)

plbabin (approved):
that look good. I would suggest to have a beta release so we could try it out within hs-app to make sure it doesn't break anything ( we could verify that on CI too)

jakubjanczyk:
> that look good. I would suggest to have a beta release so we could try it out within hs-app to make sure it doesn't break anything ( we could verify that on CI too)

Yes, I was planning on doing that. I'll do beta release once it's fully accepted :)

src/components/Drop/Drop.test.js
```diff
@@ -112,7 +112,7 @@ describe('Composed', () => {
 describe('Trigger', () => {
   test('Can render', () => {
     const TestComponent = Drop()(ContentComponent)
-    const wrapper = shallow(<TestComponent isOpen trigger={trigger} />)
+    const wrapper = shallow(<TestComponent isOpen trigger={trigger} />).dive()
```
tinkertrain, line 115: TIL: didn't know `dive` even existed, I don't write enzyme tests anymore but good to know for when updating existing ones it might make things easier, thanks!

tinkertrain (approved):
It all looks fine to me 💯  I'm sure it wasn't that much fun amending tests, thanks for all this work!

I like your solution to keep backwards compatibility, I agree that a refactor would be too much work for honestly little benefit at this point. Maybe one day...

I concur on the beta release.

src/components/Drop/Drop.test.js
jakubjanczyk: Funny story (well, funny now, it wasn't much fun back then :smile:). We've been using only `shallow` at some point and with a lot of refactoring or adding new features, tons of tests were failing, because of not rendered elements. It started to look like `shallow().dive().dive().dive().dive()...`, so we've created function `diveUpTo(selector)`, which was doing `dive()` in loop until selector was found. That wasn't great idea, but at least since then I am not using `shallow` anymore :smile:

PR #954, opened 2021-07-15
Adds onListItemSelectEvent prop to DropList

tinkertrain:
# Feature

> Copied from the addition I made to the docs for this feature:

### onListItemSelectEvent

In downshift item selection other than click is not handled directly on the list item nodes, in Select the events are handled on the Menu and in Combobox they are handled from the input. Because of this there is no access to any list item "event" on selection.

Most of the time you don't care _how_ the item was selected, but if you need to, you can use this callback where you'll find the `event` and get the `event.type`.

For this we have added `onListItemSelectEvent` which fires on any selection event (click, keydown). This callback gives you access to the actual List Item DOM node that was selected, and the original event.

The original click event will give you the actual list item, so `listItemNode === e.target`

Notice that on keyboard events `e.target` will be either the Menu (Select variant) or the Input (Combobox variant) and that is where `listItemNode` comes handy,

```js
/**
 *
 * @param {object} event Original event
 * @param {DOMNode} listItemNode The item that was selected
 */
function onListItemSelectEvent({ event, listItemNode }) {}
```

### DIFF
--- a/src/components/DropList/DropList.test.js
+++ b/src/components/DropList/DropList.test.js
@@ -729,9 +729,11 @@ describe('Togglers', () => {
 describe('Selection', () => {
   test('should select an item when clicked (string version)', async () => {
     const onSelectSpy = jest.fn()
+    const onListItemSelectEventSpy = jest.fn()
     const { getByText, getByRole } = render(
       <DropList
         onSelect={onSelectSpy}
+        onListItemSelectEvent={onListItemSelectEventSpy}
         items={beatles}
         toggler={<SimpleButton text="Button Toggler" />}
       />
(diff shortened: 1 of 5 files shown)


src/components/DropList/DropList.test.js
```diff
@@ -841,6 +867,12 @@ describe('Selection', () => {
         )
       ).toBeTruthy()
       expect(onSelect).toHaveBeenCalledWith(itemToSelect, itemToSelect)
+      expect(onListItemSelectEventSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          event: expect.objectContaining({ type: 'click' }),
+          listItemNode: getByText(itemToSelect.label).parentElement,
+        })
+      )
```
jakubjanczyk, line 875: Not to this, line best relates to tests :) Should we add some test case to cover Enter/Space keydown as well? Especially in Select version it's a new code that I believe would be worth testing

jakubjanczyk (commented):
Looks good! I've just left one comment in regards to tests, but implementation looks perfect :)

src/components/DropList/DropList.test.js
tinkertrain: Good idea, I just added them

jakubjanczyk (approved):
Looks :100:!

PR #955, opened 2021-07-19
Replace Select component with DropList in Condition 

jakubjanczyk:
# Problem/Feature

This PR replaces usage of `Select` component inside `Condition` with new DropList. There are no functional changes, but some properties had to be adjusted to the new format.

The new look can be found in Storybook -> Condition.

Screenshot from 2021-07-19 12-22-20

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.16.3",
+  "version": "3.16.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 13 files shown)


tinkertrain (approved):
Looks so much better!

One thing I noticed in the default condition story (it was present before but I just noticed it). In the first condition "specific url" changing it to any other type does not take effect as for example the last one does (time on page"). It made me think perhaps something was broken with this PR but no, as I mentioned this is present currently so I think the issue is how the story is coded in `/Condition/Condition.storiesHelpers.js`, if you have time it would be great to fix it as it is quite confusing... (I know is not your fault!)

jakubjanczyk:
Thanks @tinkertrain! I've done a small change to just not have any options for the first condition. Making it closer to real life scenarios might require a lot more work, so I'd rather leave it for later at the moment :)

PR #956, opened 2021-07-19
Fixes styled components props not working

tinkertrain:
# Problem

https://c.hlp.sc/WnuxvrQZ

Something changed in Styled Components above version 5.2 or so, that broke the following prop pattern:

```js
const PageUI = styled('div')`
  color: ${({color}) => color || 'white'}; // <-- ignored
`

//... 
<PageUI color={myColor} />
```

I don't know _exactly_ what changed but I did find that it was related to a `stylis` internal upgrade. We pass a custom stylis plugin in `HSDS.Provider` that adds the `hsds-react` scope, removing the plugin completely showed me that this was in fact where the issue was.

~This PR updates how that plugin adds the extra scope and fixes the issues.~

After discussion and assessing the issues with a few solutions we came up with we decided to pin to the last version of Styled Components without issue -> 5.1.1 as it seems that v6 will fix the issue and we won't have to have a half solution.

Also in this PR we upgrade to latest storybook (no breaking changes).


plbabin (approved):
looks good to me. We need to keep an eye on the styled-components project to see if they fix that issue internally (or if they add a way to namespace selectors).

We will also need to run cypress on hs-app to ensure this doesn't break anything (cc @ryan-mulrooney)

tinkertrain:
@plbabin Yes, let's not forget!

@ryan-mulrooney I've created a draft PR here https://github.com/helpscout/hs-app/pull/10240 for you to test it easily

Charca:
Great find with this fix, @tinkertrain! This issue had me scratching my head for a while as well.

I just have one question about this new behavior that maybe you or @plbabin can help me understand. Since this fix is causing the class names to get duplicated, once with the scope and once without it (i.e. `.foobar, .hsds-react .foobar { ... }`), isn't this the same as not using the scope plugin at all? (i.e. just `.foobar { ... }`).

My understanding was that the additional scope we're providing with this plugin was to prevent styles from "leaking out" to other elements of the page, but it seems like we'll be losing that benefit if the selectors are now scoped and unscoped for each rule. Am I missing any other benefits for which we're using this scope provider and plugin?

Charca:
By the way, another workaround that I was able to find playing around with this plugin was modifying the existing selector like this instead of pushing a new one to the array:

```
selectors[i] = `${scope.trim()} ${selectors[i]}${selectors[i]}`
```

Which causes the class names to look like this:

```
.hsds-react .foobar.foobar { ... }
```

This is also really ugly and overly specific, but it was the only way I could make styled-components don't mess around with our class names 😆

tinkertrain:
@Charca Yeah the double selectors thing is super annoying, and it seems that we can't avoid having some kind duplication.

I tried your line and it does seem to be working fine too.

We have 3 options, neither ideal but we have to choose one:

1. Double the selectors: `.someclass, .hsds-react .someclass`
	- Seems like it defeats the purpose of not leaking out styles 	
2. Double qualification: `.hsds-react .someclass.someclass`
	- Makes the inspection quite noisy, not sure what if there are other implications due to the increased specificity
3. Skip the plugin: `.someclass`

A question we should ask ourselves... do we really need the plugin? Is there a _practical_ benefit to having each selector scoped? Maxi you are more of an expert in terms of working with an app that lives in an iframe (Beacon), would removing the plugin have an effect there?

Personally, my opinion is to just go for No.3, but I might be missing something.

cc @plbabin

plbabin:
> @Charca Yeah the double selectors thing is super annoying, and it seems that we can't avoid having some kind duplication.
> 
> I tried your line and it does seem to be working fine too.
> 
> We have 3 options, neither ideal but we have to choose one:
> 
> 1. Double the selectors: `.someclass, .hsds-react .someclass`
>    
>    * Seems like it defeats the purpose of not leaking out styles
> 2. Double qualification: `.hsds-react .someclass.someclass`
>    
>    * Makes the inspection quite noisy, not sure what if there are other implications due to the increased specificity
> 3. Skip the plugin: `.someclass`
> 
> A question we should ask ourselves... do we really need the plugin? Is there a _practical_ benefit to having each selector scoped? Maxi you are more of an expert in terms of working with an app that lives in an iframe (Beacon), would removing the plugin have an effect there?
> 
> Personally, my opinion is to just go for No.3, but I might be missing something.

1. I am not afraid about our styles leaking out (I doubt classname will collide with another styled-components instance). However our global style might need a refresh because it relies on the hsds-react wrapper to set some default on the page. We already set the SC_ATTR to that multiple instance of styled-components are not playing in the same `<style>` tag. https://styled-components.com/docs/advanced#avoiding-conflicts-with-thirdparty-styles-and-scripts. Downside of this implementation is in the browser devtool (harder to read selectors). Also I do wonder if it can slow down the browser CSS parsing step 🤔 

2. I did tried that locally, but this option is causing issues with complex selectors (for example we lose some border radius on the `ButtonGroup` component)

3. Skipping it will cause side-effect, but only when integrated in hs-app. Old SCSS files contain a lot of default styles directly applied on elements (`<a>`, `<button>`, `<input>`, etc). that will messed up our HSDS pixel perfect implementation. One option could be to use https://github.com/QuickBase/babel-plugin-styled-components-css-namespace on all instance were we compile this project (HSDS, hs-app, beacon, editor).

tinkertrain:
@plbabin The plugin sounds like a good solution for now, I'm about to sign off for the week so if you want to try it on this PR or create a new one feel free to do so

cc @Charca (same thing)

Charca:
Unfortunately, it looks like the babel plugin won't solve this problem, since the same issue that broke our implementation is affecting them as well 😢 https://github.com/QuickBase/babel-plugin-styled-components-css-namespace/issues/70

I think my preferred options would be to either: a) remove the scope plugin and see if we can deal with the specificity issues in hs-app somehow, or b) keep using styled-components 5.1.1 which is the only option where we don't have to compromise functionality or do any extra work. Of course, it'll have the drawback of not being on the latest version, but since v6 is on its way, I don't think they're planning to release too many 5.x versions in the future.

As I mentioned in the Slack thread, I've tested the pre-alpha of v6 and it does include a fix for this problem, so I'm very hopeful. But if for any reason v6 doesn't fix this, we can then come back to this discussion and consider all the other options. What do you all think?

Charca (commented):
@tinkertrain @plbabin unfortunately, it looks like the babel plugin won't solve this problem, since the same issue that broke our implementation is affecting them as well 😢 QuickBase/babel-plugin-styled-components-css-namespace#70

I think my preferred options would be to either: a) remove the scope plugin and see if we can deal with the specificity issues in hs-app somehow, or b) keep using styled-components 5.1.1 which is the only option where we don't have to compromise functionality or do any extra work. Of course, it'll have the drawback of not being on the latest version, but since v6 is on its way, I don't think they're planning to release too many 5.x versions in the future.

As I mentioned in the Slack thread, I've tested the pre-alpha of v6 and it does include a fix for this problem, so I'm very hopeful. But if for any reason v6 doesn't fix this, we can then come back to this discussion and consider all the other options. What do you all think?

tinkertrain:
@Charca Thanks for all this, I think I'd go with **b**, as it sounds 6 will most likely fix this for us.

I'll update this PR to pin styled components to and do the same in HS App on this PR https://github.com/helpscout/hs-app/pull/10240

plbabin (approved):

Charca (approved):

PR #958, opened 2021-07-26
[Table] Refactor and extra features

tinkertrain:
# Feature

As part of the Mailbox Rewrite, we take a look at our current Table component and bring some features to make it usable as a suitable ConvoList.

### Refactor

While adding the select row feature, performance issues became a problem and for this reason I refactored the table to use Hooks, specifically wanting to use _memoization_ and `useReducer`.

### Access to nested data in columns

While inspecting the data in the current list, I noticed that some of it was nested, ie:

```javascript
customer: {
	name: 'Ringo',
	email: 'ringo@thebeatles.music'
}
```

We can now access this data by providing `columnKey` a dot notation string, for example:

```javascript
const columns = [
	{
    	title: 'Customer',
    	columnKey: ['customer.name', 'customer.email'],
	}
]
```

If using a customized cell rendering, you get access to this data by replacing the dot `.` with an underscore `_`:

```jsx
renderCell: ({ customer_name, customer_email, row }) => ( // replace . with _
        <div>
          <strong>{customer_name}</strong>
          <br />
          <span>{customer_email}</span>
        </div>
      ),
```

> `renderCell` now also gives you access to the `row` data in case you need it.

### Selecting rows

We add the ability to select rows, with the following props:

- `withSelectableRows`: Enable the feature
- `selectKey`: A unique key from your data to identify selection, defaults to `id`
- `onSelectRow`: Callback that gives you an array of selected items (by the key provided in `selectKey` which by default is `id`)

This feature includes select / unselect all items from the checkbox in the table header row.

🎥 https://c.hlp.sc/L1uBqEGl

### Focusable Rows

We add the prop `withFocusableRows`, which simply adds a `tabindex = 0` to each row so you use can <kbd>tab</kbd> to navigate them, and if `onRowClick` is provided, execute the action on <kbd>enter</kbd>

### Conditional class names for Rows

If you need to add a class name to a row based on a condition, you can do it with `rowClassName`:

```jsx
rowClassName={row => {
	return {
  	[CONVO_STATUS_CLASSNAMES[row.status]]: true,
  	replying: row.id === 281796231,
  	viewing: row.id === 281796229,
	}
}}
```

From the function you can return anything [classnames](https://github.com/JedWatson/classnames) accepts, like a string or an object.

### Wrapping rows

Sometimes you might need to wrap a row or add attributes to it based on certain conditions, for this we have added the prop `rowWrapper`. Works like this:

```jsx

// children: the "row" react component (Table.Row)
// The data for this particular row
rowWrapper={(children, row) => {
  if (row.id === 281796231) {
	// Wrap with a `Popover`
    return (
      <Popover
        triggerOn="mouseenter"
        appendTo={() => document.body}
        withTriggerWrapper={false}
        placement="left"
        renderContent={() => (
          <Avatar
            size="xs"
            name={AvatarSpec.generate().name}
            image={AvatarSpec.generate().image}
          />
        )}
      >
        {
			// Add a `title` attribute on the row
			React.cloneElement(children, {
	          title: 'Someone is replying',
    	    })
		}
      </Popover>
    )
  }
  return children
}}
```

Note that while you can technically wrap the Row with any element, you are dealing with a `<tr />` and generally they don't like to be inside other stuff (like a span for example)

### ConvoList story

We include a new story, which is basically a reproduction of the ConvoList, ideally we can use this as a pretty complete base to use in HS App's Mailbox Rewrite.

Visit: https://deploy-preview-958--hsds-react.netlify.app/?path=/story/components-structural-table--selecting-rows

> Doesn't include sorting at the moment

#### Checkbox

Just noticed that class names passed to it weren't being applied. Fixed.

### Design

Design is not 100% there yet as I focused on features and I think design decisions for the ConvoList might not be finalized yet.

### Tests

Updates and adds a bunch of tests

## Next

Most likely there will be more features or tweaks to be done to `Table` but the aim of this PR is to enable the usage of it asap and we can iterate as needed, so to merge this PR we don't need this to be "complete".


cen10:
@digitaldesigner Would you mind taking a look please?

src/components/Table/Table.jsx
```diff
@@ -1,203 +1,161 @@
...
-  wrapperRef: noop,
-  tableRef: noop,
-  onExpand: noop,
-  withTallRows: false,
+  tableClassName,
+  tableWidth = { min: '700px' },
+  withSelectableRows = false,
+  withTallRows = false,
+}) {
+  const [state, dispatch] = useReducer(reducer, {
```
plbabin, line 42: I think even though it is for the initialState of the reducer, the getDisplayTableData will be called on each component render phase (I could be wrong here)

plbabin (commented):
looking really good, I love how hooks/functional component cleanup the code quite a bit

What I like to do with reducer is to create a custom hook that wrap it. Then I can create multiple actions that will be returned by the hook. That way I don't have to pass around the dispatch and action.type all around.

ie: that's the one I created for the noty component: https://gist.github.com/plbabin/4840990df22a520054dcd9de5e9a0e3d

digitaldesigner:
Looking great. Couple additional things to consider here:

- Entire rows should be clickable (i.e. to get to the conversation) in addition to being selectable
- Row hover/tab-focus state (soft gray?)
- Consider how the customer name will overflow when it is a super long email address
- Thread count
- The ability to cap the conversation preview to just one line of text and overflow with a soft shadow (as we do in the main convo list today)
- Ability to replace the select box with an icon, so that we could use this component in the Customer / Profile / Conversations table

tinkertrain:
@digitaldesigner Thanks! I made several updates, getting there I think, the one I don't understand is the last one, can I see an example?

cen10:
@tinkertrain I think Buzz is referring to how we can show an icon next to the conversation here:
Screen Shot 2021-07-27 at 12 11 47 PM

tinkertrain:
@cen10 @buzz Ah, that is already possible as you can add columns as you see fit and render whatever you want in them

src/components/Table/Table.jsx
tinkertrain: Good catch, memoized it and it fixes that

tinkertrain:
@plbabin Nice pattern, I've refactored the reducer into something very similar to that. Thanks!

tinkertrain:
@cen10 I've updated the "ConvoList" code to make it easier to follow what is going on, let me know if you have any suggestions or questions on anything you find in there, my hope is eventually we can take that almost as is and bring it in and just tweak as needed.

cen10 (approved):
Awesome job!

plbabin (approved):
that is so cool that you were able to refactor with hooks and replicated (and improve) the mailbox convo list. 

One thing to keep in mind would be to QA the customer list page to make sure it didn't not break something. other then that I am good with all those change (I will never say no to hooks refactor 😄  )

cen10:
Ah PL has a GREAT point about QAing the customer list page.

buzz:
Please stop mentioning me in this issue. Thank you very much

PR #959, opened 2021-07-28
Adds classnames package and replaces our light version usage

tinkertrain:
# Problem/Feature

I added the [classnames](https://github.com/JedWatson/classnames) package to HSDS, I honestly didn't pay too much attention but we were actually using a very simple implementation coming from hsds-react-utils. I replaced all those instances with `classnames`. This package is what we use in HSApp too. Consistency ftw.


plbabin (approved):
all good with me and all tests are passing!

I like this change a lot! well done 👏

PR #957, opened 2021-07-22
Fixes Droplist scrolling the document

tinkertrain:
# Problem

@Charca reported: 
Whenever I open a dropdown, it scrolls me back to the top of the page. Any ideas why this might be happening? My usage of this component is quite simple:
```
<DropList
  items={messageOptions}
  toggler={
    <MessageOptionsButtonUI
      onClick={handleButtonClick}
      a11yLabel="Message options"
    />
  }
/>
```

https://c.hlp.sc/yAur5nZG

The issue is fixed by using `showOnCreate` instead of `visible` on the `<Tippy />` component. Don't ask me why though 😅

PR #960, opened 2021-07-29
DropList fix > manipulate tippy instance manually to show and hide

tinkertrain:
# Problem

By removing the usage of `visible` here https://github.com/helpscout/hsds-react/pull/957, we have another problem that the tippy is not getting removed from the DOM in some instances (we are only making it invisible).

# Solution

Since `visible` is the only way react-tippy provides to [control the open/close state of the tippy](https://github.com/atomiks/tippyjs-react#visible-boolean-controlled-mode) (and we do want to control it) but continues to be buggy by scrolling the document we resort to manipulating the tippy instance instead.

While at it, I added a new prop `getTippyInstance` to retrieve the tippy instance in case you want to do stuff with it. This prop exists in Tooltip too so might aswell have it here too.

### DIFF
--- a/src/components/DropList/DropList.Select.jsx
+++ b/src/components/DropList/DropList.Select.jsx
@@ -33,7 +33,6 @@ function Select({
   selectedItem = null,
   selectedItems,
   toggleOpenedState = noop,
-  setMenuBlurred,
   withMultipleSelection = false,
 }) {
   const {
@@ -70,10 +69,6 @@ function Select({
 
     onStateChange({ type }) {
       switch (type) {
-        case useSelect.stateChangeTypes.MenuBlur:
-          onMenuBlur()
-          break
-
         case useSelect.stateChangeTypes.MenuKeyDownSpaceButton:
         case useSelect.stateChangeTypes.MenuKeyDownEnter:
         case useSelect.stateChangeTypes.ItemClick:
@@ -167,6 +162,7 @@ function Select({
     >
       <A11yTogglerUI {...getToggleButtonProps()}>Toggler</A11yTogglerUI>
       <MenuListUI
+        data-event-driver
         className={`${DROPLIST_MENULIST} MenuList-Select`}
         {...getMenuProps({
           onKeyDown: handleMenuKeyDown,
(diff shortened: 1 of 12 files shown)


src/components/DropList/DropList.jsx
```diff
@@ -219,15 +212,32 @@ function DropListManager({
...
     onOpenedStateChange(shouldOpen)
+    toggleTippy(tippyInstance, shouldOpen)
   }
 
   return (
     <Tippy
       {...tippyProps}
+      onCreate={instance => {
+        setTippyInstance(instance)
+        getTippyInstance(instance)
```
jakubjanczyk, line 236: Can you explain why there's `set` and `get` called right after each other? That's probably fine, but I am just curious what it does here :)

jakubjanczyk (approved):
That was indeed tricky bug, thanks for finding and implementing the solution! :) :ship:

src/components/DropList/DropList.jsx
tinkertrain: I removed this part, now using a `ref` instead of storing the tippy instance in state

tinkertrain: FYI `getTippyInstance` is a callback prop, in case you want to do something with it from the outside

src/components/DropList/DropList.Select.jsx
```diff
@@ -166,12 +161,36 @@ function Select({
...
+               * so it sets the isOpen flag to false, and then the click event happens
+               * and sets isOpen to true, making the DropList never close on cliking
+               * the toggler.
+               *
+               * Here we wait a little bit to see if the next element gathering focus
+               * is indeed the toggler, and only close the DropList if it isn't
+               */
+              setTimeout(() => {
+                if (
+                  !document.activeElement.classList.contains('DropListToggler')
```
jakubjanczyk, line 186: If there is a custom toggler, would this class be added automatically, or when creating it, we must remember to add this class?

tinkertrain: gets added for you 👍

PR #962, opened 2021-08-16
New components: SidePanel & SimpleModal

tinkertrain:
# Feature

Design specs: https://www.figma.com/file/v8ReousbOpHN2g9n5n6StX/Standalone-Messages---2.0?node-id=342%3A2552

## SidePanel

We add a new overlay component: SidePanel. This is sort of a modal that slides from the side.

[Docs and stories](https://deploy-preview-962--hsds-react.netlify.app/?path=/docs/components-overlay-sidepanel--default-story)

### SidePanel Layouts

We have some layouts that are going to be commonly used, in order to avoid duplication we include what we call "layouts" which are simple components that you can drop in the `<SidePanel>` and forget about styling. They live inside `SidePanel.layouts.jsx`.

#### HeaderAndFooter layout

Includes:

- Header with a heading and a sub-heading
- Scrollable body (`children` get rendered here). Shadows get added to header and footer depending on the scroll state of the body.
- footer with a CTA.

Example usage:

```jsx
import { HeaderAndFooter } from './SidePanel.layouts'

function YourApp() {
  return (
    <SidePanel>
      <HeaderAndFooter
        mainActionButtonText="Customize this"
        panelHeading="Customize this"
        panelSubHeading="Customize this"
        onMainActionClick={handleClick}
        mainActionDisabled={!isDisabled}
      >
        <div>Your content here</div>
      </HeaderAndFooter>
    </SidePanel>
  )
}
```

## SimpleModal

This is a new Modal component. We already have a Modal component that is bloated and would have required more work to achieve what we needed.

The hope is that eventually we use this as a default Modal and phase out the existing Modal, which is a mess and everybody hates it. One idea is to bring a concept similar to the DropList "togglers" in which we provide default layouts for the most common modal types we use (warning, confirmation, etc) that way we can keep adding to that list without adding a gazillion different props (some of which don't work together) and keep the underlying component as lean as possible.

[Docs and stories](https://deploy-preview-962--hsds-react.netlify.app/?path=/docs/components-overlay-simplemodal--default-story)

## Considerations

Both components are basically the same in terms of implementation, here are some of the characteristics of their design:

- Both are fully controlled, they don't store any state about their opened/closed status. Since the user will normally have to create some state that holds this in their app, it seems unnecessary to add the complexity of keeping the `show` prop and an internal piece of state in sync.
	- Use the `onClose` to close the component, it gets fired when clicking the <kbd>x</kbd> and on hitting <kbd>esc</kbd>
- They both will be contained within the most immediate parent with a `position: relative` applied to them, this allows you to create layouts where the SidePanel does not cover the navigation or to open the a modal that only covers said SidePanel, just like the design spec requires.
- If you need different styles, use Styled Components, there's nothing preventing you to do this and keeps all the props to the minimum needed.
- If you need to portal the components, either use the existing Portal component or the newly added `usePortal` hook.

Full example

## Custom hooks

We add a new root-level section "hooks" where custom shareable hooks live.

### useAnimatedRender

A lot of the times we need to animate a component before it gets mounted and unmounted. On mounting is easy enough, but for unmounting you need to hold the unmount, animate and then continue with the unmounting.

This hook allow you to do that, both SidePanel & SimpleModal use this hook, see those for examples of application and the documentation provided for more info.

[Full Docs](http://localhost:8900/?path=/story/hooks-useanimatedrender--page)

### useClickOutside

Custom hook that adds a click event listener on the document that gets executed when clicking outside a given DOM Node passed as a ref.

[Full Docs](http://localhost:8900/?path=/story/hooks-useclickoutside--page)

### usePortal

We already have a Portal component, here we add a hook that you can bring into any component if you prefer using a hooks solution. The implementation is leaner and has a couple of advantages like creating a container with the selector provided if it doesn't exist.

It includes a Portal component for convenience, which you can import from the same file.

[Docs and stories](http://localhost:8900/?path=/story/hooks-useportal--page)

## Documentation

Everything is documented, and you'll see a new hooks section on storybook.


src/components/SidePanel/SidePanel.stories.mdx
```diff
@@ -0,0 +1,85 @@
...
+        'https://www.figma.com/file/v8ReousbOpHN2g9n5n6StX/Standalone-Messages---2.0?node-id=342%3A2552',
+    },
+  }}
+/>
+
+# SidePanel
+
+Component that brings a Panel from either side of the screen, it includes a default Header and Footer (both can be turned off via their respective props if you need to fully customize the panel's contents). The footer contains a CTA button.
+
+This component is fully controlled via the `show` prop, meaning it doesn't hold any internal state about whether is open or not, it's in your hands to manage this. The panel will fire the close action (`onClose` callback prop) with the &times; close button and via pressing the <kbd>esc</kbd> key.
```
jakubjanczyk, line 22: I really like that! I mean, a lot of components in HSDS have some internal state as a duplication of provided props and in most of the cases it only adds complexity. Having it outside makes that this component is quite simple :)

src/components/SidePanel/SidePanel.test.js
```diff
@@ -0,0 +1,203 @@
...
+
+    expect(queryByTestId('sidepanel-overlay')).not.toBeInTheDocument()
+
+    rerender(<SidePanel show />)
+    fireEvent.animationEnd(queryByTestId('sidepanel-overlay'))
+
+    expect(queryByTestId('sidepanel-overlay')).toBeInTheDocument()
+    expect(
+      queryByTestId('sidepanel-overlay').classList.contains('element-in')
+    ).toBe(true)
```
jakubjanczyk, line 20: What do you think about using `toHaveClass` assertion function? It would give better error message if failed :)

src/components/SidePanel/SidePanel.test.js
```diff
@@ -0,0 +1,203 @@
...
+    const header = container.querySelector('.SidePanel__Header')
+
+    expect(header).toBeInTheDocument()
+    expect(header.tagName).toBe('HEADER')
+
+    const heading = header.querySelector('h1')
+
+    expect(heading).toBeInTheDocument()
+    expect(heading.getAttribute('id')).toBe('sidepanel-header-heading')
+    expect(heading.textContent).toBe('Review and Start')
```
jakubjanczyk, line 47: Similar to above, in those 2 lines we could use `toHaveAttribute` and `toHaveTextContent` as well. And the same could be applied to some tests below as well.

src/hooks/useAnimatedRender.js
```diff
@@ -0,0 +1,65 @@
...
+      setRender(true)
+    }
+  }, [show])
+
+  function onAnimationEnd(e) {
+    if (e.target === toAnimateRef.current) {
+      if (!show) {
+        setRender(false)
+
+        if (toAnimateRef.current) {
```
jakubjanczyk, line 26: I think it would never be false? if `e.target === toAnimateRef.current` is true, and I think `target` would never be undefined/null (this is the place I have not checked, but my intuition tells me that :smile:), then there should be no need to have this additional check

jakubjanczyk (commented):
Looks and works great! 
I have only found one visual issue with scrollbars appearing for a moment while animation is running, you can see it here: https://c.hlp.sc/P8uGYL8K. I've checked in Chrome and Firefox, both looks the same. It's not a big issue, but would be good if can be fixed to avoid the jump on the page.

src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,153 @@
...
+        aria-modal="true"
+        aria-labelledby="sidepanel-header-heading"
+        className="SidePanel"
+        data-cy="sidepanel"
+        data-testid="sidepanel"
+        id="sidepanel"
+        panelWidth={panelWidth}
+        ref={panelRef}
+        onFocus={e => {
+          console.log('FOCUSED')
```
Charca, line 80: Stray console.log here. Also, should `onFocus` come as a prop to this component?

src/components/SidePanel/SidePanel.jsx
```diff
@@ -0,0 +1,153 @@
...
+      ref={overlayRef}
+      zIndex={zIndex}
+    >
+      <SidePanelUI
+        aria-modal="true"
+        aria-labelledby="sidepanel-header-heading"
+        className="SidePanel"
+        data-cy="sidepanel"
+        data-testid="sidepanel"
+        id="sidepanel"
```
Charca, line 76: Do we need an ID here? Looks like a hard-coded ID will restrict us to a single instance of the side panel on each page.

src/components/SimpleModal/SimpleModal.jsx
```diff
@@ -0,0 +1,101 @@
...
+      ref={overlayRef}
+      zIndex={zIndex}
+    >
+      <SimpleModalUI
+        aria-modal="true"
+        aria-labelledby={ariaLabelledBy}
+        className="SimpleModal"
+        data-cy="simple-modal"
+        data-testid="simple-modal"
+        id="simple-modal"
```
Charca, line 61: Same question about the ID here.

plbabin (approved):
I really like that we start creating custom hook to help improve other component within hs-app. I think the `usePortal` one could make his way into other hsds component as well. 

I did also had the scrollbar issue with the SidePanel story. I think we could have an optional prop to "fixed" the SidePanel on the screen (or we let dev implement that if needed by overwriting the css with styled-components). Either way I am fine with that. 


Both component looks really good, they are minimal in their implementation and I like that!

Charca (commented):
Looks great @tinkertrain, fantastic job! I left a couple of minor comments here and there.

src/hooks/useAnimatedRender.js
tinkertrain: Good catch!

src/components/SidePanel/SidePanel.jsx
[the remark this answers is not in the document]
tinkertrain: Yes... thanks!

tinkertrain: This one is leftover from debugging... removed!

src/components/SidePanel/SidePanel.jsx
tinkertrain: I added the `{...rest}` so you can override any of those attributes on the Panel element

src/components/SimpleModal/SimpleModal.jsx
tinkertrain: Same as above :)

tinkertrain:
@jakubjanczyk @plbabin @Charca 

I managed to get rid of the scrollbars I think, let me know!

Also, sorry to move the goal posts here... but I modified a bit the component, simplifying further.

I added the concept of "component layouts"... let me know what you think! (more info on the description and the documentation on storybook)

src/components/SidePanel/SidePanel.test.js
tinkertrain: I wonder if we have that in HSApp, it's a good idea, though I'd like to add those in a different PR and perhaps add them to HSApp (or bring them from there)

src/components/SidePanel/SidePanel.test.js
tinkertrain: ditto

src/components/SidePanel/SidePanel.test.js
jakubjanczyk: I have been using them, so they are for sure :smile: It's from `@testing-library/jest-dom` library that is set up in `config/jest/setupTests.js` in hsds repo already

src/components/SidePanel/SidePanel.layouts.jsx
```diff
@@ -0,0 +1,125 @@
...
+  const scrollableHeight = scrollable.offsetHeight
+  const scrollablePaddingTop =
+    Number.parseInt(style.paddingTop.replace('px', ''), 10) - 10
+  const scrollablePaddingBottom =
+    Number.parseInt(style.paddingBottom.replace('px', ''), 10) + 10
+
+  let isTopScrolled = false
+  let isBottomScrolled = false
+
+  if (scrollable.scrollTop - scrollablePaddingTop > 0) {
```
jakubjanczyk, line 95: What do you think about using ternary operator here, so that there's no need to create `let` variables? I like it more because it's more concise, but not something that is blocking :)

jakubjanczyk (approved):
Looks good! No more scrolls for me and I like the concept of easily replaceable content!
I've left one question, but it's good to go :)

src/components/SidePanel/SidePanel.layouts.jsx
tinkertrain: Ahh actually variables are not even needed, updated :)

src/components/SidePanel/SidePanel.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+  // Leave the `selector` prop out if you want document.body
+  return (
+    <Portal selector="#some-id">
+      <SidePanel>Some content</SidePanel>
+    </Portal>
+  )
+}
+```
+
+## SidePanel Layouts
```
plbabin, line 44: oooh I like that, 💯  sure layouts will make his way to Modal 😄

src/components/SidePanel/SidePanel.test.js
tinkertrain: Talk about not knowing what we have... thanks! (I changed the assertions)

Charca (approved):

PR #963, opened 2021-08-16
[Feature] Table - Column chooser

tinkertrain:
# Feature: Table > Column Chooser

CleanShot 2021-08-19 at 15 12 32

Enable by setting `withColumnChooser` to `true`.

When setting your columns:

- Add a `show: true` to each column you want to show by default
- Add a `group: 'Something'` to group columns in the Dropdown, if skipped, they will be grouped under 'Other'
- If you need a column to be there but disabled, add `disabledForChoosing: true`, this way the option will be disabled for the user (you can have it selected or not with show = true or false).

Example:

```js
const columns = [
  {
    title: 'Name',
    columnKey: 'name',
    width: '25%',
    group: 'Customer',
    show: true,
    disabledForChoosing: true,
  },
  {
    title: 'Company',
    columnKey: 'companyName',
    width: '25%',
    group: 'Customer',
    show: false,
  },
]
```

### Header content

Pass a valid element to `headerContent` to render it inside the `header` above the table on the left.

```jsx
<Table
  columns={myColumns}
  data={createFakeCustomers({ amount: 10 })}
  tableDescription="Example table"
  withColumnChooser
  headerContent={<h2>Such a cool table</h2>}
/>
```

To complete this feature, DropList got an addition:

### With Reset Selection Item

Use `withResetSelectionItem` to add an extra item to the end of the list that will have a type of 'reset_droplist' so you can assign a different action when clicked, this normally only makes sense in multiple selection DropLists.

Pass a string to customize the text, by default is "Reset".

Example:

```jsx
<DropList
  items={items}
  onSelect={(selection, clickedItem) => {
    if (clickedItem.type === 'reset_droplist') {
      clearSelection()
    }
  }}
  selection={columns.filter(shouldColumnShow)}
  withMultipleSelection
  withResetSelectionItem="Clear selection"
/>
```

So did Pagination:

### renderCustomContent

Sometimes you'll need complete control over the info section of the pagination, use `renderCustomContent` render prop:

```jsx
<Pagination
  totalItems={totalItems}
  rangePerPage={30}
  subject="conversation"
  activePage={5}
  renderCustomContent={({
    startRange,
    endRange,
    numberOfPages,
    currentPage,
    pluralizedSubject,
  }) => {
    return (
      <div style={{ fontFamily: 'monospace' }}>
        {totalItems} total {pluralizedSubject} • 26 active
      </div>
    )
  }}
/>
```


digitaldesigner (changes requested):
@tinkertrain You were absolutely right with your initial implementation of the disabled row value — I'm so sorry. Could you change the disabled row text to regular font weight and `charcoal.400` (grey checkmark is fine). Also would you be able to remove the blue focus circle on the icon? I think it's best we keep consistency on this one for now

digitaldesigner:
This might not be the correct place for general table rebuild feedback, but I wanted to document _somewhere_ — feel free to disregard or move if it's not relevant right now.

[**Watch walkthrough of feedback**](https://www.loom.com/share/1d99d40a1ccc46de86f644c43c99eb29 )

### Details
- Font sizes to be `13px` except for customer name (stay `14px`)
- Font color depends on convo status
- - Active row: `charcoal.600` for everything except convo preview (stay `charcoal.400`)
- - Assigned/pending row: `charcoal.500` for dark text
- - Closed/spam row: `charcoal.400`
- Remove row hover
- Selected rows to have a `blue.100` highlight (instead of yellow)
- Make the tab focus border (left blue line) `2px` width
- Any possibility of using the solid shadow on the folder headings when the page is scrolled? [See demo](https://cdpn.io/team/helpscout/debug/7a04a0396b147b146a3c71a3d0d6de0a) and [Figma documentation](https://www.figma.com/file/LSGu4LyWM46lG2xgrWraGP/1.-Foundations?node-id=402%3A272)

### Further Reading
- [Figma design files](https://www.figma.com/file/8rXMLTfIx9RTkihPhHU5B4/Spec-Custom-Columns-Folder-Rebuild?node-id=23%3A0)
- [Spec doc](https://paper.dropbox.com/doc/Spec-List-View-H9rEIGV717YlfL8JFCxoW)

cen10:
> Any possibility of using the solid shadow on the folder headings when the page is scrolled?

As of now, the folder header isn't actually part of the table component and I've added that in separately. The scrolling mechanism is also outside of the table component as well. So I should be able to add this in myself with the mailbox rewrite work.

src/components/Table/Table.jsx
```diff
@@ -37,13 +40,22 @@ export function Table({
...
   tableWidth = { min: '700px' },
+  withColumnChooser = false,
   withFocusableRows = false,
   withSelectableRows = false,
   withTallRows = false,
   ...rest
 }) {
-  const [state, actions] = useTable(data, maxRowsToDisplay)
+  const defaultColumns = columns.map(col => {
+    if (col.show == null) {
```
cen10, line 52: Is the `==` intentional or can this be `===`?

src/components/Table/Table.jsx
```diff
@@ -72,13 +86,27 @@ export function Table({
         dataCy={dataCy}
         {...rest}
       >
+        {withColumnChooser ? (
+          <HeaderUI>
```
cen10, line 90: For MVP we won't have the column chooser but we will still have a header with the View name. [relevant spec](https://paper.dropbox.com/doc/Spec-List-View--BRCgjs8z4tVOOsjbO2y_dKhDAg-H9rEIGV717YlfL8JFCxoW#:uid=700259760955942417510932&h2=Folder-Header).
 Can we tweak this so that whether we use columns or not, we can see the view name in the Header? Feels like we'd need to check for the presence of `withHeader` first and then also `withColumnChooser`, and in MVP only `withHeader` would be true.

If it's useful, in my current branch I made the view name an `<H1/>` with the following css (may need to be modified in your branch)
```
  h1 {
    height: 60px;
    font-size: 18px;
    font-weight: 400;
    line-height: 23.04px;
    padding-top: 16px;
    padding-bottom: 19px;
    padding-left: 15px;
    color: ${getColor('charcoal.800')};
  }
```

src/components/DropList/DropList.ListItem.jsx
```diff
@@ -90,7 +94,7 @@ const ListItem = forwardRef(
         <ListItemTextUI>
```
cen10, line 94: This is perhaps opening a can of worms and I don't think it's a blocker here - but it occurs to me that if we're essentially just providing a list of `<span>s` to Tippy to render in a dropdown, are we creating any accessibility concerns? How does the user know that these are actually more like buttons that can be toggled on and off?

src/components/Table/Table.jsx
cen10, line 90: Is it possible to tweak this so that it can handle the MVP goal of having a header that shows the view name but not column sorting? [Relevant spec.](https://paper.dropbox.com/doc/Spec-List-View--BRCgjs8z4tVOOsjbO2y_dKhDAg-H9rEIGV717YlfL8JFCxoW#:uid=700259760955942417510932&h2=Folder-Header) Feels like we'd need to check for `withHeader` and `withColumnChooser` but for MVP only `withHeader` would be true.

If it's helpful, in my current branch I use an `H1` with the following CSS:
```
  h1 {
    height: 60px;
    font-size: 18px;
    font-weight: 400;
    line-height: 23.04px;
    padding-top: 16px;
    padding-bottom: 19px;
    padding-left: 15px;
    color: ${getColor('charcoal.800')};
  }
```

cen10 (commented):
Looking really good - I had some questions and the big ask is to allow for the view name to show up in the Header so I can use it for MVP, but I could also just continue using what I've got until we need to introduce column sorting. The work would need to get done at some point before we need column sorting but it could be put off for a bit.

Also had a question or two for my own learning and a couple non-blocking suggestions.

cen10:
@digitaldesigner I think we discussed this at some point but just want to confirm that the behavior in Storybook is correct - when someone clicks on a column in the droplist, the droplist closes. If they want to add another column, they have to go through the process of opening the droplist again.

digitaldesigner:
@cen10 Correct, that interaction tested the best with users. It's made with the following assumptions:

1. People aren't really all that likely to be turning on/off loads of columns, so it's optimized for more of a one-off interaction
2. During testing, columns that were added which appeared beneath the dropdown or were quite small went without notice, and users felt like something had broken

So yeah, it's kinda jarring — but closing the dropdown when a selection is made is entirely by-design. Ideally we'd also add a noty to confirm the change, [as per this prototype](https://cdpn.io/team/helpscout/debug/d616f139a938ba69ae5d355c877e3f57)

src/components/DropList/DropList.jsx
```diff
@@ -180,6 +187,8 @@ function DropListManager({
               item: itemToRemove,
               key: contentKey,
             })
+          } else if (isItemReset(selectedItem)) {
+            updatedSelection = selectedItems
```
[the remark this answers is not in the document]
tinkertrain: Good question! Let me see if I can answer somewhat clearly hehe

First it's useful to separate the DropList behaviour from the Table Chooser one.

DropList adds an item at the bottom of the list when `withResetSelectionItem` is added (like you I also don't love the naming but the reason I used it was to avoid adding a boolean + a string props, perhaps there is still a better name, specially because this item doesn't necessarily mean it adds any resetting functionality, or any other functionality for that matter).

An item gets appended that when clicked doesn't do _anything_ (but this doing nothing is what makes it special, because it behaves just like any other item except for being selected), and is of type `reset_droplist`. It's up to you what to do here, it's very likely you'll either need to "reset" (like in the column chooser case) or "clear selection" (can't really think of another use case and that is why "reset" didn't sound too bad in the end).

This leaves us with implementing the actual resetting on the column chooser, in `Table.ColumnChooser.jsx:25` the relevant bit is (modified for clarity here):

```jsx
<DropList
  onSelect={(selection, clickedItem) => {
    if (clickedItem.type === 'reset_droplist') {
        setDropListSelection(defaultColumns) 
         // In our case we are resetting, but you could do something like:
         // setDropListSelection(null)
      }
    }
  }
  selection={state.dropListSelection}
/>
```

> One thing to consider is that this means your DropList selection needs to be _controlled_ by using `selection` 

I hold on to a reference to the original shape of the columns here `Table.jsx:51` and use that to restore the columns. This reference is really only used for that and for setting the initial state, the columns we use for rendering are stored in state (`columns={state.columns}`).

Hope this helps!

src/components/Table/Table.ColumnChooser.jsx
```diff
@@ -0,0 +1,64 @@
...
+      toggler={
+        <IconBtn
+          a11yLabel="Choose columns to show or hide"
+          iconName="column-check"
+          iconSize={24}
+          withCaret={false}
+        />
+      }
+      withMultipleSelection
+      withResetSelectionItem={columnChooserResetLabel}
```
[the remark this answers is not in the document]
tinkertrain: I agree... I mention the reasons in the previous comment (and accept suggestions). I just really dislike adding a gazillion amount of props.

src/components/Table/Table.jsx
tinkertrain: intentional, `== null` is a way to catch `null` or `undefined`, little side note, you'll notice eslint doesn't complain about it, whereas it will in any other case.

src/components/Table/Table.jsx
tinkertrain: I added a `headerContent` prop, you can pass any valid element and it will render it on the left inside the `<header>` element of the table (along with column chooser on the left, if enabled). I added your h1 styles as an example on the ConvoList demo

src/components/DropList/DropList.ListItem.jsx
tinkertrain: all elements have the appropriate aria roles (and aria selected attr too) which downshift takes care of updating, also an element (a11y-status-message) that gets read by a SR in the page gets updated, for example: https://i.hlp.sc/9sVqj9

src/components/Table/Table.jsx
tinkertrain: Done (use `headerContent` and/or `withColumnChooser`)

src/components/DropList/DropList.ListItem.jsx
tinkertrain: I did have a little out of sync thing in that department that I just fixed... so thanks for bringing this up!

tinkertrain:
@digitaldesigner Some of this stuff belongs in HS App where the Convo List will be implemented:


- (Hs App) Font sizes to be 13px except for customer name (stay 14px)
- (Hs App) Font color depends on convo status
    - (Hs App) Active row: charcoal.600 for everything except convo preview (stay charcoal.400)
    - (Hs App) Assigned/pending row: charcoal.500 for dark text
    - (Hs App) Closed/spam row: charcoal.400
- (Hs App) Remove row hover
    - Can be overriden using styled components
- (HSDS) Selected rows to have a blue.100 highlight (instead of yellow)
  ❓ Are we ok just having the blue line on the left as the sole distinction of a focused row? https://i.hlp.sc/I8JpMY
- [x] (HSDS) Make the tab focus border (left blue line) 2px width
- () Any possibility of using the solid shadow on the folder headings when the page is scrolled? See demo and Figma documentation
  ❓ Is the header always going to be fixed?

src/components/Table/Table.hooks.js
```diff
@@ -24,6 +26,7 @@ export function useTable(data = [], maxRowsToDisplay = null) {
   const [state, dispatch] = useReducer(reducer, {
     selectedRows: [],
     currentTableData: initialDisplayTableData,
+    columns: [...columns],
```
plbabin, line 29: curious about why we need to create a new array instance?

plbabin (approved):
looks good 🎉🎉🎉, I did add a comment/question but nothing major

src/components/DropList/DropList.jsx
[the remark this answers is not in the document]
cen10: Thank you!

src/components/Table/Table.ColumnChooser.jsx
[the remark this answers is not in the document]
cen10: It can definitely be left as is, there's a reason we have PropTypes and people will figure it out like I did. The only alternative I could think of is leaving off the `with`, so it's just `resetSelectionItem`.

src/components/Table/Table.jsx
cen10: TIL!

cen10 (approved):
👍🏻 👍🏻

src/components/Table/Table.hooks.js
tinkertrain: I had issues at some point with 😈 mutation 😈 on the `columns` (when toggling the "show" flag, but I need the original columns untouched for the "reset to defaults" option)

digitaldesigner:
>Are we ok just having the blue line on the left and the checkbox as the sole distinctions of a focused row?

@tinkertrain Yeah I think we're going to have to leave the background alone as there's too much meaning already tied up there. So the single blue line will have to be the only focus distinction

tinkertrain:
@digitaldesigner Thanks! Last thing is about the scroll question:

- [ ] (HSDS) Any possibility of using the solid shadow on the folder headings when the page is scrolled? See demo and Figma documentation
	❓ We have a "fading" thing when scrolling the table horizontally... should it stay like that? be "solid"? remove the horizontal scroll effect?

But I think that is out of scope for this particular PR anyways, let's put in a different ticket/PR. If the rest look OK to you do you mind approving the PR then I can release it.

digitaldesigner:
Yeah the soft fade (horizontal) is fine to keep, there's documentation around that (see below). One change that'd be awesome to make is to the selected rows (blue):

- Hover state, keep them blue (currently goes gray)
- The soft fade is transparent -> gray (should be transparent -> blue)
- Make all the column headers and row contents font-size 13px, except for the Customer column

digitaldesigner (approved):
Thorough work, as always

PR #964, opened 2021-08-19
HSDS-198 New KeyboardBadge component

plbabin:
# Feature
This update adds a new component: KeyboardBadge. This simple component represent a keyboard shortcut. It is intent to be used in Tooltip for now, but creating a standalone component will allow us to use it anywhere.

CleanShot 2021-08-19 at 16 15 52@2x

image

To make it even better, the component will look for the `mod` keyword, and will replace it with `⌘` on mac & `Ctrl` on other OS. It will also format a bit the shortcut to make it look good (ie: adding space around the `+` sign).


tinkertrain (commented):
💯

digitaldesigner:
Looking really close, couple of very minor tweaks — mainly to account for the fact that this version of the tip is a little larger in a few aspects than the default:

- Increase font-size to `13px` (currently `12px`)
- Increase border-radius to `4px` (currently `3px`)
- Padding-left to `10px` (currently `8px`)
- Padding-right to `6px` (currently `8px`)
- Background-color to `charcoal.800` (currently `charcoal.700`)
- Any chance of adjusting the line-height so the text sits vertically centered? Currently sitting a little high

plbabin:
> Looking really close, couple of very minor tweaks — mainly to account for the fact that this version of the tip is a little larger in a few aspects than the default:
> 
> * Increase font-size to `13px` (currently `12px`)
> * Increase border-radius to `4px` (currently `3px`)
> * Padding-left to `10px` (currently `8px`)
> * Padding-right to `6px` (currently `8px`)
> * Background-color to `charcoal.800` (currently `charcoal.700`)
> * Any chance of adjusting the line-height so the text sits vertically centered? Currently sitting a little high

All done. I did add a prop to the Tooltip component to support all that different style. I wasn't thrilled with modify the UI of the tooltip with vs without a badge, but so be it for now.

image

Vertical alignment should be better too, as you can see it can take multi line easily if needed

tinkertrain:
@plbabin After pulling my hair wondering why the hell the tippy wasn't created or anything...

THIS is the culprit:

```js
jest.mock('@tippyjs/react/headless', () => {
  const Tippy = ({ children }) => <div>{children}</div>
  return Tippy
})
```

Remove that mock... all tests pass (there's only a typo 👇 )

```
expect(getByText('alt + a')).toBeTruthy()

//alt should be capitalized
expect(getByText('Alt + a')).toBeTruthy()
```

tinkertrain:
@plbabin First I tried to unmock just for that test... unsuccessfully (seems like [jest issue/quirk](https://github.com/facebook/jest/issues/2649)), then I just removed the mock expecting some files to fail, but no, it all passes 👍

src/components/Tooltip/Tooltip.test.js
```diff
@@ -1,5 +1,7 @@
 import React from 'react'
 import { mount } from 'enzyme'
+import { render, waitFor, screen, fireEvent } from '@testing-library/react'
```
tinkertrain, line 3: One more thing, for consistency with other tests use `import user from '@testing-library/user-event'` instead of `fireEvent`, then `user.click(YOUR_ELEMENT)`

plbabin:
nice catch @tinkertrain! my latest commit fixed the test and use `user-events`. I think this PR is good to go!

tinkertrain (approved):
🚢

PR #965, opened 2021-08-24
Adds SimpleModal export

tinkertrain:
Forgot to export the SimpleModal in index.js main export

Also, adds `mainActionNode` prop to the `HeaderAndFooter` layout to retrieve the DOM node of the button


jakubjanczyk (approved):
:100: Thanks for making those changes!

PR #967, opened 2021-08-25
[Feature] AvatarList: Adds vertical stack option + Tooltip

tinkertrain:
CleanShot 2021-08-25 at 16 37 07

## Avatar ssmd size

The spec has these Avatars being 38px, but the closest we have is 35px for "smmd". This PR updates to 38px for consistency but we need design to OK this, if not either make them all 35px or add yet another size to the list:

```js
// all sizes
{
    xl: {
      size: 64,
      fontSize: 13,
    },
    lg: {
      size: 52,
      fontSize: 13,
    },
    md: {
      size: 46,
      fontSize: 13,
    },
    smmd: {
      size: 38, // <-- This was 35
      fontSize: 15,
    },
    sm: {
      size: 30,
      fontSize: 10,
    },
    xs: {
      size: 28,
      fontSize: 10,
    },
    xxs: {
      size: 26,
      fontSize: 10,
    },
  },
```

## Vertical stack

AvatarList has an option to stack, using the `stack` prop, this prop used to be a boolean but I searched in our projects and is not being used at all, so I changed it to be either 'horizontal' or 'vertical'. Horizontal behaves just like it used to if `stack === true`.

## Tooltip on the "extra" Avatar

When we limit the number of avatars displayed with `max`, we show an "extra" avatar that shows the count of the hidden avatars. Here we add a prop: `extraTooltipProps`, pass this and this Extra Avatar will be wrapped with a Tooltip, use all valid Tooltip props, example:

```jsx
<AvatarList
  max={3}
  center
  stack="vertical"
  size="smmd"
  extraTooltipProps={{
    title: ['John', 'Paul', 'Ringo'].join('\n'),
    placement: 'bottom',
  }}
>
  <Avatar borderColor="#f9fafa" image={image} name={name} />
  <Avatar borderColor="#f9fafa" image={image} name={name} />
  <Avatar borderColor="#f9fafa" image={image} name={name} />
  <Avatar borderColor="#f9fafa" image={image} name={name} />
</AvatarList>
```

### Tooltip on Avatar

The above is possible by the addition of `tooltipProps` on the Avatar:

```jsx
  <Avatar
	borderColor="#f9fafa"
	image={image}
 	name={name}
	tooltipProps={{
    	title: ['John', 'Paul', 'Ringo'].join('\n'),
	    placement: 'bottom',
    }}
  />
```

### DIFF
--- a/src/components/AvatarList/AvatarList.jsx
+++ b/src/components/AvatarList/AvatarList.jsx
@@ -1,8 +1,8 @@
 import React from 'react'
 import { PropTypes } from 'prop-types'
 import getValidProps from '@helpscout/react-utils/dist/getValidProps'
-import Avatar from '../Avatar'
 import Animate from '../Animate'
+import Avatar from '../Avatar'
 import { ItemUI, AvatarListUI, AvatarListWrapperUI } from './AvatarList.css'
 import classNames from 'classnames'
 import { getComponentKey } from '../../utilities/component'
@@ -22,13 +22,18 @@ const wrapAvatar = (props, avatar, index) => {
 
   let zIndex = max - index
 
-  if (stack && count > 2 && isOdd(`${count}`)) {
-    if (isOdd(index)) {
-      zIndex = zIndex + 1
+  if (stack === 'horizontal') {
+    if (count > 2 && isOdd(count)) {
+      if (isOdd(index)) {
+        zIndex = zIndex + 1
+      }
+
+      if (index === getMiddleIndex(count)) {
+        zIndex = zIndex + 2
+      }
     }
-    if (index === getMiddleIndex(`${count}`)) {
-      zIndex = zIndex + 2
-    }
+  } else if (stack === 'vertical') {
+    zIndex = index + 1
   }
 
   return (
@@ -50,7 +55,7 @@ const getCurrentCount = ({ count, max }) => (count < max ? count : max)
 export const getAvatarSize = ({ size: sizeProp, stack, ...rest }) => {
   const currentCount = getCurrentCount(rest)
 
-  if (!stack) return sizeProp
+  if (stack !== 'horizontal') return sizeProp
 
   let size = 'md'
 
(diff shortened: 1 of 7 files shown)


src/components/AvatarList/AvatarList.jsx
```diff
@@ -65,7 +70,16 @@ export const getAvatarSize = ({ size: sizeProp, stack, ...rest }) => {
...
+    children,
+    max,
+    className,
+    center,
+    stack,
+    grid,
+    extraTooltipProps,
+    ...rest
+  } = props
   const avatars = React.Children.toArray(children)
```
luketlancaster, line 83: So this is something that Terrence and I talked a little about: it feels a bit odd that we're passing in children to this component, versus some kind of `renderComponent` or `renderChild` (it'd be `childView` or something in marionette land...) prop that tells it "hey, I want all the items in the collection I pass to be rendered as this component." I think I've seen that pattern elsewhere in HSDS (though maybe I'm just imagining it). Mostly I bring this up because we're `toArray`-ing here anyway, so wondering if it doesn't make sense to skip the middle-man? I don't mind one way or the other, just curious mostly

luketlancaster (approved):
This looks good to me. Seems like adding in the `extraTootipProps` will be pretty straightforward, so I'm down with that. Had one question about the architecture of passing children (instead of a collection prop), but that's more for conversation/thought than anything we need (or want) to change.

I shared a link with @nikkiathelpscout, so she can take a look at the sizing question and provide feedback there

src/components/AvatarList/AvatarList.jsx
tinkertrain: This is a bit of an old component and don't remember all of the decisions that went into it,  I don't quite follow what you mean though, are you suggesting that instead of something like:

```jsx
<AvatarList>
	<Avatar />
	<Avatar />
	<Avatar />
</AvatarList>
```

We should have...

tinkertrain: Ahh I see, passing children would make sense _if_ you could put whatever you wanted there (which for example would allow you to wrap the avatars with a Tooltip right there), but really all you can pass are avatars. Changing it doesn't really gives many benefits and as awkward as the API currently is, it works so will leave it for now and make sure we don't do this kind of thing anymore.

src/components/AvatarList/AvatarList.jsx
```diff
@@ -50,7 +55,7 @@ const getCurrentCount = ({ count, max }) => (count < max ? count : max)
 export const getAvatarSize = ({ size: sizeProp, stack, ...rest }) => {
   const currentCount = getCurrentCount(rest)
 
-  if (!stack) return sizeProp
+  if (stack !== 'horizontal') return sizeProp
```
TerrenceLJones, line 58: @tinkertrain could we add support for consuming the `sizeProp` no matter the stack setting?

PR #968, opened 2021-08-30
[Fix] RadioCard styles

tinkertrain:
# Problem

@alexestrada found:

> When reviewing some of the work for Standalone Messages, I noticed the radio card text doesn't seem to change when selected or in their focus state. It'd be great if we can get this fixed before Standalone launches! 🙏 If not, let's work together to determine a timeline. 
Below you'll find screenshots + the Storybook link as well as the Figma spec for reference. 

https://helpscout.atlassian.net/browse/HSDS-202

## Solution
Updates color of text when RadioCard is checked to charcoal.600

[Story](https://deploy-preview-968--hsds-react.netlify.app/?path=/story/components-forms-radiocard--with-illos)

CleanShot 2021-08-30 at 12 19 22@2x

This change also affect the color of the text when there is a heading:

CleanShot 2021-08-30 at 12 41 12@2x

Previously:
CleanShot 2021-08-30 at 12 43 15@2x

I think it's OK but let us know if we only want to change the case above.


jakubjanczyk (approved):
Looks good!

alexestrada:
@tinkertrain thanks for this update! I am noticing the text weight needs to be Aktiv Medium upon selection. The color is correct.

Appreciate the callout on the RadioCard with the heading, but let's keep this just to the RadioCard w/ illos when going from unselected to selected.

tinkertrain:
@alexestrada Thanks for looking at it!

I made the updates, I also noticed the initial state when nothing is selected yet was wrong compared to the Figma specs:

You can see they all look dimmed (which is the state the unselected cards get when there is a selection):
CleanShot 2021-09-01 at 12 24 04

This is what it looks like now:
CleanShot 2021-09-01 at 12 24 53

PR #969, opened 2021-08-30
[Fix] EditableTextArea: updates validation icon position

tinkertrain:
# Problem

The validation icon is positioned too close to the border bottom.

CleanShot 2021-08-30 at 15 00 25@2x

## Solution

Update the `position`, instead of `top` use `bottom`:

One line:

CleanShot 2021-08-30 at 14 48 08@2x

Multiple lines:

CleanShot 2021-08-30 at 14 55 19@2x


luketlancaster (approved):
👏🏻

PR #970, opened 2021-08-30
[Feature] Modal: Adds focusModalOnShow prop 

tinkertrain:
# Problem

Often we want to focus something inside the modal as it renders (like the first input). Using `onOpen` should allow you to do this when creating your Modal, but currently as soon as the Modal mounts it focuses the content, so effectively when focusing your own thing its focus state immediately gets taken by the Modal.

## Solution

Like in SimpleModal, add `focusModalOnShow` prop to allow the user to remove this default behaviour in case they want to focus something themselves.

Example:

```jsx
export function MyModal() {
  const inputRef = React.useRef(null)

  return (
    <Modal
      focusModalOnShow={false}
      onOpen={() => {
        inputRef.current.focus()
      }}
      trigger={<button>Clicky</button>}
      version={2}
    >
      <Modal.Body version={2}>
        <input type="text" ref={inputRef} />
      </Modal.Body>
    </Modal>
  )
}
```

https://i.hlp.sc/kauzj2


src/components/Modal/Modal.jsx
```diff
@@ -440,6 +446,8 @@ Modal.propTypes = {
   description: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
   /** Used with `path` and React Router. Renders if path matches _exactly_ */
   exact: PropTypes.bool,
+  /** If you don't want the focus to be moved to the Modal when it enters */
+  focusModalOnShow: PropTypes.bool,
```
Charca, line 450: Looks like this prop doesn't have a default value, so I'm guessing it's defaulting to `undefined`, right? Should this default to `true` in order to maintain backward compatibility?

Charca (commented):
Thanks for adding this, @tinkertrain! I just left a comment.

src/components/Modal/Modal.jsx
tinkertrain: Yes! I forgot to add it thanks 👍

Charca (approved):

PR #966, opened 2021-08-25
Removes deprecated components

tinkertrain:
🪓 SplitButton
🪓 SelectDropdown
🪓 SearchableDropdown

➕ Add deprecated comments on Dropdown, remove it's tests from runnning and stories from being shown
➕ Remove duplicated stories of ActionSelect

Companion HSApp PR https://github.com/helpscout/hs-app/pull/10347

PR #974, opened 2021-09-02
Change SplitButton focus ring color

plbabin:
# Problem
When the SplitButton is a primary one, the focus ring blended with the button. It looks weird
CleanShot 2021-09-02 at 13 56 49@2x

# Solution
In that situation, we bump the ring color to "blue.600". We also add an inset white shadow to increase the floating effect
CleanShot 2021-09-02 at 13 58 18@2x

### DIFF
--- a/src/components/SplitButton/SplitButton.css.js
+++ b/src/components/SplitButton/SplitButton.css.js
@@ -1,6 +1,17 @@
 import styled from 'styled-components'
 import Button from '../Button'
+import { FocusUI } from '../Button/Button.css'
 import { getColor } from '../../styles/utilities/color'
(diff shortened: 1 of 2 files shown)


digitaldesigner (approved):

tinkertrain:
SplitButton is no longer used https://github.com/helpscout/hsds-react/pull/966, these changes might need to be applied there instead

Got replaced by using DropList + SplittedButton toggler

plbabin:
> SplitButton is no longer used #966
> 
> Got replaced by using DropList + SplittedButton toggler, these changes might need to be applied there instead

nice catch, I transferred the styles to the DropList trigger in 8e7885c

PR #975, opened 2021-09-02
New Tag UI

plbabin:
# New color
CleanShot 2021-09-02 at 16 16 34@2x
CleanShot 2021-09-02 at 16 17 09@2x

This update change the color for all tag. Exceptionally we are getting out of the HSDS color palette to have the best accessible color possible. Since the text is small it needs to be readable and color needs to be out of the way. 

All tags have the same text color (charcoal.600). We have brand new colors (`teal` / `pink`). Sadly, we had to part ways with `lightBlue`. Thank you for your service

# Size
We now support three different sizes (`sm`, `md`, 'lg'). `sm` has a 18px height, `md ` is 22px and 'l

# Clickable
Tag can now be clickable in the right context. If a Tag contain either a `onClick` prop or an href, the tag will have a pointer cursor. Only a href will act as a simple `<a>` while an onClick will make the tag a `<button>`

CleanShot 2021-09-02 at 16 21 20

Being clickable means that each tag needs a hover state. To keep it simple, we did darken the main color by 30%.

# Count
We add a `count` prop that will be visible only when the size is `md`. It will render a badge after the tag text with the count prop value

CleanShot 2021-09-02 at 16 19 47@2x

# `isRemovable`
We still support the is removable prop. However, we did refactored how it is working internally. Now it relies on the `onTransitionEnd` event to unmount the component. We remove the onBeforeRemove promise implementation (it was not used anywhere). Now there is only a `onRemove` callback.

We did add a focus state to the remove icon to make it more accessible. If the tag is clickable, the tab order will still make sense when moving through the focusable elements

CleanShot 2021-09-02 at 16 26 51

# Limit
We now support limiting the number of tags visible in `<TagList>`. We will show a badge at the end to let the user display all the tag in the list:

CleanShot 2021-09-03 at 16 15 41

With this addition, we deprecated the `showAll` prop. Now all tags are visible by default. To limit the size of the element, we can either use the new `limit` prop or the `overflowFade`. The latter will keep all tags on a single row, with a fade effect at the end of the element.

# RTL
Lastly, we did refactored all Tag & TagList test to use react-testing-library 🎉


digitaldesigner (commented):
@jaredmcdaniel is taking point on this — so take this as guidance only... but my gut is that we need a little X icon that works well at the small size (it's getting lost in there). Also I reckon `font-weight:500` might be necessary

tinkertrain (approved):
Looks pretty good to me!

jaredmcdaniel:
@plbabin right off the bat, agree with Buzz—let's start with these few CANIs. Going to spend some more time on the interactions and will circle back.

### Typography

```
font-weight: 500;
font-size: 11.5;
letter-spacing: -0.1px;
line-height: 12px;
```

### isRemovable versions
Lets update the **X** icon styles to:

```
width: 18px;
height: 18px;
margin-top: -9px;
opacity: 0.6; DONT LOOK AT ME
```

For `sm` size, remove the `right` positioning

### Question
Will we be implementing the `lg` versions in this PR as well?

plbabin:
Latest update includes the `lg` size variant, the limit prop and all CANIs. 

@tinkertrain I also created a focusRing mixing so we could bring that blue border on any element quickly https://github.com/helpscout/hsds-react/pull/975/commits/af1078df076966de511be55649aacab5d22f3f5d

plbabin:
@jaredmcdaniel do you mind have another look at this PR? I would like to merge it soon so I can update tags in both hsds/hs-app all together

jaredmcdaniel:
@plbabin looking great. My last requests here pertain mostly to the recently added `lg` version:

1. `is-lg` version should have a border-radius of 4px (vs 3)
2. `is-lg` version should have 13px font-size (vs 11.5)
3. `is-lg` version: counts should be 12px (vs 11.5)
4. Lastly, let's update the default grey outline version (all sizes) to use `grey-500` (currently using `grey-400`)

[Figma file](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=9648%3A0) if helpful.

plbabin:
> @plbabin looking great. My last requests here pertain mostly to the recently added `lg` version:
> 
> 1. `is-lg` version should have a border-radius of 4px (vs 3)
> 2. `is-lg` version should have 13px font-size (vs 11.5)
> 3. `is-lg` version: counts should be 12px (vs 11.5)
> 4. Lastly, let's update the default grey outline version (all sizes) to use `grey-500` (currently using `grey-400`)
> 
> [Figma file](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=9648%3A0) if helpful.

you got it! 

CleanShot 2021-11-10 at 09 02 36@2x

jaredmcdaniel:
@plbabin tell me if this adds too much complexity, but the hope would be for grey-500 to apply **_only_ for the outline versions**. The filled version would still use grey-400 as it did before.

Contrast still checks out so I wont die on this hill, but wanted to ask. 😄

plbabin:
> @plbabin tell me if this adds too much complexity, but the hope would be for grey-500 to apply **_only_ for the outline versions**. The filled version would still use grey-400 as it did before.
> 
> Contrast still checks out so I wont die on this hill, but wanted to ask. 😄

that wasn't too bad, how does it feel now?

CleanShot 2021-11-10 at 10 10 48@2x
CleanShot 2021-11-10 at 10 10 41@2x

jaredmcdaniel (approved):

PR #976, opened 2021-09-07
[HSDS-205] DropList features

tinkertrain:
# Feature

https://helpscout.atlassian.net/browse/HSDS-205

To enable this [workflow](https://helpscout.atlassian.net/browse/HSAPPUI-137):

Screen Shot 2021-09-07 at 2 21 56 PM

When there are no items either on the initial mount of the DropList or when in combobox mode we filter down to 0 results.

➡️

We want to show "extra" items, that enables us to perform some action

## customEmptyListItems
For this, we add the new prop `customEmptyListItems` which accepts an array of items, any item that is accepted by DropList is valid here.

> `customEmptyList` still exists, this customizes the rendering of an empty list, use this if what you need is to display something static

```jsx
customEmptyListItems={[
  {
    label: 'No tags found',
    type: 'inert',
  },
  {
    type: 'divider',
  },
  {
    label: 'Create tag',
    type: 'action',
  },
]}
```

### Combobox feature

On a combobox, you can customize the label of your items with the value of the input by adding a `customizeLabel` function on the item

```jsx
customEmptyListItems={[
  {
    label: 'No tags found',
    type: 'inert',
  },
  {
    type: 'divider',
  },
  {
    label: 'Create tag',
    customizeLabel(inputValue) {
      return `Create ${inputValue} tag`
    },
    type: 'action',
  },
]}
```

## New item type: 'action'

Some times you need an item in the list that is not "selectable" but "actionable", add a type of `action`. To reiterate, these items:

1. Fire `onSelect`
2. Are interacted with like any other DropList item (ie with keyboard)
3. Do not get added to the selection, so `onSelect` will fire like `onSelect(null, clickedItem)`

## New item type: 'inert'

If you need an item that is just there and really doesn't do anything (like the "No tags Found" in the screenshot above), add an item with a type of `inert`. Technically similar to disabling an item, but with different semantics.

## Styles

Several class names are added to the items so you can target and style yourself if you need other than the default list item styles.

### Variant select, no items

CleanShot 2021-09-07 at 14 16 10

### Variant Combobox, no items on mount

CleanShot 2021-09-07 at 14 17 31

### Variant Combobox, filtered down to no items

CleanShot 2021-09-07 at 14 16 50

## Reset item

With the addition of `action` items, there is no need for the reset functionality previously added as this can easily be implemented with these. The Table Column Chooser has been updated to use this workflow instead and the reset options and implementation have been removed from DropList.

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -90,8 +90,10 @@ function Combobox({
 
           return item
         })
+
         filtered = filtered.concat(processed)
       }
+
       setHighlightedIndex(filtered.findIndex(isItemHighlightable))
       setInputItems(filtered)
     },
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -91,7 +91,7 @@ function Combobox({
           return item
         })
 
-        filtered = filtered.concat(processed)
+        filtered = processed
       }
 
       setHighlightedIndex(filtered.findIndex(isItemHighlightable))
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -2,10 +2,12 @@ import React, { useState, useRef, useEffect } from 'react'
 import { useCombobox } from 'downshift'
 import useDeepCompareEffect from 'use-deep-compare-effect'
 import { noop } from '../../utilities/other'
+import { isFunction } from '../../utilities/is'
 import {
   itemToString,
   isItemSelected,
   renderListContents,
+  isItemHighlightable,
 } from './DropList.utils'
 import {
   getA11ySelectionMessageCommon,
@@ -25,6 +27,7 @@ function Combobox({
   closeOnBlur = true,
   closeOnSelection = true,
   customEmptyList = null,
+  customEmptyListItems,
   'data-cy': dataCy = `DropList.${VARIANTS.COMBOBOX}`,
   handleSelectedItemChange = noop,
   inputPlaceholder = 'Search',
@@ -41,6 +44,11 @@ function Combobox({
   withMultipleSelection = false,
 }) {
   const [inputItems, setInputItems] = useState(items)
+  const isListEmpty = items.length === 0
+  const allItems =
+    isListEmpty && Array.isArray(customEmptyListItems)
+      ? customEmptyListItems
+      : inputItems
   const inputEl = useRef(null)
 
   const {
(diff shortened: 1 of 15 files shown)


tinkertrain:
@cen10 you might be interested in this, technically doesn't affect you as the column chooser works the same from the outside but you might be glad to hear that the `withResetSelectionItem` prop is gone and no more weird string/boolean types mixed.

@jakubjanczyk I added you because you have played extensively with DropList and reviewed several PRs and thought you'd be a great addition to catch if I broke something 😅

src/components/DropList/DropList.Combobox.jsx
```diff
@@ -67,11 +76,24 @@ function Combobox({
...
+      if (isListEmpty && Array.isArray(customEmptyListItems)) {
+        const processed = customEmptyListItems.map(item => {
+          if (isFunction(item.customizeLabel)) {
+            item.label = item.customizeLabel(inputValue)
+            item.inputValue = inputValue
+          }
+
+          return item
+        })
+        filtered = filtered.concat(processed)
```
jakubjanczyk, line 93: Do we need to make concat here? It seems to be calling this code only if `isListEmpty`, so we might as well just replace the list, to not have to think here why there's `concat` :smile:

jakubjanczyk (approved):
Thanks @tinkertrain, that looks great and this time I wasn't able to break it :smile: 
I think `action` might be useful for some cases in Messages - @Charca this might be handy for Dashboard/Report pages droplists.

As for this PR, I have noticed 2 things, but it might be as designed, so to be verified:
- While `inert` item cannot be selected, when you move mouse over it, there's a style change as if it was regular item
- On the first screenshot, the `create` action has different color than in storybook - would it make sense to make actions color customizable? I would assume it should be blue for tags, but for example in Messages usage, we would have regular color.

src/components/DropList/DropList.Combobox.jsx
tinkertrain: Ahh yeah, good catch!

tinkertrain:
@jakubjanczyk Thanks for looking at this!

1. I have removed the `inert` hover styles... thanks!
2. I wasn't trying to duplicate the "tags" styles on the screenshot, this will be up to you (the user) to style (I'm sure every instance is going to differ slightly, see this vs the column chooser droplist). Styling can easily be achieved in a couple of ways, by targeting the item class name `is-type-action` like I do on [the table column chooser's css](https://github.com/helpscout/hsds-react/pull/976/files#diff-ffdad83bbbfdb377dd43f525bd63a91fb722daea9d9443232188a8cfd60406e3R47), or by adding a class name directly on your item (see [laser targeting in this section](https://hsds-react.netlify.app/?path=/docs/components-dropdowns-droplist--default-story#list-items).

TerrenceLJones (approved):
LGTM! Thank you so much for quick turn around on this new functionality!

PR #978, opened 2021-09-10
DropList: onDropListLeave

tinkertrain:
# Feature

Adds a callback `onDropListLeave` that gets fired when:

- Blur ouf toggler only when the next focused element is not the DropList Menu
- Blur out of Menu (only when the next focused element is not the toggler)
- Click outside

Also tweaks some behaviours:

- If `focusTogglerOnMenuClose` is `false`:
  - Selecting an item still focuses the toggler if `closeOnSelect` is `true`
  - Pressing <kbd>esc</kbd> also focuses the toggler

Tests

- For some reason there was some duplicated tests... oops
- At the beginning the use of `waitFor` was needed everytime the menu was open/close, I think with some of the refactoring I did later on, this became unnecesary, so I cleaned up the tests.


jakubjanczyk (approved):
Looks great, thanks!

PR #979, opened 2021-09-14
[HSDS-206] New component: AvatarRow

tinkertrain:
[HSDS-206](https://helpscout.atlassian.net/browse/HSDS-206)

We have `AvatarList` which has several features like "stacking" avatars horizontally or vertically. We are not going to need any of those extra features anymore, instead we need a component that "adapts" to the available space and shows/hides avatars accordingly.

AvatarList also has a weird API and it's too complex for what it does.

Since refactoring and then replacing instances is more effort (plus we need design to OK every instance where stacking is being used for example), a new component instead was created, and eventually we'll say bye bye 👋 to AvatarList.

# AvatarRow

All you need to know is in the documentation, in a nutshell:
- Shows/hides avatars depending on size with one of two mechanisms:
  - **ResizeObserver**: the default, will resize always when the container resizes for any reason, but [is not supported in IE](https://caniuse.com/resizeobserver), related: [Help Scout's supported browsers](https://docs.helpscout.com/article/1292-supported-browsers-and-system-requirements)
  - **"resize" event** on `window`: Use `ieCompatible` to use this instead, the component will adapt only when the window is resized, but this is what you will want most of the time anyways.

See it: 📹 https://i.hlp.sc/EZPpiy
Play with it: [📚 story](https://deploy-preview-979--hsds-react.netlify.app/?path=/story/components-structural-avatarrow--default-story)

Example code:

```jsx
<AvatarRow
  avatars={[
    {
      id: '001',
      image: 'some_image_path',
      name: 'McNulty',
    },
    {
      id: '002',
      image: 'some_image_path',
      name: 'Bunk',
      tooltipProps: { title: 'Hey Bunk' },
    },
  ]}
  size="lg"
/>
```


digitaldesigner (changes requested):
This is incredible, such nice work @tinkertrain! Couple of super minor tweaks, since you're so close to perfection:

- On the overflow avatar, chuck a `font-weight: 500` on the number
- Number should be `charcoal.500` and not `400`

tinkertrain:
@digitaldesigner Thanks for looking! I have updated the styles

src/components/AvatarRow/AvatarRow.jsx
```diff
@@ -0,0 +1,168 @@
...
+import PropTypes from 'prop-types'
+import throttle from 'lodash.throttle'
+import classNames from 'classnames'
+import { noop } from '../../utilities/other'
+import { AvatarRowUI, CounterAvatarUI } from './AvatarRow.css'
+import Avatar from '../Avatar'
+import Tooltip from '../Tooltip'
+import { config } from '../Avatar/Avatar.css'
+import { setupObserver, splitAvatarsArray } from './AvatarRow.utils'
+import { number } from '@storybook/addon-knobs'
```
TerrenceLJones, line 11: Should this import be here?

src/components/AvatarRow/AvatarRow.jsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  function onResize({ width: containerWidth }) {
+    /** Only act if we have more than 1 avatar */
+    if (!containerWidth || avatars.length <= 1) {
+      return
+    }
+
+    const spaceForAllAvatars =
```
TerrenceLJones, line 96: Could you add a comment to this explaining this calculation? I'm having trouble understanding the `(avatars.length - 1)` part.

src/components/AvatarRow/AvatarRow.jsx
```diff
@@ -0,0 +1,168 @@
...
+    >
+      {shownAvatars.map(props => (
+        <Avatar key={props.id || props.name} {...props} {...avatarProps} />
+      ))}
+      {hiddenAvatars.length > 0 && (
+        <Tooltip
+          title={hiddenAvatars.map(props => props.name).join('\n')}
+          withTriggerWrapper={false}
+          placement="bottom"
+          {...extraTooltipProps}
```
TerrenceLJones, line 132: And just for clarification...you mentioned something similar in the the Storybook file, but with `extraTooltipProps` we could provide a custom `title` value? Say if we wanted to say that certain users are typing.

TerrenceLJones (commented):
This looks excellent! I just have a couple comments/questions.

src/components/AvatarRow/AvatarRow.jsx
tinkertrain: Ha, I don't even know how that got there... I think VSCode's was being super "helpful" auto importing it...

TerrenceLJones: Yep, that's what I figured too!

src/components/AvatarRow/AvatarRow.jsx
tinkertrain: No problem! Let me know if it makes it clearer

src/components/AvatarRow/AvatarRow.jsx
tinkertrain: Yes, anything on `extraTooltipProps` overrides what is here

.remake/component/remake-name/remake-name.jsx
```diff
@@ -1,49 +1,32 @@
...
+function <%= name %>({
+  className,
+  'data-cy': dataCy = '<%= name %>',
+  onClick = noop,
+}) {
+  const [] = useState()
+
+  return (
+    <<%= name %>UI
+      className={classNames('<%= name %>', className)}
```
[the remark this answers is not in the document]
tinkertrain: I guess we can, I don't follow it myself though 😳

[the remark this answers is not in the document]
TerrenceLJones: Haha...I'm open to whatever! Just wanted to check!

src/components/AvatarRow/AvatarRow.jsx
TerrenceLJones: Nice thanks!

TerrenceLJones (approved):
Thanks again for picking up and implementing this work so quickly! LGTM! 🚢

src/components/AvatarRow/AvatarRow.utils.js
```diff
@@ -0,0 +1,30 @@
...
+      : []
+
+  const shownAvatars = hiddenAvatars.length
+    ? avatars.slice(0, itemsDisplayed - 1)
+    : avatars
+
+  return { shownAvatars, hiddenAvatars }
+}
+
+export function setupObserver(callback) {
```
plbabin, line 16: is that the kind of utils we would like to make it more global so we could _potentially_ reused it in other component?

src/components/AvatarRow/AvatarRow.jsx
```diff
@@ -0,0 +1,172 @@
...
+      const numberOfGaps = numberOfItemsOnDisplay - 1
+      const itemsThatFit = Math.floor(
+        (containerWidth - numberOfGaps * gap) / avatarSize
+      )
+
+      setNumberOfItemsOnDisplay(itemsThatFit > 0 ? itemsThatFit : 1)
+    }
+  }
+
+  const { shownAvatars, hiddenAvatars } = splitAvatarsArray(
```
plbabin, line 115: I do wonder if we should wrap that with useMemo and update it only if `numberOfItemsOnDisplay` and/or the avatars value change (might be an issue if the parent doesn't memoize the avatar list). I just don't like having to split the avatars array on every call.

plbabin (approved):

digitaldesigner (approved):

src/components/AvatarRow/AvatarRow.utils.js
tinkertrain: Yes I think so! `ResizeObserver` is pretty cool and useful, I'm going to make a comment to move this to the long coming "utils" revamp that I hope I can tackle soon 😅

src/components/AvatarRow/AvatarRow.jsx
tinkertrain: In this case I didn't add it because the splitting of the avatars is only a couple of `slice` operations which I believe (but have not benchmarked) are very fast, so I'm not sure if the overhead of memoizing this is larger than the slicing.

src/components/AvatarRow/AvatarRow.utils.js
tinkertrain: (the only drawback is that is not that easy to test)

PR #980, opened 2021-09-16
Some tweaks on PopOver, Tooltip and DropList

tinkertrain:
## PopOver and Tooltip

https://helpscout.atlassian.net/browse/HSDS-211

Adds the `withArrow` (default `true`) prop to be able to render without it.

PopOver's styles get update to use depth.500

Screen Shot 2021-09-16 at 12 44 37 PM

Screen Shot 2021-09-16 at 12 49 26 PM

## DropList

https://helpscout.atlassian.net/browse/HSDS-212

The initial desire of not deviating too much from the original design of the DropList menu made us make the choice of making it a bit difficult to customize some things, including the width of the menu.

I've noticed almost everywhere that is used, this is the first thing that gets overridden... 🙈

So might as well add the prop 🤷‍♀️

I introduce you `menuWidth`

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -36,6 +36,7 @@ function Combobox({
   items = [],
   isOpen = false,
   menuCSS,
+  menuWidth,
   onDropListLeave = noop,
   onMenuBlur = noop,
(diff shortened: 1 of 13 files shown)


TerrenceLJones (approved):
Reviewed and tested! LGTM!

PR #981, opened 2021-09-20
COMMS-1435: Add ellipsis icon (which may or may not have food-themed name)

justinwolfe:
# Problem/Feature
[Jira](https://helpscout.atlassian.net/browse/COMMS-1435)

CleanShot 2021-09-20 at 14 09 46

For the overflow menu in the conversation view redesign, @digitaldesigner mentioned that we need to add a new version of the vertical ellipsis icon, so this PR is to do that!

Buzz noted that the designers have been calling this `Ellipsis Vertical Large`. I noticed that we call the current slimmer version of this `kebab`, so I've initially called it `kebab-large` so it'll show up next to the slimmer version in Storybook, but I figure we could:

- Just change it to `ellipsis-vertical-large`
- Include it under the alias `ellipsis-vertical-large` as well (maybe something similar going on with `ellipsis`, `meatball`, and `option-dots`)


plbabin:
@justinwolfe yeah not sure why it was named kebab and meatball originally (maybe from the alternate name in figma). Looking at the `Misc` section in figma and they all follow the ellipsis pattern. In hs-app there is only one `DropList` component using meatball and in hs-app-ui there is only `OverflowMenu` that is using kebab as the icon.

So I am confused on what to do 😕. I think like you suggest we could create aliases for the kebab icon

`ellipsis-vertical-large: kebabLarge`
`ellipsis-vertical: kebab` 

and  when  @tinkertrain come back from PTO see what we can do on that naming pattern. How does that sound?

justinwolfe:
@plbabin Thanks for that context—yep, sounds good to me! Just to make sure I'm understanding what I should do now, should I:
- Add those aliases now so we can merge/release
- Wait till next week so we can go through it with JP before making a decision

I'm assuming the latter (which is totally fine, this isn't a blocker) but wanted to double check!

plbabin:
@justinwolfe I am fine with adding the aliases and do a release

I would wait next wait to see if we ditch entirely meatball and kebab (we would need to update hs-app, hs-app-ui and some internal code within HSDS to make that happen)

justinwolfe:
@plbabin, Ahh, gotcha, awesome! Added the aliases ✅

tinkertrain:
I also don't know why they're named differently and have no opinion except that the meat options are way more fun

plbabin (approved):

PR #982, opened 2021-09-21
HSDS-215: Add onchange prop for DropList combobox input

justinwolfe:
# Problem/Feature
[Jira](https://helpscout.atlassian.net/browse/HSDS-215)
Per @luketlancaster's comment [here](https://github.com/helpscout/hs-app-ui/pull/110#pullrequestreview-759920710-permalink), for the design requirement of this card, I need to make the DropList work so that if:
-if there's no text entered into the search input, we don't render the dropdown at all, _just the input_
-if there's text in the search input and there are matches, we render the matches
-if there's text in the search input and there are no matches, we render the custom empty list items

ex

I don't think I can accomplish that first requirement with the current DropList implementation because there's not a way to access the search input value. In order to do that, I would think that what's needed is a prop for DropList that passes down a callback for onChange events for the search input. I could then use that to hide the dropdown (using CSS) where there's no text in the search input. 

That's what I've PRed simply here—does that sound correct or is there another way you'd recommend approaching this?

## Guidelines

Make sure the pull request:

- [x] Adds unit tests

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -41,6 +41,7 @@ function Combobox({
   onMenuBlur = noop,
   onMenuFocus = noop,
   onListItemSelectEvent = noop,
+  onComboboxInputChange = noop,
   renderCustomListItem = null,
   selectedItem = null,
   selectedItems,
@@ -201,6 +202,9 @@ function Combobox({
                 ['DropListToggler'],
                 onDropListLeave
               )
+            },
+            onChange: event => {
+              onComboboxInputChange(event.target.value)
             },
             onFocus: event => {
               onMenuFocus(event)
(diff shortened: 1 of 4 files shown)


justinwolfe:
(Thinking about it now, maybe another way to do it is to pass a prop to `DropList` like `hideListIfEmptyInput` or something and have the component handle whether or not to display the dropdown based on the input)

plbabin:
@justinwolfe yeah I would rather have a prop name `hideCustomListIfEmptyInput` instead of doing triggering a huge re-renderer by doing everything outside the DropList. That prop need to be implement only in the combobox variant. 

I think `renderListContents` could received that prop and handle not rendering the list (if hideCustomListIfEmptyInput is `true` and the `inputValue` is empty

justinwolfe:
@plbabin Thank you, that sounds good! I'll work on it 👍

justinwolfe:
Closing up—will approach differently in new PR

PR #983, opened 2021-09-22
HSDS-215: Add hideCustomListIfEmptyInput option to combobox DropList

justinwolfe:
# Problem/Feature

Re: @luketlancaster's comment [here](https://github.com/helpscout/hs-app-ui/pull/110#pullrequestreview-759920710-permalink), for the design requirement of this card, I need to make `DropList` work so that if:
-if there's no text entered into the search input, we don't render the dropdown at all, just the input
-if there's text in the search input and there are matches, we render the matches
-if there's text in the search input and there are no matches, we render the custom empty list items

The current `DropList` implementation doesn't provide the ability to meet requirement 1, so this adds a prop, `hideCustomListIfEmptyInput` to support that functionality. Pretty straightforward (thanks to helpful directions from @plbabin !). This is a second go at #982—per guidance in the comments [there](https://github.com/helpscout/hsds-react/pull/982#issuecomment-925246898).

There's a small CSS tweak that's added because `MenuListUI` usually has some extra padding on the bottom when the menu is rendered and that was still there without the items.

## Guidelines

Make sure the pull request:

- [x] Adds unit tests
- [x] Has it been tested in [Help Scout's supported browsers](https://docs.helpscout.com/article/1292-supported-browsers-and-system-requirements)?

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -233,7 +233,9 @@ function Combobox({
       <MenuListUI
         className={classNames(
           `${DROPLIST_MENULIST} MenuList-Combobox`,
-          hideCustomListIfEmptyInput && 'hideCustomListIfEmptyInput'
+          hideCustomListIfEmptyInput &&
+            !inputValue &&
+            'hideCustomListIfEmptyInput'
         )}
         {...getMenuProps()}
       >
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -22,6 +22,7 @@ import {
 } from './DropList.css'
 import ListItem, { generateListItemKey } from './DropList.ListItem'
 import { DROPLIST_MENULIST, VARIANTS } from './DropList.constants'
+import classNames from 'classnames'
 
 function Combobox({
   clearOnSelect = false,
@@ -32,6 +33,7 @@ function Combobox({
   'data-cy': dataCy = `DropList.${VARIANTS.COMBOBOX}`,
   focusToggler = noop,
   handleSelectedItemChange = noop,
+  hideCustomListIfEmptyInput = false,
   inputPlaceholder = 'Search',
   items = [],
   isOpen = false,
@@ -229,11 +231,15 @@ function Combobox({
         />
       </InputSearchHolderUI>
       <MenuListUI
-        className={`${DROPLIST_MENULIST} MenuList-Combobox`}
+        className={classNames(
+          `${DROPLIST_MENULIST} MenuList-Combobox`,
+          hideCustomListIfEmptyInput && 'hideCustomListIfEmptyInput'
+        )}
         {...getMenuProps()}
       >
         {renderListContents({
           customEmptyList,
+          hideCustomListIfEmptyInput,
           inputValue,
           items: allItems,
           renderListItem,
--- a/src/components/DropList/DropList.css.js
+++ b/src/components/DropList/DropList.css.js
@@ -34,6 +34,10 @@ export const MenuListUI = styled('ul')`
     padding: 0 0 5px 0;
   }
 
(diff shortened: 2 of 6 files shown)


src/components/DropList/DropList.Combobox.jsx
```diff
@@ -32,6 +33,7 @@ function Combobox({
   'data-cy': dataCy = `DropList.${VARIANTS.COMBOBOX}`,
   focusToggler = noop,
   handleSelectedItemChange = noop,
+  hideCustomListIfEmptyInput = false,
```
justinwolfe, line 36: I set a default value at the top level of `DropList`; should I not set it here?

plbabin: that's fine, all props here has a default value, let's keep the same pattern

plbabin (commented):
when the list create tag option is visible (with the flag set to `true`), the padding is missing at the end of the list:
CleanShot 2021-09-24 at 13 34 38@2x

that's the same example without the flag:
CleanShot 2021-09-24 at 13 34 30@2x

justinwolfe:
> when the list create tag option is visible (with the flag set to `true`), the padding is missing at the end of the list:
> CleanShot 2021-09-24 at 13 34 38@2x
> 
> that's the same example without the flag:
> CleanShot 2021-09-24 at 13 34 30@2x

Ooh, great catch, will fix!

justinwolfe:
@plbabin Fixed!
CleanShot 2021-09-24 at 10 46 47
!

tinkertrain:
Hello all, I'm looking at this and the first thing I ask is.... why?

"if there's no text entered into the search input, we don't render the dropdown at all, just the input" ---- why? what is the purpose of that?

I know is not your call @justinwolfe (and the code is fine) but... I don't get it, a combobox displays a list of available options that you can filter + create one if you don't find what you need. I'm guessing there's a reason (other than the thing being "cleaner") and I'm curious to know what it is.

I don't mean to be too negative, I just tried hard to avoid this type of overly specific prop that sort of deviate from expectations.

One of the big reasons that prompted the complete redoing of Dropdown into DropList was to simplify and standardize our dropdowns, but in no time we'll have the same thing again.

justinwolfe:
@tinkertrain Fair questions! Perhaps I should have asked this _why_ more directly myself before digging into the implementation—I was just working from the mocks [in the card](https://helpscout.atlassian.net/browse/HSAPPUI-137) and perhaps I've over-interpreted what's there and caused some unnecessary extra work 😓 . @digitaldesigner, what's your take on this?

digitaldesigner:
> @digitaldesigner, what's your take on this?

If I'm understanding correctly, you're talking about why there's no dropdown shown when there is no value, vs. showing it when there is? Interestingly the designs did originally have a droplist shown of the most popular tags immediately on focus of the input — this was only removed in the interest of (perceived) scope. The complications were:

- What does "popular" mean, are they per-user or per-mailbox?
- What if there are >5 tags in total?
- Is it more useful to show the 5 most recently created tags by the current User?

You get the idea. And generally when an idea results in this type of decision tree of complexity, we scrap it. But if you wanted to show the droplist on focus, and wanted it to contain _something_ then I don't think there's risk in showing 5 recent tags by any logic. Something could be more helpful than nothing.

justinwolfe:
@digitaldesigner 
Sorry for not being clearer with my question! I was basically just trying to ascertain whether [what I saw in the card](https://helpscout.atlassian.net/browse/HSAPPUI-137) with the dropdown not showing anything on an empty state was considered important to the design/intentional and something that I was correct in trying to implement here or whether I was reading too much into a screenshot and it doesn't actually matter that much and I shouldn't have tried to implement it.

From the conversation here, It sounds like we should just close this up and use the default! (As a retro thing, I should have checked in with you about the card before going so far down this development rabbit hole as I did here—I'll try to take more care with that in the future 🙏).

PR #984, opened 2021-09-28
Adds closeOnClickOutside to SidePanel

tinkertrain:
# Feature

Adds the ability to close the Panel when clicking outside, same as SimpleModal


jakubjanczyk (approved):
Looks and works great! Thanks for adding this :1st_place_medal:

PR #985, opened 2021-09-28
Upgrade colorway to 0.9.7

plbabin:
This update adds two new color to the palette: `green.650` and `yellow.950`. We introduce those two blended shades to improves our accessibility effort across all application. They will be used to increase readability on a green or yellow background.

CleanShot 2021-09-29 at 15 17 06@2x
CleanShot 2021-09-29 at 15 18 27@2x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2011,9 +2011,9 @@
       }
     },
     "@helpscout/colorway": {
-      "version": "0.9.6",
-      "resolved": "https://registry.npmjs.org/@helpscout/colorway/-/colorway-0.9.6.tgz",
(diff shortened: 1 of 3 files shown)


TerrenceLJones (approved):
LGTM!

digitaldesigner (changes requested):
Urh I hate to be *that* guy, but could we make that green `750` instead of `650`? On review its actually darker than `700`... so makes a bit more sense. Sorry. I'm sorry. Sorry.

plbabin:
> Urh I hate to be _that_ guy, but could we make that green `750` instead of `650`? On review its actually darker than `700`... so makes a bit more sense. Sorry. I'm sorry. Sorry.

you had to be _that_ guy 😉.

Everything has been updated from colorway to the PR description

digitaldesigner (approved):
You're a hero, thanks!

PR #986, opened 2021-09-30
AvatarRow: dynamic number of avatars

tinkertrain:
## Problem

Adding or removing avatars is not supported. The ResizeObserver callback gets binded on mount and needs to be re-done if the number of avatars change.

## Solution

- Add the number of avatars to the useEffect hook so that the `onResize` gets updated if that number changes
- Make sure not to add multiple ResizeObservers by holding on to values with a `ref`

https://deploy-preview-986--hsds-react.netlify.app/?path=/story/components-structural-avatarrow--dynamic-number-of-avatars

### Other
- Remove ie compat mode, will re add if necessary
- Refactor the onResize fn to be able to test easily (jsdom does not support ResizeObserver)

### DIFF
--- a/src/components/AvatarRow/AvatarRow.jsx
+++ b/src/components/AvatarRow/AvatarRow.jsx
@@ -27,8 +27,9 @@ function AvatarRow({
     avatarProps.size || 'md'
   ]
   const avatarRowRef = useRef(null)
+  const numberOfAvatars = avatars.length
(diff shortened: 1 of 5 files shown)


luketlancaster (approved):
I'll leave the final approval to @plbabin, but this looks good to me 👯‍♂️

PR #995, opened 2021-11-02
Fix portal trigger to not re-focus when unrelated props changed [BEMBED-305]

jakubjanczyk:
# Problem/Feature
Relates to: https://helpscout.atlassian.net/browse/CTRIAGE-2697

There was a bug that trigger element of a Modal has be re-focused when `isOpen` state value is false and if any property has changed. This is causing some issues with a Beacon message form. Follow the steps to reproduce:

- Go to docs.helpscout.com and click the Contact Us button to open the Beacon
- Start typing a message into the main body text area
- Add an image attachment
- Try editing the message - as soon as you start typing, focus moves to the "1 file" button and you can't type any more

This happens because the button for files is a trigger and we don't have modal opened. The implementation was lacking a check if the portal has been opened before, so we re-trigger when it is closed. Adding the check solves the problem.

I have added a story that I have used to find the issue (for future reference) and fixed some tests that were not really checking anything (one was always success and 2 more are added based on the fixed version).

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/Modal/Modal.stories.mdx
+++ b/src/components/Modal/Modal.stories.mdx
@@ -8,7 +8,7 @@ import DropList from '../DropList'
 import { SimpleButton } from '../DropList/DropList.togglers'
 import EditableTextarea from '../EditableTextarea'
 import EditableField from '../EditableField'
-import { MyModal } from './Modal.storiesHelpers'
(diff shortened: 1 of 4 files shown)


tinkertrain (approved):
Nice, thanks for the fix...! :shipit:

plbabin (approved):

PR #996, opened 2021-11-09
Allow component props to overwrite pagination attributes

plbabin:
# Problem

While testing hs-app with the  refactored pagination, we realized that it was not possible to overwrite some props from the parent component.

This update move the location of `getValidProps` after `data-testid`, `role` and `aria-label`. That way the parent component will be able to change the value of those props _if needed_


tinkertrain (approved):
🚢

PR #1001, opened 2021-11-16
Component: ScrollableContainer

tinkertrain:
The newest and shiniest stuff coming from design includes a revamped [scrollable shadow effect](https://cdpn.io/team/helpscout/debug/7a04a0396b147b146a3c71a3d0d6de0a).

In this PR:

- We already had the effect implemented for the layout on SidePanel, so we extract that into a custom hook: `useScrollShadow`
- Another custom hook: `useClientRect` to measure DOM elements.
- We create a new component `ScrollableContainer`, meant to replace `Scrollable` completely at some point.
	- With support of [simplebar](https://github.com/Grsmto/simplebar), as this library is recently getting used for scrolling.
- Add deprecation notes to `Scrollable`

## useScrollShadow

The hook is provided in case you need extra control and `ScrollableContainer` is a bit restrictive for your needs, do notice you have to add your own layout styles to the container and its elements (header, footer, body).

For an example of how to use it, check `SidePanel.layout` and the new `ScrollableContainer`

## useClientRect

Hook that gets you the measurements of a DOM element as calculated by `element.getBoundingClientRect()`, based on [React's guidelines](http://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node).

## ScrollableContainer

Powered by the aforementioned hook, this a very simple container that has the layout styles and effects already in place for you, all you have to provide is content (header and/or footer + body).

> Note that "look" styles (background, spacing, corners, etc) are up to you, `ScrollableContainer` just makes sure the top and bottom sections are fixed to their respective positions, the body scrolls and the shadow effects are added.

> The container supports being a child of a css grid, if you want it to fill an entire cell make both height and width 100%

## Stories:

Regular: https://deploy-preview-1001--hsds-react.netlify.app/?path=/story/components-structural-scrollablecontainer--default-story

With simplebar enabled and full height: https://deploy-preview-1001--hsds-react.netlify.app/?path=/story/components-structural-scrollablecontainer--full-height-with-simplebar


tinkertrain:
@jakubjanczyk Added you to make sure SidePanel stuff is not broken :)
@jaredmcdaniel Added you to check the UI as @digitaldesigner is out for a few days

jaredmcdaniel:
Sweet! Just a couple things from me @tinkertrain ...
1. I believe the header should always have a single line box-shadow "border"?

**Codepen (with border)**
image

**Storybook (without)**
image

2. Lets update the box-shadow transition timing to `0.2s` like the codepen vs `0.5s` here.

src/hooks/useScrollShadow.js
```diff
@@ -0,0 +1,94 @@
...
+  if (!scrollableRef || !scrollableRef.current) {
+    return {
+      isTopScrolled: false,
+      isBottomScrolled: false,
+    }
+  }
+
+  const scrollable = scrollableRef.current
+  const style = window.getComputedStyle(scrollable)
+  const paddingTop = Number.parseInt(style.paddingTop.replace('px', ''), 10)
```
jakubjanczyk, line 77: I believe `parseInt` would ignore any non-numeric characters in a string, so it's not necessary to do a `replace` here. (same in line 79)

src/components/ScrollableContainer/ScrollableContainer.test.js
```diff
@@ -0,0 +1,112 @@
...
+    const footer = container.querySelector('.ScrollableContainer__Footer')
+
+    // Elements present
+    expect(containerScroll).toBeInTheDocument()
+    expect(header).toBeInTheDocument()
+    expect(body).toBeInTheDocument()
+    expect(footer).toBeInTheDocument()
+
+    // Passing custom classnames
+    expect(header.classList.contains('TESTING__HEADER')).toBeTruthy()
```
jakubjanczyk, line 84: this can be replaced with `expect(header).toHaveClass('TESTING__HEADER')` (and the same for assertions below). It would be giving more meaningful error messages in case it's not true, instead of `expected false to be truthy` or something like this :smile:

src/components/ScrollableContainer/ScrollableContainer.test.js
```diff
@@ -0,0 +1,112 @@
...
+    expect(body.classList.contains('TESTING__BODY')).toBeTruthy()
+    expect(footer.classList.contains('TESTING__FOOTER')).toBeTruthy()
+
+    // Correct layout styles
+    expect(window.getComputedStyle(containerScroll).display).toBe('flex')
```
jakubjanczyk, line 89: I have not checked this (:sweat_smile:), but maybe this matcher could be used to make those assertions a bit more concise? https://github.com/testing-library/jest-dom#tohavestyle

src/components/ScrollableContainer/ScrollableContainer.test.js
```diff
@@ -0,0 +1,112 @@
...
+    )
+    expect(window.getComputedStyle(containerScroll).flexGrow).toBe('1')
+    expect(window.getComputedStyle(containerScroll).overflow).toBe('hidden')
+    expect(window.getComputedStyle(body).overflow).toBe('auto')
+    expect(window.getComputedStyle(body).flexGrow).toBe('1')
+    expect(window.getComputedStyle(header).zIndex).toBe('3')
+    expect(window.getComputedStyle(footer).zIndex).toBe('3')
+
+    // starter shadows
+    expect(header.getAttribute('style')).toBeFalsy()
```
jakubjanczyk, line 101: could be replaced with `expect(header).not.toHaveAttribute('style')` (and similar below)

src/components/SidePanel/SidePanel.layouts.jsx
```diff
@@ -19,32 +19,12 @@ export function HeaderAndFooter({
...
-      footerRef.current.classList.add('with-shadow')
-    } else {
-      footerRef.current.classList.remove('with-shadow')
-    }
-  }, 100)
+  const [handleOnScroll] = useScrollShadow({
+    bottomRef: footerRef,
+    scrollableRef: bodyRef,
+    topRef: headerRef,
+    shadows: { initial: 'none' },
```
jakubjanczyk, line 26: There has been one issue (even before, but still present), that when there is scroll, the shadow at the bottom does not appear until the content is scrolled. This can be misleading users that there is nothing to scroll to (especially when scrollbar is not visible). Here's example: https://c.hlp.sc/xQujRO7Q

Do you think this setting needs to be changed or maybe it's some different issue to be solved?

src/hooks/useScrollShadow.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        const { isBottomScrolled } = handleShadows(scrollableRef)
+
+        if (isBottomScrolled) {
+          bottomElement.style.setProperty('--scroll-shadow', scrolledShadow)
+        }
+      }
+    }, 100)
+  }, [scrollableRef, bottomRef, topRef, scrolledShadow])
```
jakubjanczyk, line 40: I think the refs here are not needed. Since they are not mutable from react perspective, the effect would never be called because the ref has changed, so there's no need to put them as dependencies. At least, that's what linter is telling me in hs-app-ui :smile:

src/hooks/useScrollShadow.js
```diff
@@ -0,0 +1,94 @@
...
+  bottomRef,
+  scrollableRef,
+  topRef,
+  shadows = {},
+}) {
+  const initialShadow = shadows.initial || BOX_SHADOW_INITIAL
+  const scrolledShadow = shadows.scrolled || BOX_SHADOW_SCROLLED
+
+  useEffect(() => {
+    setTimeout(() => {
```
jakubjanczyk, line 20: should we clear timeout when effect re-runs? Or rather when it's unmounted, by using the returned function from the `useEffect`?

src/hooks/useScrollShadow.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }, 100)
+  }, [scrollableRef, bottomRef, topRef, scrolledShadow])
+
+  const handleOnScroll = throttle(e => {
+    const { isTopScrolled, isBottomScrolled } = handleShadows(scrollableRef)
+    const topElement = topRef && topRef.current
+    const bottomElement = bottomRef && bottomRef.current
+
+    if (topElement) {
```
jakubjanczyk, line 47: this if and the one below could be extracted to a function to avoid some duplication (just passing element, or even ref, and `isScrolled` flag as parameters)

jakubjanczyk (commented):
Looks great and SidePanel seems to be working! :) 
I've left a couple of suggestions and one question about shadow for side panel, but otherwise perfect example of iterative generalization of code :smile:

src/hooks/useScrollShadow.js
tinkertrain: I see, I kind of thought the same but the ESlint rule about dependencies kept nagging me to put them, I'll just add an ignore directive

src/hooks/useScrollShadow.js
tinkertrain: I think you are right, I always try to be explicit in these type of cases, because you know... JavaScript. But I will remove it because there are other cases that could happen like padding in % or ems or something.

src/hooks/useScrollShadow.js
jakubjanczyk: If it's bothering, then maybe I am wrong :smile: But I am sure it was nagging me the other way around. What about when you do `scrollableRef.current` instead?

src/hooks/useScrollShadow.js
tinkertrain: Thanks 👌

src/hooks/useScrollShadow.js
tinkertrain: 👍

src/components/ScrollableContainer/ScrollableContainer.test.js
tinkertrain: I always forget there are so many matchers! thanks for the reminder

src/components/ScrollableContainer/ScrollableContainer.test.js
tinkertrain: ditto, looks nicer

src/components/ScrollableContainer/ScrollableContainer.test.js
tinkertrain: ditto

tinkertrain:
@jaredmcdaniel Thanks for looking! Both things taken care of 👍

jaredmcdaniel (approved):
Looks great!

src/hooks/useScrollShadow.js
plbabin: what you could do is extract the current value and used that as a dependency

```javascript
const {current: scrollableRefElement} = scrollableRef

useEffect(() = > {
....
}, [scrollableRefElement, bottomRef, topRef, scrolledShadow]))
```

tinkertrain: The hook supports passing DOM elements directly too now, so I added the dependencies

plbabin (approved):
thank you for adding all the code needed to support `SimpleBar`

amazing work🎉 🎉 🎉

TerrenceLJones:
Hey @plbabin @tinkertrain Is this component supposed to eventually handle the UI elements like present on the Conversation View page?

Screen Shot 2021-11-24 at 12 42 17

Screen Shot 2021-11-24 at 12 42 24


Because if so, can this currently handle the transition from the inline header/footer to the fixed header/footer? We are currently handling this via the [`useShouldShowFixedHeader`](https://github.com/helpscout/hs-app-ui/blob/9ed29363d1b11bd5fbebb33d3e85ebb5e1f92b67/src/apps/mailbox/screens/ConversationView/components/Content/hooks/useShouldShowFixedHeader.js) hook in conjunction with the `ScrollableContent` component. 

Also will this implementation handle browser resizing?

tinkertrain:
@TerrenceLJones Currently it doesn't handle that transition, but the internal mechanisms should allow me to add it. I can work on that next.

I don't quite understand about "handling" browser resizing, the container can have any width (directly as a prop or you could use styled components) and you should be able to use media queries too, also both the header and footer have their widths set to 100%.

[This story has the width as 70%](https://hsds-react.netlify.app/?path=/story/components-structural-scrollablecontainer--full-height-with-simplebar). Is there anything else beyond that I'm not considering?

PR #1002, opened 2021-11-17
HSDS-217: Add AvatarsSelectionModal component

jakubjanczyk:
# Problem/Feature
For the new Messages Avatar feature ([spec](https://paper.dropbox.com/doc/Spec-Message-Avatars--BWWk8YUR7q~oktQpiyRcoHZaAg-6sXfLQiYbxa5ZQFUHterJ)), we need to have avatar selection modal available. Component like this has laready been created as part of work on Beacon avatars. Here's demo of how it looks right now: https://c.hlp.sc/8Lu9qYdn

This PR is adding new component to HSDS (`AvatarsSelectionModal`), that is a copy of the HSAPP component. [Here](https://github.com/helpscout/hs-app/tree/develop/site/js/apps/beacons/react/components/ContactFormShowAvatars/components/AddNewUserAvatarsModal) is the original component.

### Designs

The original design: https://www.figma.com/file/hOhDp1xvPMFI0kyx0y3i0y/Discovery%2F-Custom-Avatars?node-id=132%3A8152

Slightly updated design for Messages: https://www.figma.com/file/BQCNhAY4nh0AI53y0B1EXB/Spec%2F-Message-Avatars?node-id=4%3A4668

One thing I am aware of, that is not in-line with the design is the search input. It's bigger on implementation than on desgin, however, I am using standard `Input` component, so I think we should stick with it here. There is an option to make it smaller, but then it's not the same as on the design as well :shrug: 

### Changes
The component was in most part just copied from HSAPP, but a few things needed to be changed:
- Enabled single selection mode. For Messages we need to allow selecting only single avatar and not force users to deselect previous to select a new one. Limiting to 1 could be done already, but the behavior was as mentioned - needed to deselect first. Change was made to set `multiSelect` prop on `CheckMarkCard.Grid` component used internally.
- Adjusted texts inside the modal, based on the provided limit (it's different for `1` and anything `>1`)
- Some refactoring to make it more generic and removed code specific to Beacon avatars
- Added tests

### Storybook
A few stories has been added for this component, presenting most significant features (single selection, multi selection, pre-selected avatars)

### Testing
It should be enough to check the mentioned stories to verify if everything is working. Unit tests should cover most cases.

Screenshot from 2021-11-17 12-13-29

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests


src/components/AvatarsSelectionModal/AvatarsSelectionModal.stories.mdx
```diff
@@ -0,0 +1,55 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs/blocks'
+import AvatarsSelectionModal from './AvatarsSelectionModal'
+import {
+  DefaultAvatarsSelectionModal,
+  PreSelectedAvatarsSelectionModal,
+  SingleSelectionAvatarsSelectionModal,
+} from './AvatarsSelectionModal.storiesHelpers'
+
+<Meta
+  title="Components/Overlay/AvatarsSelectionModal"
```
jakubjanczyk, line 10: I am not sure if this is under the correct folder

jakubjanczyk:
@tinkertrain This is the first time I am adding new component to HSDS, so any comments/suggestions are appreciated :smile:

tinkertrain:
@jakubjanczyk Good work! A couple of notes:

- I see this is using the current or should I say "old" scrolling pattern (with the gradient when overflowing), I think you should suggest the change to the newest "solid shadow" pattern, as I believe it will get used from now on, so it will probably have to be changed anyways at some point, as a reminder here's the PR which will be released pretty soon: https://github.com/helpscout/hsds-react/pull/1001 cc @alexestrada 
- I'd like us to try and use `SimpleModal` instead of `Modal`, I believe it should support what you need here, and also I'd like to use the pattern of "layouts" just like the SidePanel so that we don't end up with several custom modals that are not shareable.

So the layout can be something like [this](https://deploy-preview-1001--hsds-react.netlify.app/?path=/story/components-structural-scrollablecontainer--default-story), I think this will be a regularly used pattern (the modal with the header, footer and scrollable content).

Let's say you create the layout as "HeaderAndFooter" (named like the one in SidePanel, as it's basically the same thing).

You'd have something like:

```jsx
// SimpleModal/SimpleModal.layouts.js

import ScrollableContainer from './ScrollableContainer'
import { BodyUI, FooterUI, HeaderUI } from './SimpleModal.layouts.css'

export function HeaderAndFooter({headerContent, footerContent, children}) {
  return (
    <ScrollableContainer
		header={<HeaderUI>{headerContent}</HeaderUI>}
        body={<BodyUI>{children}</BodyUI>}
		footer={<FooterUI>{footerContent}</FooterUI>}
    />
	)
}
```

Then `AvatarSelectionModal` would be something like:

```jsx
// AvatarSelectionModal.js
import SimpleModal from '../SimpleModal'
import { HeaderAndFooter } from './SimpleModal.layouts'

<SimpleModal>
  <HeaderAndFooter headerContent={} footerContent={}>
	//.. your stuff inside the scrollable area
  </HeaderAndFooter>
</SimpleModal>
```

The only thing I'm unsure is whether that last one belongs in HSDS or not, conceptually I think not, but I guess it makes it easier since it's supposed to be used in Beacon and HS App UI.

Let me know what you think, I realize this suggestion is a bit more work if you need a hand just let me know! Happy to 🍐 on this.

plbabin:
Just want to shime in as well. I don't think this component belong in HSDS. 

I can see the reasoning to build it here (to share it between hs-app and hs-app-ui) but we are trying to move away from building complex component that contains tons of smaller components within HSDS (like ChatScroller, Condition, etc).

It's not a hard no, meaning that I can change my mind if we think the HSDS context is the right one, but my stand right now would be to move it as a shared component within hs-app-ui. I think we can live with the duplicate code with the old component for now, that's the downside of having two frontend repos.

If it start to happen a lot, maybe we should find a way to make the shared component from hs-app available in hs-app-ui (maybe through git submodule....I don't know)

jakubjanczyk:
@plbabin Yeah I was considering where it goes and I share your doubts. The main reason is that it's in 2 repos of course, so there is no other place to put component to be used in 2 places. At the moment I don't see any other components like this in hs-app that would need to be used in hs-app-ui, so looks like one time thing.

Good thing is that, apart from the storybook, my work here would not be a waste of time :smile: I can just copy the code from here to hs-app-ui (and leave the hs-app code as it is).

If you and @tinkertrain feel strongly about not adding it here, let's close this one and I'll move the code as part of other task.
Then, per @tinkertrain comment, adding those Layouts to modal could be made in some another PR.

tinkertrain:
Thanks @jakubjanczyk and @plbabin 

I think we can move forward with:

1. Creating a SimpleModal layout in HSDS
2. Moving the Avatar Selection Modal specific code to hs app ui and leave hs app as is 🤭 If it becomes an issue for whatever reason, it should be easy to port.

@jakubjanczyk Let me know if you want to create that layout (all you need is that ScrollableContainer PR to be released, which I will do today) or I can do tackle that, no problem

tinkertrain:
@jakubjanczyk One last thing to confirm before moving on with the above is to get design approval on using the "solid shadow" pattern instead of the "subtle gradient"

jakubjanczyk:
@tinkertrain Great! I can add it, but if you have available time, feel free to take it :smile: Feels quite straightforward and quick to add.

About the design confirmation - @alexestrada could you take a look? Details are in one of the previous comments. However, I think it wouldn't block moving the AvatarsSelectionModal to hs-app-ui either way, so I'll just go with it now and wait for the final confirmation before creating PR :)

alexestrada:
@jakubjanczyk @tinkertrain I'm on board with using the solid shadow pattern — took a look at the PR and it looks great!

jakubjanczyk:
Closing this, as the component has been moved to hs-app-ui, here's the related PR: https://github.com/helpscout/hs-app-ui/pull/209

PR #1003, opened 2021-11-17
Button Standardization

plbabin:
This update standardize the `Button` to make it easier for devs and design to use button in multiple context

## theme
We removed the `kind` and the `state` props and move them to a theme props. The theme will handle the coloring of the button:

| value |   |
|-------|---|
|  blue  | CleanShot 2021-11-30 at 16 47 19@2x  |
|   red   |  CleanShot 2021-11-30 at 16 47 31@2x |
|   grey    | CleanShot 2021-11-30 at 16 47 25@2x  |
| green | CleanShot 2021-11-30 at 16 47 05@2x |

## outlined
Instead of having multiple kind and state, it's not possible to convert any theme to an outlined version:

```javascript
<Button theme="red" outlined>Button</Button>
```
will look like that:

CleanShot 2021-11-30 at 16 50 57@2x

## linked
Same as outlined, it's now possible to set any button to look like a link and keep all focus ring/size properties

```javascript
<Button theme="red" linked>Button</Button>
```
will look like that:

CleanShot 2021-11-30 at 17 02 16@2x

## IconButton
This component follows the same rewrite as the `Button` component. It still uses the `Button` behind the scene, but limit some props when composing the component with hooks. Only the `lg` and `xl` are supported and the radius is overwritten in the `IconButton` css. It's not possible anymore to change the icon size (it will always be `24`)

We added support to render an avatar instead of the icon. The avatar size will be 36px or 32px.

```javascript
const avatarProps = {
  image: '...'
  fallbackImage: '...'
}
<IconButton avatarProps={avatarProps} theme="blue" />
```
CleanShot 2021-12-01 at 09 09 48@2x

We also added a label to the `IconButton`. Passing down text as a children will render it right after the icon/avatar:

```javascript
const avatarProps = {
  image: '...'
  fallbackImage: '...'
}
<IconButton avatarProps={avatarProps} theme="blue" filled>with button label</IconButton>
```
CleanShot 2021-12-01 at 09 11 03@2x

Default props are a little bit different for the IconButton. If not styles are specified (`outlined` or `filled`) the default style will be `outlined`.

| prop     | default |
|----------|---------|
| theme    | grey    |
| outlined | true    |

### Functional component && forwardRef
First, we update the `Button` to be a functional component. The component was already really small, moving to functional allow us to implement `forwardRef` for the component. Because we had to support innerRef and buttonRef, but now those are all gone from the props list.

### RouteWrapper
This wrapper component was also problematic for forwardRef. It was messing up the ref prop passed down the Button. We did a quick investigation and we are not really sure what is the benefits of this old component. We decided to drop it while keeping the wrapper around just in case. If an issue occur with a `Button` or `Link`, we'll be able to fix it quickly by wrapping the component directly in hs-app.

### CSS variables
The other big change on the `Button` component is the way CSS is handle.

Now we leverage CSS variables to lay down the tokens foundation of the component. Each variable is assigned at the right place, and each kind, size & states will update only the variables.

Also, every time we wanted to use the Button component and update the style we had to increase the specificity of the selector by using `is-default`. We dropped styles from the `default` classname and moved them on the root component. That way any overwrites could be done by updating the CSS variables directly, making it easier than before to leverage `Button`

### Focus ring
`link` and `default` didn't had any focus ring. It didn't make sense since we want to have a visible focus state for every element on the page.

### Props cleanup
While we were working on the component, we removed a lot of props that didn't make sense or were not used at all:

|  🪓 Props  |
|---|
| allowContentEventPropagation |
| disableOnLoading |
| fetch |
| buttonRef |
| spinButtonOnLoading |
| canRenderFocus |
| innerRef |
| isBlock |

## Droplist Togglers IconBtn & MeatButton
Instead of creating a button from scratch, we update `MeatButton` and `IconBtn` to use `IconButton` behind the scene

## CopyButton
This component was refactored to be a functional one.  

## FocusRing
The focusRing mixin was update to leverage CSS variables, making it easier to reconfigure it based on the component using it.

```css
--focusRingOffset: -2px;
--focusRingShadow: ${focusShadow};
--focusRingRadius: inherit;
```

## Fixing things
Updating the component had some side effect on `Table`, `Pagination`, `DropList.SplittedButton`, `Condition`. Even though everything was minor, we fixed everything as part of this PR.

# Breaking change

We change a lot here, the blocks team will create a codemod to facilitate the button migration on hsapp and hsappui. Here is a couple of things we'll have to keep in mind that might not fit in a codemod:

* Component overwriting through the `.is-default` classname
* Button `buttonRef` / `innerRef`
* NavLink & Link (`Customers` and `Team sidenav`)

We are also planning to create a beta release of HSDS to run the migration with. That way we will not block any other HSDS update that will be needed in our app.


tinkertrain (approved):
I ❤️ this.

The use css variables is a win! I need to keep them in mind and use them more.

Also, thanks for updating related components like icon button, droplist togglers, etc 🙇

digitaldesigner:
This is cool, I'm all up for consistency and consolidation. Some discussion points below — basically the following would enable me to find consistency with this PR and the proposed new convo designs:

### Consolidation
- Seems like `kind=tertiary` is the same as `kind=secondary, state=success` — do we need tertiary at all?
- What is `state=grey` on the `kind=primary` button? Feels redundant, but if we keep I would encourage renaming to "muted" or perhaps change the name from "state" to "style" (since state in my mind refers to hover, focus etc)

### Min-Width and Padding
- Currently the secondary button doesn't have the same min-width as the primary, and I think it should (i.e. `min-width: 120px`
- However, `shape=rounded` shouldn't have any min-width
- The `SplittedButton` should also have no min-width on the text size of the component

### The Rounded Conundrum
- I need the `shape=rounded` to have have different properties to its square counterpart (see below). So I'm wondering whether it should be considered a `kind`, rather than a prop
  - Font-weight medium
  - Padding left/right: 20px
- Additionally, I need the rounded shape to be `44px` height. This is so it can match the `IconButton` in the fixed footer of the new convo design. However, 44px exists somewhere outside of the preset sizes. I'm reluctant to suggest another size

### Updates
- IconButton icon color should change from `charcoal.400` to `charcoal.500` on hover
- Would love for the IconButton to have a darker hover background (`grey.300` instead of `grey.200`) and an inset white border so that the background color and border don't clash (see image)

Not sure if this is helpful, but here's the breakdown of the ideal button types:

digitaldesigner:
Trying to think out a more robust structure, wondering whether the "Primary", "Secondary" classifications are redundant? The following could apply to the regular button (i.e. continue to separate `Button` from `SplittedButton` and `IconButton`) and gives us 400+ style options:

| Style | Size | Is Rounded | State | Theme |
| :--- | :--- | :--- | :--- | :--- |
| Filled _(Default)_  | Large _(Default)_  | False _(Default)_  | None _(Default)_  | Blue  _(Default)_
| Outline  | Medium  | True  | Hover  | Mono
| Link | Small  |   | Active | Green
|  | X-Small  |   | Disabled | Red
|   | X-Large ⚠️  | | Loading
| | XX-Large ⚠️ 

⚠️ The existing `X-Large` (50px) today would become `XX-Large`, and the `X-Large` size can be used for the 44px chaos required by convo redesign

It's pretty chaotic, but allows us to support whatever is in the app currently but also the new convo stuff https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=10648%3A27724

tinkertrain:
@digitaldesigner 

> Trying to think out a more robust structure, wondering whether the "Primary", "Secondary" classifications are redundant? The following could apply to the regular button (i.e. continue to separate Button from SplittedButton and IconButton) and gives us 400+ style options

I'd love to have this if we get the chance, I really dislike how you can combine "kinds", "states" (success, warning),  and more into invalid or confusing states, for example I can set the button to state="warning" and kind="secondary" and I get a grey outlined button, (when warning is supposed to be yellow), same with kind="link" and any of the state colours. The "state" colours only seem to apply on "primary" or "tertiary" buttons. Even if we _never_ use red links, passing "link" and "red" should give you a red link, as that is what you would expect looking at the code.

The only problem is it will take some time to then apply the changes in HSApp and the rest... but I think is worth it.

digitaldesigner:
@tinkertrain Absolutely — I think that's just what happens as a result of bolting-on lots of properties over time. My proposed classifications still produce some versions of buttons we don't really use, but they're not terrible (and still accessible). For example in order to have a `rounded, outline` button for the new convo status, there are a couple of filled and non-rounded versions available of that same button we don't use elsewhere. That said, it gives us a lot more creative freedom and less chance of a designer needing to create a new variant since the available overrides are so expansive:

digitaldesigner:
Was talking to @plbabin earlier about keeping `IconButton` as a separate component rather than provide it as a property within `Button`. The rationale is that by adding it to the Button scope, there's potential to create a whole load of outcomes that we don't ever want to support. Instead I think we should provide a similar set of configurations as `Button` but keep them limited to intended usage (i.e. no `theme:blue`, `isRounded:false`, `size:xs` or `state:loading` etc since we never have a need for those things).

So instead we'd have the following configs (where the sizes, themes and states correspond exactly to their Button counterpart):

| Style | Size | Has Text | State | Theme |
| :--- | :--- | :--- | :--- | :--- |
| Outline _(Default)_ | X-Large _(Default)_ | False _(Default)_ | None _(Default)_ | Gray _(Default)_ |
| Filled | Large | True | Hover | Green |
| | | | Active | Red |

Which result in [this many variants](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=10658%3A27762) (72). As a proof of concept, the following designs are created entirely using the `Button` and `IconButton` components in varying states/props:

New Convo Actions

plbabin:
I think it's ready for another review, all internals buttons has been updated to the new props system

this is the script to update buttons: [scripts/codemod/transforms/ButtonStandardizationTransform.js](https://github.com/helpscout/hsds-react/blob/df005bd4b22fcf1c845095c2385a8931e99b4f1b/scripts/codemod/transforms/ButtonStandardizationTransform.js). That will help speeding thing up on hs-app-ui and hs-app

digitaldesigner (changes requested):
Ooof, phenomenal amount of work — nice work @plbabin. Think I've gone through and tested just about every variant (I think), feedback below:

## Global
- This applies to every filled state — the border color should change to match the background so that it's one solid shape. So [in this case](https://c.hlp.sc/X6uby04N), the border should be `blue.600` (where it currently remains unchanged at `blue.500`) 

## Buttons/ Button
- `20px` instead of `30px` left and right padding on the `md,lg,xl,xxl` rounded buttons
- `theme:green, outlined:true, state:hover` border-color is too dark. There shouldn't be any change in the border color

## Icon Button
- `filled:false` should have an inner white border when there is no text (i.e. fill doesn't go to the edge on icon-only buttons). See [this image](https://c.hlp.sc/geupXolW)

## Question
- On all buttons, I'm wondering whether we could forego min-width and use only minimal padding when `linked:true` on Buttons/ Button? Feels unnecessary to force lots of whitespace. [This is the most common example I can think of](https://c.hlp.sc/2NuPZ02v)

plbabin:
@digitaldesigner all points have been taken care off, including the link `min-width`, I do think it look better that way! Let me know if you need anything else!

jakubjanczyk:
Hey @plbabin! I've heard about this update to the Button and I have one thing to consider.
I absolutely love the use of CSS variables here, however there must be an but :smile: 
The problem I see if for Beacon Embed. We don't use Button from HSDS directly there (as far as I understood it's because it was too heavy for our needs), but there is one instance of a Button that is used indirectly: `MessageCard.Button`.

Now there are two complications:
1. Since we are not using Button directly, we would need to update the styles in Beacon code to match the new styles, if they apply of course (I have not checked it yet, though)
2. CSS variables are not supported in IE11, that might make a problem for `MessageCard.Button`. To be honest, I have not read through the whole code, because there's a lot and I figured out it would be easier to ask at this point :) Are there any fallbacks that could be used in this case?
Also, Do I see correctly that usage of `theme` from `ThemeProvider` and `brandColor` has been removed from button? If so, the problem might be that we were using this feature to color MessageCard button with color selected for a beacon.

All of this might have been addressed already, but I have not seen anything in the discussion here or in the code. Sorry if it was! :)

plbabin:
> 
@jakubjanczyk 

That's some really great point. I went through the code in beacon and didn't found any Button instance without realizing that the brandColor code was for the MessageCard button. I went back and update the new button state to support brandColor https://github.com/helpscout/hsds-react/pull/1003/commits/488b5af1dbb46e4908c66d324275b7cd57017669

CleanShot 2021-12-02 at 14 24 21@2x

It should be used only in a specific scenario like beacon

I added a couple of fallback values. It will not be a perfect solution, but the button should look like the grey one. Downside is that user on IE11 will not see the branded color, but at least they will see something. We would need to QA that! https://github.com/helpscout/hsds-react/pull/1003/commits/84c165383db60c0efca89408f935a6fdeb20a491

digitaldesigner:
Looking awesome, only one small change from me on the Button / IconButton where there's no text — the background fill should be a little darker on each of the themes. Borders and icon colors are fine, just the fills that need an update :)

colors

Last thing, wondering if you'd be up for adding a couple darker pinks to the mix? Currently none of the pink range pass WCAG AA compliance, but it looks _so_ much better than red. If we added `#E9074B` as `pink.950` and `#C5003B` as `pink.1000` (both pass AA compliance) then we could swap out the entire red theme on all buttons for something a little... pinker. Thoughts?

Pink

plbabin:
Yes let's do it for pink. I can add it to colorway. Can you update the button with those colors in figma and I'll work on that tomorrow ! 

And the darker fill is really just for icon only button right?

digitaldesigner:
All done buddy — pink.950 and pink.1000 added and replaced all `theme:red` instances ([see Figma](https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=10648%3A27724)). Updated the previously attached image too, see below — and yes, the background fill only applies to the icon-only button

Frame 17

jakubjanczyk:
Thanks @plbabin! I've went ahead and updated hsds to the latest beta version in beacon repo and it all looks good in Chrome. I have no way to verify IE11 yet. I am only thinking if maybe the default value should be blue instead of grey, but that's something we can agree on :)

plbabin:
@jakubjanczyk I updated the default to be the `blue` theme. 

@digitaldesigner `theme=red` has been updated! Same with the icon only hover color

digitaldesigner:
This is looking great @plbabin — everything is addressed and looking great. However in testing I noticed that we use a version of the IconButton in `size:sm` and `theme:green` for the operator buttons in [`Forms/Condition`](https://hsds.helpscout.com/?path=/story/components-forms-condition--default-story). So in order to make that work, I'm going to need a couple spacing tweaks to the `size:sm` version of IconButton:

- No space between icon and text
- `10px` padding left
- `15px` padding right
- I know we discussed down-sizing the icon on Slack, but it feels like it should remain `24x24` in this instance.
- I'm also wondering whether we need a property for `isRounded` for this one too? Seeing as Condition uses these as square icons?

Here are the adjustments vs. the closest I could get using the existing component:
Frame 17

plbabin:
@digitaldesigner I went in another direction. It felt like the proposed change to IconButton looked more like a regular button. So I added a prefix/suffix icon prop to allow adding an icon near the button content:

CleanShot 2021-12-06 at 10 37 38

For the padding, since I wanted to support all sizes, I do a calculation with the current padding when there is an icon. I thought 8px looked better to have the same "padding" on both side, happy to change the value to whatever feels best to you 
https://github.com/helpscout/hsds-react/blob/424bbcaa1a4d575ac942d9689e9f9ed598319a32/src/components/Button/Button.css.js#L273-L280

This is how it looks within Condition (both `and` and `or` button used the new prop):

CleanShot 2021-12-06 at 10 41 32@2x

digitaldesigner:
Yeah I love this. Everything is looking great from my side, and I've updated HSDS Figma to reflect the prefix/suffix icon

digitaldesigner (approved):
This is looking great. Million $ question though... how do we test this? Is there a way to get it onto Staging?

plbabin:
@digitaldesigner I'll tag you when it's ready on hs-app

PR #1005, opened 2021-11-30
ScrollableContainer enhancements (1)

tinkertrain:
## Feature

A follow up for the new `ScrollableContainer` component, we add:

- Enables nesting a `ScrollableContainer` inside another (generally the header or footer)
- The ability to do something when we reach the top or bottom of the content.
- The ability to smoothly adapt if either the header or footer change size

[Showcase video](https://i.hlp.sc/FTzc0E)
[Story](https://deploy-preview-1005--hsds-react.netlify.app/iframe.html?id=components-structural-scrollablecontainer--full-example&args=&viewMode=story)

### useScrollShadow
= 🗞 updated =

Besides the `onScroll` handler it now returns:

2. `isTopScrolled`: content has been scrolled and there is content overflowing under the top or header section. When `false` it means we are at the top of the scrollable node.
3. `isBottomScrolled`: content has been scrolled and there is content overflowing under the bottom or footer section. When `false` it means we are at the bottom of the scrollable node.

###  useClientRect -> useMeasureNode
= 🗞 updated =

Changed the name of this custom hook as it now has the ability to use ResizeObserver and it's more general and descriptive.

### useFancyAnimationScroller
= ✨ new =

Custom hook that will adjust the height of a given element in relation to the scrolling amount of a container.

- [See the effect](https://i.hlp.sc/dlbxNC)
- [Story](https://deploy-preview-1005--hsds-react.netlify.app/iframe.html?id=components-structural-scrollablecontainer--full-example&args=&viewMode=story)

## Other

- Creates a more general utility to setup an observer inside `useMeasureNode` replacing the one in `AvatarRow`
- Got rid of a potential memory leak warning on EditableField.Input


digitaldesigner (approved):
This is awesome — approved from me, but I had some points below nonetheless:

1. In the Full Example, ideally that form would extend height-wise to fill the available real-estate so that the scroll overflow between the form and the content was a last-resort. This is most applicable to the convo redesign stuff where the bottom action bar can extend to fill as much height as possible [leaving a minimum content area of 280px](https://paper.dropbox.com/doc/Spec-Replying-Bottom-Bar--BXO6dNEDY4N2Lof7LKADo8WeAg-pJj6Kk07zbKX7atv6jQZq#:h2=Content-Overflow-Required). This might already be the case, but thought I'd flag it anyway :)
2. This is my no means a requirement _at all_ — but when I was scrolling the content quickly, there was a noticeable disconnect between the scroll speed and the header animation. I've always loved UI (notably on iOS) whereby the header animation is dependent on scroll percentage, [this is an example](https://twitter.com/itsJonQ/status/1234271959388540928) of what I mean. Again, just thought I'd offer this up as a "nice-to-have"

src/components/ScrollableContainer/ScrollableContainer.jsx
```diff
@@ -131,6 +150,11 @@ const shadowShape = PropTypes.shape({
   scrolled: PropTypes.string,
 })
 
+const resizeObserversShape = PropTypes.shape({
```
TerrenceLJones, line 153: Was there a need to create a variable for this prop?

TerrenceLJones (approved):
LGTM! I had one question concerning the footer. In the example it wasn't clear, but can the footer transition from a fixed state to an inline state in the same way the header can?

tinkertrain:
@digitaldesigner OK... I think I have it 

https://i.hlp.sc/dlbxNC

Full example story updated too: https://deploy-preview-1005--hsds-react.netlify.app/iframe.html?id=components-structural-scrollablecontainer--full-example&args=&viewMode=story

@TerrenceLJones yeah, with `onScrollableSectionsStateChange` (long name! I wanted it to be clear that this is different from `onScroll`) you get flags when you reach both the top and the bottom. Then you can act accordingly. This would be the easiest way to do stuff on header and footer and probably the most common. For the footer in the full example, since it's quite a bit more complex UI it's using a custom hook, one that I named `useConvoLayoutEffect`, as I said in the video above this will not be part of HSDS, but you'll be able to grab it from the full example code and put in hs-app-ui pretty much verbatim.

src/components/ScrollableContainer/ScrollableContainer.jsx
tinkertrain: On black friday they were only a nickel a pop so I just said why not

TerrenceLJones:
> @digitaldesigner OK... I think I have it
> 
> https://i.hlp.sc/dlbxNC
> 
> Full example story updated too: https://deploy-preview-1005--hsds-react.netlify.app/iframe.html?id=components-structural-scrollablecontainer--full-example&args=&viewMode=story
> 
> @TerrenceLJones yeah, with `onScrollableSectionsStateChange` (long name! I wanted it to be clear that this is different from `onScroll`) you get flags when you reach both the top and the bottom. Then you can act accordingly. This would be the easiest way to do stuff on header and footer and probably the most common. For the footer in the full example, since it's quite a bit more complex UI it's using a custom hook, one that I named `useConvoLayoutEffect`, as I said in the video above this will not be part of HSDS, but you'll be able to grab it from the full example code and put in hs-app-ui pretty much verbatim.

@tinkertrain Thank you for the response and the video! As you spoke about it, I remembeedr that the functionality to determine whether to show then inline vs fixed footer wasn't going to be determinate on page scroll, but instead on the calculation of whether the height of the thread content + height of expanded editor would cause a page overflow. So this would be handled by the application code and not by the HSDS component.

src/components/ScrollableContainer/ScrollableContainer.jsx
TerrenceLJones: 😂

digitaldesigner:
@tinkertrain You absolutely nailed it. This is perfect and _exactly_ what I was referring to. Also no dramas about not animating between footer heights — you're right, there's not much need :)

src/components/EditableField/EditableField.Input.jsx
```diff
@@ -81,6 +81,7 @@ export class EditableFieldInput extends React.Component {
 
   inputRef
   inputWrapperRef = React.createRef()
+  _isMounted = false
```
plbabin, line 84: is it a cleanup change from you, or it's an overwrite to `main` that we don't want?

plbabin (approved):
all look good

left a comment on `EditableField.Input` but it might be nothing

src/components/EditableField/EditableField.Input.jsx
tinkertrain: From me, got rid of a memory leak warning on EditableField.Input that was bothering me and included it here.

PR #1006, opened 2021-12-01
trapped focus and ChoiceGroup fixes

jakubjanczyk:
# Problem/Feature
In this PR two things has been change, all related to issues found in Messages app: https://github.com/helpscout/hs-app-ui/pull/209

## Trapped focus
There was an issue that if container was focused, for example SimpleModal, and user pressed Shift+Tab, the focus was moved to the previous element in the DOM, not witihn the container, even if `trapFocus` was set to true.
The fix here is to always move within container, even if it is the one focused.

To test this, I have create `SimpleModal` -> `with trigger` story that would open and focus modal once the button is clicked.

## ChoiceGroup single select 
If ChoiceGroup was used with `multiSelect={false}` and children are `Checkboxes` (or variant of them) - then clicking the same item twice was de-selecting on the UI, but clicked third time was not selecting it again. The reason being that we were always returning the clicked value from `onChange` in `ChoiceGroup` component, no matter what was the `checked` argument. The change here is to include the `checked` value.

It only affects usage with Checkbox, as Radio would not call `onChange` if clicked again on the same element (it is not possible to de-select radio button)

In the case I am fixing here, there were `CheckMarkCard` items used.

To test this, go to `CheckMarkCard` -> `inside group` story and change `Knobs` to have `multiSelectLimit = 1` and `multiSelect` to not be checked.

`RadioCard` -> `inside a group` story can be used to verify the change does not break if `Radio` is used.

## CheckMarkCard accessibility fix
Before the fix, the CheckMarkCrad had multiple labels that was causing heading to be read twice (or even 3 times with avatar). I have changed to remove label prop from the Checkbox itself, because it's already being labelled by `label` element and `htmlFor` prop is used. I've also hidden avatar from the screen readers as it was also containing a name.

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5596,9 +5596,9 @@
       }
     },
     "@testing-library/jest-dom": {
-      "version": "5.12.0",
-      "resolved": "https://registry.npmjs.org/@testing-library/jest-dom/-/jest-dom-5.12.0.tgz",
-      "integrity": "sha512-N9Y82b2Z3j6wzIoAqajlKVF1Zt7sOH0pPee0sUHXHc5cv2Fdn23r+vpWm0MBBoGJtPOly5+Bdx1lnc3CD+A+ow==",
+      "version": "5.15.1",
+      "resolved": "https://registry.npmjs.org/@testing-library/jest-dom/-/jest-dom-5.15.1.tgz",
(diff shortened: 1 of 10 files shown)


tinkertrain (approved):
⭐ Nice work, thanks for the fixes!

PR #1008, opened 2021-12-03
Call onLoad for Avatar in constructor, when image already in cache

jakubjanczyk:
# Problem/Feature
This PR improves Avatar to call `onLoad` callback in constructor for case that image is already in cache (so regular onLoad after fetching image would not be fired).
The use case for that is in Messages app where we need to perform some action when image is loaded and can be rendered. 

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests


tinkertrain (approved):
🚢

PR #1010, opened 2021-12-06
New CopyValue Component

plbabin:
# Feature
CleanShot 2021-12-06 at 13 59 36

This update brings the copy conversation id (from hs-app-ui) into a new HSDS component. This behavior will be reused throughout the application so we thought it was a good idea to create a component flexible enough to be customize when needed but also strict enough to keep the behavior the same across the app. 

The component will receive a `value` prop. When the button is clicked, the value will be added to the clipboard. It it still possible for dev to extend the `onClick` action. 

```javascript
// when the user click on the icon, "123456789" will be added to the clipboard
<CopyValue value="123456789" />
```

### Props
| prop name     |  |
|----------|---------|
| prefix    | will add  a muted prefix before the value  |
| renderValue |  custom render for the value    |
| onReset | Callback when the indicator has fade out    |
| value | the text that will be copied to the clipboard |
| resetTimeout | how long the indicator will be visible, default to 1000ms |

### `--HSDSGlobalFontFamilySystem`

We started to see a lot more application of a system font to some component. We decided to extract that to an HSDS css variable to make it easier to use across HSDS and hs-app.

```javascript
export const ValueUI = styled(Text)`
  font-family: var(--HSDSGlobalFontFamilySystem);
  color: ${getColor('charcoal.400')};
`
```

the variable value is `-apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"`

### DIFF
--- a/src/components/CopyButton/CopyButton.jsx
+++ b/src/components/CopyButton/CopyButton.jsx
@@ -1,9 +1,12 @@
-import React, { useState, useRef, useEffect } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 import getValidProps from '@helpscout/react-utils/dist/getValidProps'
 import Icon from '../Icon'
 import classNames from 'classnames'
 import { noop } from '../../utilities/other'
+
+import { useCopyConfirmation } from './CopyButton.utils'
+
 import {
   CopyButtonUI,
   ConfirmationIconWrapperUI,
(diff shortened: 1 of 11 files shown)


justinwolfe (approved):
This looks great, @plbabin! I think the design of the component/prop choices are really smart in terms of making this plug and play but also allowing for custom render values and callbacks if we need them (also yes re: a reusable CSS variable for that long font string!). I tested it out locally and it works/looks as I think it should 🙌 .

digitaldesigner (changes requested):
Only one tiny tiny change is that it'd be great to make this a perfect circle (looked slightly wider than intended from the GIF). Whatever height it is, could you match that for the width?

plbabin:
> Only one tiny tiny change is that it'd be great to make this a perfect circle (looked slightly wider than intended from the GIF). Whatever height it is, could you match that for the width?

should be ok now!

digitaldesigner (approved):

tinkertrain (approved):
Looks good, thanks! 🚀

PR #1012, opened 2022-01-05
AvatarRow: Show a minimum amount of avatars before folding (minAvatarsShown)

tinkertrain:
## Feature: Show a minimum amount of avatars before folding

Currently `AvatarRow` always has minimum width to allow for 1 Avatar. Sometimes we might need to make sure we are always showing at least a different amount, independently of the parent's width.

This is mostly of an edge case, you might need this if the parent's width is dynamic. To illustrate, an example:

Conditions:
- The container's width adapts to its content (like Popover)
- The number of Avatars change over time
 
1. We start with 1 Avatar, so AvatarRow's width is of 1 Avatar
2. Popover extends its width to allow 1 Avatar
3. Before we activate the Popover, another avatar is loaded
4. AvatarRow checks the available width, since Popover only grew enough for 1 Avatar, AvatarRow "folds" the next avatar (and subsequent) into the space, thus showing the circle with the number of Avatars.
5. Activate the popover and see:

Screen Shot 2022-01-12 at 11 05 20 AM

Using `minAvatarsShown`, we make sure AvatarRow is always the required width for N number of avatars, say in that Popover you want space for minimum of 4, setting `minAvatarsShown = 4`. With that Popover will see that the content and grow its width accordingly.

Screen Shot 2022-01-12 at 11 07 37 AM

---

- Why not setting a min-width on the parent? 
Because in this case the number of avatars is unknown at any given point so you would have either too much space or too little.

Screen Shot 2022-01-12 at 11 13 30 AM

--- 

See the [story](https://0ba4c4bb.hsds-react.pages.dev/?path=/story/components-structural-table--selecting-rows) for a full example that dynamically adjusts it all based on thu current number of avatars loaded.


src/utilities/specs/avatar.specs.js
```diff
@@ -13,14 +13,14 @@ const avatarSpec = createSpec({
         id: prop.id,
         name: prop.name,
         firstName: prop.firstName,
-        image: prop.image,
+        // image: prop.image,
```
cen10, line 16: Leftover comment?

src/utilities/specs/avatar.specs.js
```diff
@@ -13,14 +13,14 @@ const avatarSpec = createSpec({
...
       }
     })
   }
   return {
     id: props.id,
     name: props.name,
-    image: props.image,
+    // image: props.image,
```
cen10, line 23: Leftover comment as well?

cen10 (approved):
Amazing, thank you! I think there are two commented out lines that can be removed, otherwise looks great!

src/utilities/specs/avatar.specs.js
tinkertrain: Was waiting for the button PR to be merged (done now), avatar images stopped working and PL switched to another service in that PR 👍

PR #1013, opened 2022-01-06
HSDS-229: MessageCard body variable transformer

jakubjanczyk:
[JIRA](https://helpscout.atlassian.net/browse/HSDS-229)

# Problem/Feature
This PR adds support for highlighting variables in a MessageCard. Variables are provided via props, so this is not connected to any specific available variables, but whatever would be needed.

Design: https://www.figma.com/file/H2oEYNwn2RMpSVSR94B2vB/Spec%2F-Variable-Support-in-Messages?node-id=34%3A6520

## Taken path
I initially though about implementing just some utility function that is exported and can be used whenever needed. Main reason for this being re-usability, customization (can use different transformation if necessary) and the fact that it is not needed in Beacon Embed, so to not add additional size. 
However, I thought about it again and decided to, at least for now, just include implementation inside MessageCard component. Reason for that being:
- There is no customization I can think of right now that we might need
- The implementation doesn't add a lot of code, so size impact on Beacon would not be significant
- it wouldn't need to be remembered to use transformer in multiple places and if we need to modify implementation, it can be done in one place instead of multiple (like adding variables in heading, in the future)
The transformation function is separated, so at any point we could just change the approach if this one wouldn't work.

## Solution
For the main part - transforming the variables - I have used RegExp to find all occurences of variables and replace them with span element, that can then be styled accordingly. Only those variables that are provided in a separate list would be replaced. This prevents unknown variables provided by hand (or just text looking as if it's variable) from being highlighted.

The replacement span get's defined class name that is then used in MessageCard component for styling, according to the design. I've exported this to a variable so it's easy to change it if necessary.

One more thing I have done here was to re-write MessageCard tests from `enzyme` to `testing-library` - they check the same thing, just in a modern way :)

## Testing
I've added a story `Message Card -> With variables` that can be used for visual testing.

Screenshot from 2022-01-06 18-57-18

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -292,6 +292,7 @@ export const BodyUI = styled.div`
     font-style: normal;
     font-weight: normal;
     text-decoration: none;
+    white-space: nowrap;
   }
 `
 
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -281,6 +281,7 @@ export const BodyUI = styled.div`
     display: inline-flex;
     align-items: center;
     padding: 3px 8px;
+    margin: 0 4px;
     height: 20px;
     line-height: 17px;
 
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -281,7 +281,7 @@ export const BodyUI = styled.div`
     display: inline-flex;
     align-items: center;
     padding: 3px 8px;
-    margin: 0 4px;
+    margin-right: 4px;
     height: 20px;
     line-height: 17px;
 
--- a/src/components/MessageCard/MessageCard.css.js
+++ b/src/components/MessageCard/MessageCard.css.js
@@ -8,6 +8,7 @@ import Button from '../Button'
 import Heading from '../Heading'
 import Image from '../Image'
 import ArticleCard from '../ArticleCard'
+import { messageVariableClassName } from './utils/MessageCard.utils'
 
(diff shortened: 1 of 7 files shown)


Charca (commented):
Looks good to me! Great work with that Regex @jakubjanczyk!

I just have one question that is mostly from my lack of familiarity with this project: when the variable is replaced by the actual value (when the Message is shown to the real user), should the value be wrapped in the styled span, or should it show up as plain text?

So instead of:

Hi, `Jakub`!

Should it show up as:

Hi, Jakub!

If yes, then should the MessageCard component be aware of this distinction?

jakubjanczyk:
@Charca Real Message would never get variables to replace - it would be automatically done one the API side.
So if we have message with body `Hi {%customer.firstName,fallback=there%}` in DB, the real Message would already get `Hi Jakub`. And if someone used variable that does not exist, then we just display it as a plain text without any replacement. 
Additonally, for Messages in Beacon Embed, we wouldn't be providing `variables` prop, so there would be nothing to replace :) I was considering if maybe we should add like `preview` prop that would indicate if it's a preview or not (obviously) and as a result if we should do the replacement, but I thought just using `variables` and the fact that API is doing the replacement would be enough.

plbabin (approved):
thank you for taking care of that change. Looks good to me 🥇🥇🥇

alexestrada (commented):
This looks good to me. Any chance we might be able to add a space before/after the variable? Just so the tag isn't bumping up against the text. I'm thinking ~4px.

jakubjanczyk:
@alexestrada please take a look now :)

alexestrada (approved):
Looks good! Thank you!

PR #1014, opened 2022-01-17
Allow Collabsible/Accordion.Body content to be pre-rendered

jakubjanczyk:
# Problem/Feature
For one case in Messages app we'd like to use Accordion component for it's functionality. However, we need the content of the Accordion Body to be rendered, because some background operations need to be performed, even if Accordion is closed. Currently the content is conditionally not rendered in `Collapsible` component (until opened)

This PR adds a new flags to `Accordion.Body` and `Collapsible` components so that the content would be rendered to the DOM. It is still hidden from user, but React lifecycle would fire for them.

I've not added any story for that, but there are tests confirming it works as expected.

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/Accordion/Accordion.Body.jsx
+++ b/src/components/Accordion/Accordion.Body.jsx
@@ -52,7 +52,7 @@ const getComponentClassName = ({
 }
 
 export const AccordionBody = props => {
-  const { className } = props
+  const { className, preRender } = props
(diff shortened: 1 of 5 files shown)


tinkertrain (approved):
Looks good :shipit:

PR #1016, opened 2022-01-18
Fix/hsappui-306

tinkertrain:
## Table

When setting up a sortable table, header cells currently are only click-actionable. In this PR we switch the `div` for a  `button` for them to become actionable with a keyboard.

Sorting the table is most likely going to be an async task, and when triggering it with the keyboard the UI state gets lost: in this case we care about the focus state. To make it a better UX, we can now return a Promise on the `sorter` function and we'll take this cue to refocus the button.

Trying it out is better than trying to explain it: https://dad6d9e3.hsds-react.pages.dev/iframe?id=components-structural-table--with-sorting&args=&viewMode=story

cc @digitaldesigner, what would we want as a focus style? Currently (on firefox):

Screen Shot 2022-01-18 at 4 54 53 PM

## DropList

Downshift automatically adds a `aria-labelledby` attribute to the menu dropdown, but we don't use labels in DropList, making it always point to an inexistent element. We add a `menuAriaLabel` prop to set a custom `aria-label` directly on the menu and we remove the `aria-labelledby` set by downshift

### DIFF
--- a/src/components/DropList/DropList.Combobox.jsx
+++ b/src/components/DropList/DropList.Combobox.jsx
@@ -35,6 +35,7 @@ function Combobox({
   inputPlaceholder = 'Search',
   items = [],
   isOpen = false,
+  menuAriaLabel,
   menuCSS,
   menuWidth,
   onDropListLeave = noop,
@@ -238,6 +239,8 @@ function Combobox({
       <MenuListUI
         className={`${DROPLIST_MENULIST} MenuList-Combobox`}
         {...getMenuProps()}
+        aria-label={menuAriaLabel}
+        aria-labelledby={null}
       >
         {renderListContents({
           customEmptyList,
--- a/src/components/DropList/DropList.Select.jsx
+++ b/src/components/DropList/DropList.Select.jsx
@@ -27,6 +27,7 @@ function Select({
   handleSelectedItemChange = noop,
   isOpen = false,
   items = [],
+  menuAriaLabel,
   menuCSS,
   menuWidth,
   focusToggler = noop,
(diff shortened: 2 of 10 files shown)


digitaldesigner:
@tinkertrain The focus should be the blue border on the left of the row, see https://www.figma.com/file/8rXMLTfIx9RTkihPhHU5B4/Spec%2F-Convo-Table-Rebuild?node-id=225%3A10261 (which is what we do today 🙂 )

tinkertrain:
@digitaldesigner Thanks for looking at this, I meant the focus on the table header cells (the dotted outline in this screenshot) when they are actionable (sorting):
Screen Shot 2022-01-18 at 4 54 53 PM

tinkertrain:
@digitaldesigner Thanks for looking at this, I meant the focus on the table header cells (the dotted outline in this screenshot) when they are actionable (sorting):
Screen Shot 2022-01-18 at 4 54 53 PM

Edit: One more thing, PL noticed that the header row height is a bit wonky when it spans to 2 lines...
CleanShot 2022-01-19 at 09 33 24@2x
In that screenshot you can see that 2 lines don't fit in the current height of the header row... do I make it taller? Or adjust line height? You can also see in that screenshot PL added the current focus style of buttons...

digitaldesigner:
Oh right sorry mate, I reckon we use the 2px blue.500 border we use elsewhere — but no background color change (unless they are hovering AND focussing: 

Screen Shot 2022-01-21 at 2 30 48 PM

tinkertrain:
Changes folded into https://github.com/helpscout/hsds-react/pull/1017

PR #1021, opened 2022-01-31
Adds animated rows to Table

tinkertrain:
# Table: animated addition/removal of rows

In this PR we add the ability of animating the removal and addition of rows on `<Table>`.

Using `react-transition-group` to handle the animation, we add a new prop `animateRows`. If `true` is passed, it enables the animation with default settings and a default animation, it is still possible to customize the actual animation by overriding the css that is applied on the `<tbody>`.

Default class names and css:

```css
.animated-row-enter {
  opacity: 0;
}

.animated-row-enter-active {
  opacity: 1;
  transition: opacity 500ms ease-in;
}

.animated-row-exit {
  opacity: 1;
}

.animated-row-exit-active {
  opacity: 0;
  transition: opacity 500ms ease-in;
}
```

CleanShot 2022-01-31 at 17 20 41

Example override:

```js
const TableUI = styled(Table)`
  .TableBody--animated {
    .animated-row-enter {
      background-color: red !important;
    }

    .animated-row-enter-active {
      background-color: transparent !important;
      transition: background-color 500ms ease-in;
    }

    .animated-row-exit {
      background-color: transparent !important;
    }

    .animated-row-exit-active {
      background-color: red !important;
      transition: background-color 500ms ease-in;
    }
  }
`
```

CleanShot 2022-01-31 at 17 46 53

If you need more control over the animation, pass an object to `animateRows` with any valid prop for [CSSTransition](https://reactcommunity.org/react-transition-group/css-transition) and [Transition](https://reactcommunity.org/react-transition-group/transition).

For example:

```jsx
<Table animateRows={{ timeout: 300, onExit: () => console.log('Exit!') }} />
```

## Tests

I found quite difficult and a bit pointless to test (mocking `react-transition-group` for example) so instead the story I created tests different scenarios.

# Extra

## AvatarRow

There was an issue where sometimes when dynamically loading avatars and the container width was also dynamic we got into a corrupt state and show all avatars collapsed even when there seemed to be enough space to show more than one.

The problem was that calculating how many avatars would fit depended on an item coming from state, and sometimes that got out of sync, to solve this we change the math and now we don't depend on state, making the calculation less brittle.


src/components/Table/Table.jsx
```diff
@@ -211,6 +213,8 @@ export function Table({
 }
 
 Table.propTypes = {
+  /** Enable animation for row addition / removal. */
+  animateRows: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
```
knicklabs, line 217: Instead of overloading this prop to both enable and configure the animation, what do you think about splitting it out into two props, one to toggle animation on/off and one to configure animation? I think it might lead to a more robust design, particularly if we get into situations where we want to respect reduced motion settings for accessibility.

knicklabs (approved):
LGTM! I don't think it's a blocker, but I have a question about the component API that I left for you.

src/components/Table/Table.jsx
tinkertrain: Thanks for the input! I went with this approach because I really dislike adding tons and tons of props and end up with something like `Input`. Whenever I can merge 2 props into 1, I go for it 😁 (if it makes sense of course!)

It's very likely the animation will be customized somehow almost always if only to adjust the `timeout`, or get access to `onEnter` or `onExit`, so I see no reason to make folks also having to enable the thing with another prop.

And about reduced motion settings, you'll achieve that by customizing the css with `prefers-reduced-motion: no-preference`, something like:

```css
const TableUI = styled(Table)`
  .TableBody--animated {
	  .animated-row-exit {
	    background-color: transparent;
	  }
	  
	  .animated-row-exit-active {
	    background-color: red;
	    transition: background-color 500ms ease-in;
	  }
	  
	  @media screen and (prefers-reduced-motion: no-preference) {
	      .animated-row-exit {
	        background-color: transparent;
	      }
	  
	      .animated-row-exit-active {
	        background-color: red;
	        transition: none;
	      }
	   }
	}
`
```

knicklabs (approved):

PR #1027, opened 2022-02-08
[HSDS4-30] HSDS 4: prep work

tinkertrain:
https://helpscout.atlassian.net/browse/HSDS4-30

Preparing the ground to make migration easier by reducing superfluous inter-dependencies.

- Removes `noop` dependency from components
	- The dependency on a utility that does pretty much nothing just complicates the dependency graph needlessly (creating noop functions is fast, it won't have a real impact)
	- ~If there's more than 2 `() => undefined` declarations in a component, we add a `const noop = () => undefined` at the top of the component file, otherwise, declaration is put in place.~ / After PL's note, decided to just declare a function.
-  Removes `Centralize` as dependency for other HSDS components
	- This is a silly component that just applies some flexbox styles, declaring them in place reduces complexity.
	- We don't remove the component for now, there is one use in HS App and a few more in Beacon, will declare something in those repos at another time.
- ChatList: Removes Hr dependency
- styled: removes unused testHelpers and tests. Make sure to import directly from `styled-components` in HSDS
	- Won't remove until we update HSApp and the rest to do the same
- Tooltip: Remove `GlobalContext`
	- We only have 2 scopes: `hsds-react` and `hsds-beacon`, hard coding them to remove the dependency seems like a good trade off, adding both doesn't impact negatively either.
- Accordion: removes `GlobalContext` dependency. Same as above.
- Remove `ScopeProvider` from Messages stories

### Notes
- The `noop` change is the most superficial but it made this PR noisy, the best way to see the other changes is looking at each commit, I tried to be organized about it.
- Making these changes diminished the testing coverage below what we currently accept for some reason (makes you wonder how effective this metric/tool is uh?), so I lowered the thresholds to 90% across the board. In HSDS 4 this won't matter anyway as we have several components way below those thresholds and is only because there's an average that it passes.


plbabin (approved):
this is an amazing cleanup PR. well done!

One thing I am not 100% convince, is the noop change. I agree that it will improve the dependency graph, but we are creating a new function on each component call. Not really afraid of memory management, but more about unnecessary renders since the noop reference would have change. Unless babel is smart and memoize default function param. So anyway, not a big deal, just something I wanted to point out

tinkertrain:
> this is an amazing cleanup PR. well done!
> 
> One thing I am not 100% convince, is the noop change. I agree that it will improve the dependency graph, but we are creating a new function on each component call. Not really afraid of memory management, but more about unnecessary renders since the noop reference would have change. Unless babel is smart and memoize default function param. So anyway, not a big deal, just something I wanted to point out

Thanks for pointing this out, I don't think it would create a function every time if the `noop` function is declared outside the component definition (and this might not be what you are referring to), but the declaration in place might (I think this is what you are referring to) so I just made an update to declare the function anytime `noop` is used.

plbabin:
@tinkertrain yes that was exactly what I was thinking, creating them outside of the default parameters is perfect, thank you for making the change!

PR #1036, opened 2022-02-21
FancyAnimationScroller hook improvements

tinkertrain:
## Problem

Previously `FancyAnimationScroller` required fixed heights to animate the header: "from" and "to". This prevented having a header that could change in size.

## Solution

Use a `ResizeObserver` on the header to detect the "from" height of the header instead of it being fixed. This way it can change at any time and the hook will know what size to use.

We also improve the signature of the hook by removing nested parameters and renaming arguments. It's all documented [here](https://04cb6d1a.hsds-react.pages.dev/?path=/story/hooks-usefancyanimationscroller--page).

See this [ScrollableContainer story](https://04cb6d1a.hsds-react.pages.dev/?path=/story/components-structural-scrollablecontainer--full-example) to see an example of its usage.


tinkertrain:
@luketlancaster It was already reviewed by design, this work doesn't change the interaction only allows you to have a dynamic initial height on the header 👍

PR #1037, opened 2022-02-22
[HSDS4-32][HSDS4-15] Processing utils, pt 2

tinkertrain:
Continue the processing of utilities inside HSDS. 

https://helpscout.atlassian.net/browse/HSDS4-32
https://helpscout.atlassian.net/browse/HSDS4-15

> Should merge https://github.com/helpscout/hsds-react/pull/1031 first ✅ 

After analyzing the usage of utilities in:
- HSDS
- HS App
- HS App UI
- Beacon
- Editor

[utilities.pdf](https://github.com/helpscout/hsds-react/files/8116105/Utils.-.hsds.-.src.-.utilities.pdf)

This PR makes the following changes:

> Again, I tried to be very clean with commits so one change is one commit for the most part

#### utilities/actions.js
Removed. Only used in the old Dropdown, incorporated there.

#### utilities/animation.js
Removed. Integrated `getSequenceNames` into Animate

#### utilities/browser.js
Removed. Only used internally twice, so moved there (`isFirefox` to Overflow, `isSafari` to ScrollLock)

#### utilities/check.js
Removed. Moved `allDefined` to ChatScroller.utils.js. Rest unused

#### utilities/classNames.js
- Use classNames package inside HSDS
- Left `classNames` as is used outside of HSDS
- Move `variantClassNames` into Grid.Col

#### utilities/collection.js
Removed. Not used.

#### utilities/debounce.js
Removed. Use lodash.debounce instead (already a dependency)

#### utilities/env.js
Removed. Not used.

#### utilities/events.js
- Removed. 
- Deleted `addEventListener` & `removeEventListener` and just used event listener declaration in place.
- Moved `isMouseWheelYEvent` into Overflow

#### utilities/get.js
Removed. Use lodash.get instead (already a dependency)

#### utilities/globalManager.js
Removed. Integrated `setupManager` into PortalWrapper

#### utilities/history.js
Removed. Incorporate `createLocation` into RouteWrapper.

#### utilities/Images.js
Removed. Incorporate `calculateAspectRatioFit` into Image.utils.js

#### utilities/memoize.js and utilities/memoizeOne.js
Removed. Only used in the old Dropdown, incorporated there.

#### utilities/nodePosition.js
Removed. Incorporated into Drop.Positioner

#### utilities/other.js
Removed `requestAnimationFrame` shim, reason for it no longer exists (JSDOM incorporates it)

#### utilities/pluralize.js
- Incorporate it into Pagination
- Left as it's used in HS App once (ChatSidebar)

#### utilities/scrolling.js
Incorporate `remapScrollingPlane` into Overflow

#### utilities/storybook.js
Removed `withAktiv` from stories

#### utilities/strings.js
- Incorporated `repeat` into Input.Resizer.jsx (unfortunately `String.repeat` is not supported in IE). Lef in place because Beacon uses it once. Should declare there
- Incorporated `wordHasSpaces` into Link.jsx
- Incorporated `stripUrlPrefix` into Notification.jsx
- Incorporated `truncateMiddle` into Truncate.jsx

#### utilities/timestamp.js
Removed. Incorporated into Timestamp

#### utilities/react-router/matchPath.js
Removed. Incorporated into PortalWrapper

## Test Coverage
Again, percentage dropped even though no tests of actually used utils were deleted, but moved. Things like the utils I moved into the old Dropdown were artificially upping the percentage. Lowered to 85%

## Apps test
Released a beta and tested in:
- ✅ HS App
- ✅ Beacon
- ✅ hsapp-ui


src/components/ScrollLock/ScrollLock.jsx
```diff
@@ -2,7 +2,13 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import ReactDOM from 'react-dom'
 import { handleWheelEvent } from './ScrollLock.utils'
-import { isSafari } from '../../utilities/browser'
+
+const isSafari = () => {
+  if (!navigator) return false
+  const isChrome = navigator.userAgent.toLowerCase().indexOf('chrome') > -1
```
plbabin, line 8: could be replace with `.includes('chrome')`

src/components/Image/Image.utils.js
```diff
@@ -1,5 +1,24 @@
...
+
+  const ratioWidth = maxWidth / width
+  const ratioHeight = maxHeight / height
+  const ratio = Math.min(ratioWidth, ratioHeight)
+
+  return {
+    width: width * ratio,
+    height: height * ratio,
+  }
+}
```
plbabin, line 21: I do wonder if all that calculation could be replace with `aspect-ratio: 1`

plbabin (approved):
Moving back shared utils to the component they belong to is such a brilliant and elegant solution. Well done!

I left two comments, nothing major though

src/components/ScrollLock/ScrollLock.jsx
tinkertrain: 👍

PR #1038, opened 2022-02-22
Add tiny icons for line items

justinwolfe:
# Problem/Feature

For [line item rows in the new conversation view design](https://paper.dropbox.com/doc/Spec-Convo-Content-inc.-Threads--BcfPdGbEtmca2tfROF114rHKAg-SdORCzUg4Zu45aBO6wreY), we needed to add two new icons because (per design) the size of the current icons is not appropriate for use in that context.

So this PR adds those two new icons from @kyliedunkley! 🥳  

CleanShot 2022-02-22 at 11 18 23
CleanShot 2022-02-22 at 11 18 53

**Note**: It was stressed to Kylie that these should be "tiny icons"™. I couldn't find any references to what "tiny icons" means searching in various places and this is the only `Icon` component I know about (and its `size` prop is number based rather than `small` `medium` `large` etc), so I assume these go here and maybe that was just some design hazing™, but if I should be putting them somewhere else or doing something different, let me know!

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests


plbabin (approved):
I would add two aliases to `user-small`
- `assign-small`
- `person-small`

Also, since you are in that part of the code, could you add an alias `assign` -> `person`. that would make sense to have all look-alike icon in at least one namespace

After that, it is ready to go, thank you :)

justinwolfe:
💯  will do, thank you!

PR #1042, opened 2022-03-03
Allow table columns to be changed

knicklabs:
# Problem/Feature

It is currently not possible to dynamically change the columns of a table because the state does not update when the props change. This small change ensures that the columns state stays in sync with the columns prop.

This was needed to support changing Mailbox views. When we change views, we load new conversations that get loaded into the table as rows; however we also need to change the columns to show based on the view. See https://github.com/helpscout/hs-app-ui/pull/319 for more context.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -82,6 +82,10 @@ export function Table({
     updateTableData(data, maxRowsToDisplay)
   }, [data, maxRowsToDisplay, sortedInfo])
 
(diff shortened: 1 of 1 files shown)


knicklabs:
@tinkertrain I'm not overly familiar with the table component and while I opened a PR with a proposed change, I'm totally up for working with you to figure out the best way to move forward.

tinkertrain (approved):
This works great, thanks! 💯

PR #1045, opened 2022-03-14
DropList combobox new features HSDS-252

tinkertrain:
Story: https://b87ad319.hsds-react.pages.dev/?path=/story/components-dropdowns-droplist--with-action-item

## Emphasize string on items with input value

Screen Shot 2022-03-14 at 2 47 46 PM

## Filtering items

1. When filtering, do not remove the action item
  - Divider: keep it if there are items left (besides the action), remove it if only the action left
Screen Shot 2022-03-14 at 2 48 18 PM
Screen Shot 2022-03-14 at 2 48 28 PM

2. Leave groups if not empty after filtering

Screen Shot 2022-03-14 at 2 49 01 PM
Screen Shot 2022-03-14 at 2 49 08 PM

## Action item

New: 

1. Add a `template` key to your action item if you would like update the action item label with the input value, the special placeholder `__inputValue__` will be replaced with the input value.

```jsx
const actionItem = {
    label: 'Add tag',
    template: 'Add __inputValue__ tag',
    type: 'action',
    action: 'ADD_TAG',
  }

// ....

<DropList
  onSelect={(selection, clickedItem) => {
    if (clickedItem.type === 'action') {
      console.log('add tag clicked, do something!')
      // `clickedItem` will include the updated label and the input value in `clickedItem.inputValue`
    }
  }}
/>
```

Screen Shot 2022-03-14 at 2 55 59 PM

Screen Shot 2022-03-14 at 2 51 59 PM

2. Add a `hideOnEmptyInputValue` key to any item if you want them to be hidden when the input of a combobox is empty:

```jsx
<DropList
        variant="combobox"
        items={regularItems
          .slice(0, 3)
          .concat({
            type: 'divider',
            hideOnEmptyInputValue: true,
          })
          .concat({
            label: 'Add tag',
            template: 'Add __inputValue__ tag',
            type: 'action',
            action: 'ADD_TAG',
            hideOnEmptyInputValue: true,
          })} 
/>
```

Screen Shot 2022-03-15 at 5 23 26 PM

Screen Shot 2022-03-15 at 5 23 36 PM

## Filtered items on `onInputChange`

When filtering, `onInputChange` will return both the input value and the resulting filtered items

### DIFF
--- a/src/components/DropList/DropList.ListItem.jsx
+++ b/src/components/DropList/DropList.ListItem.jsx
@@ -9,6 +9,7 @@ import {
   isItemAction,
   isItemADivider,
   isItemAGroupLabel,
+  emphasizeSubstring,
   objectHasKey,
 } from './DropList.utils'
 import {
@@ -29,6 +30,7 @@ const ListItem = forwardRef(
       isSelected,
       renderCustomListItem,
       isDisabled,
+      inputValue,
       ...itemProps
     },
     ref
@@ -92,9 +94,13 @@ const ListItem = forwardRef(
         withMultipleSelection={withMultipleSelection}
         {...itemProps}
       >
-        <ListItemTextUI>
-          {isPlainObject(item) ? item[contentKey] : item}
-        </ListItemTextUI>
+        <ListItemTextUI
+          dangerouslySetInnerHTML={getListItemText({
+            label: isPlainObject(item) ? item[contentKey] : item,
+            emphasize: inputValue && !isItemAction(item),
+            substr: inputValue,
+          })}
+        />
         {withMultipleSelection && !isItemAction(item) ? (
           <SelectedBadge isSelected={isSelected} />
         ) : null}
@@ -102,6 +108,17 @@ const ListItem = forwardRef(
     )
   }
 )
+
+/**
+ * If `substr` is passed, we want to emphasize that with html's <strong />
+ */
+export function getListItemText({ label, emphasize, substr }) {
+  const __html = emphasize ? emphasizeSubstring(label, substr) : label
+
+  return {
+    __html,
+  }
+}
 
 export function generateListItemKey(item, index) {
   const contentKey = getItemContentKeyName(item)
--- a/src/components/DropList/DropList.ListItem.jsx
+++ b/src/components/DropList/DropList.ListItem.jsx
@@ -35,6 +35,10 @@ const ListItem = forwardRef(
     },
     ref
   ) => {
+    if (!inputValue && item.hideOnEmptyInputValue) {
+      return null
+    }
+
     if (isItemADivider(item)) {
       return (
         <DividerUI className="DropListItem--divider" key={`divider_${index}`} />
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.39.2",
+  "version": "3.39.3-0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.39.2",
+  "version": "3.39.3-0",
   "private": false,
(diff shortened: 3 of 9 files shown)


src/components/DropList/DropList.ListItem.jsx
```diff
@@ -103,6 +109,17 @@ const ListItem = forwardRef(
...
+ * If `substr` is passed, we want to emphasize that with html's <strong />
+ */
+export function getListItemText({ label, emphasize, substr }) {
+  const __html = emphasize ? emphasizeSubstring(label, substr) : label
+
+  return {
+    __html,
+  }
+}
+
```
justinwolfe, line 122: Out of curiosity, what's the reason for creating an HTML string and using `dangerouslySetInnerHTML` vs. creating and passing React children to the `ListItemTextUI` component? 

If the answer is "it was faster and easier to write the utils for doing this that way!" that is totally an acceptable answer to me, just curious!

justinwolfe (approved):
@tinkertrain Awesome and quick work here, you are my hero 🦸‍♂️ !! In particular, the approach you've come to on the action item design is so much more elegant than the direction I was thinking in 😻 . I think this will definitely meet our needs—just a couple of notes/questions:

- It looks like there are some small rough edges around what happens when the user is removing/deleting characters from the combobox input.

https://i.hlp.sc/xCcClM/ (action item label persisting input value after input is cleared)
https://i.hlp.sc/BIvGgA/ (inconsistent filtering behavior when deleting characters from input)

- A question for @digitaldesigner—does the design for the state in which there are no matches and so just the action item is present work okay for you? 

CleanShot 2022-03-14 at 09 19 53

When items are present in the list, the divider serves as a kind of signal that the action item is not a normal list item:

CleanShot 2022-03-14 at 09 26 27

but the changes in this PR remove the divider if only the action item is left. That makes sense (what is there to divide!) and maybe is fine but I wanted to double check on whether that erasure of the conceptual distinction between the two things is okay and you would want to either keep the divider or differentiate the action item design in some other way.

CleanShot 2022-03-14 at 09 24 07

Per the tags menu in the rewrite designs, we currently change the styling of the action item:

CleanShot 2022-03-14 at 09 33 39

So not doing any styling in the HSDS component and allowing individual usages to override that might be the most sensible default if we think it's appropriate to approach this design differently in different contexts, but wanted to check on whether you'd rather have the default baked into the component be different.

- Another one for @digitaldesigner. Minor edge case, so we could just punt, but along the same lines as the previous question, while we're here, do we want to take this opportunity to HSDS-ize removing the dropdown entirely if there are:

1. No items 
2. No input value (which means nothing to use the action item to create something from the input value with).

For the mailbox rewrite Tags Menu, per the designs, when the above conditions are true, we're overriding the dropdown with CSS to hide the empty list item, which works just fine, but wanted to check whether you think that should be the canonical behavior rather than an override.

CleanShot 2022-03-14 at 09 56 03

src/components/DropList/DropList.ListItem.jsx
tinkertrain: no reason, different ways to achieve the same things. I like to live dangerously every now and then

justinwolfe:
@tinkertrain 😹  That reasoning makes sense to me! Glitch fixes look good, except (and apologies for not articulating this more clearly before) in the state where there's no input value, shouldn't the action item be gone? 

CleanShot 2022-03-15 at 08 53 25

In my use case, at least, without input value, there's nothing to create with the action item. Are there other places where that might not be true and so we'd want to preserve the behavior? If so, I can override in my component and this is G2G, but wanted to check!

tinkertrain:
@justinwolfe don't remember exactly how the tags workflow is, is it a modal to create the tag? or is it just instantly created as soon as you choose the option? my assumption was that you can create a tag always and just the name input being autofilled with the input value (which could be an empty string).

Regardless, if you want to hide it, you might have to do it on your end (removing it by default and using `onInputChange` to add it to the items) because I can see the case of persisting the action item more often, for example a "reset defaults" item:
 
Screen Shot 2022-03-15 at 5 05 26 PM

PR #1047, opened 2022-03-23
Add custom remove label option to Tag component

justinwolfe:
# Problem/Feature

In the beautiful CC/BCC implementation that @plbabin did for the mailbox rewrite, we're using the Tag component as a badge to represent the email address. [During accessibility review](https://github.com/helpscout/hs-app-ui/pull/351), we found it would be best to be able to customize the label, since it reads "Remove Tag" but in this context there isn't a tag. This adds a prop to be able to set the aria-label and test ID of the removal element to a different value:

CleanShot 2022-03-23 at 14 28 43

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests
- [x] Did you verify some accessibility (a11y) basics?

### DIFF
--- a/src/components/Tag/Tag.jsx
+++ b/src/components/Tag/Tag.jsx
@@ -32,6 +32,7 @@ const ForwardedTag = forwardRef(function Tag(props, ref) {
     className,
     color,
     count,
+    customRemoveLabel,
     display,
     filled,
     id,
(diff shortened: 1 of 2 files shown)


justinwolfe:
@plbabin had the great idea, rather than just passing a custom label, to pass in a `removeProps` object so that we could have flexibility to pass additional props, so I've switched to that!

plbabin (approved):

PR #1050, opened 2022-04-14
Delay PortalWrapper content render

jakubjanczyk:
# Problem/Feature

We've started using Preact in Beacon Embed for lighter bundle size and better performance. It's in most part compatible with React, so a lot of things work the same. However, there apparently is one difference we've noticed.
If there is a Portal created from within another Portal, the rendering to the DOM is very "transactional", meaning that some intermediate states are not being flushed to the DOM. In our case the issue is with Animate component, and the fact that intermediate `entering` state is not rendered, but it goes straight to rendering `entered` state. Because of that animations are not working - there is not change to the styles as seen from the DOM/browser perspective.

I am not 100% sure what is the exact issue, it might be some combination of Animate, PortalWrapper and Modal. However, the solution I have found that is working is to delay a bit the rendering of the content of PortalWrapper, by adding additional state element, and setting it to `true` when `isOpen` is changed to `true` as well.
I am not setting it back to `false` when modal is closed, because there seems to be some error related to this showing up in application, that I cannot locate the source for, but we don't need to change this flag at this point.

I've put this behind a prop flag, so that it doesn't break any existing Modal usages outside of Beacon (and all Beacon Modals would be given this flag). That's mostly because that are some callback available on Modal that might rely on content to be rendered immediately.

### Testing
This can be of course tested by opening storybook and making sure Modals are working correctly :)

However, proper testing of the issue cannot be really done in this repo, only when using Beacon :( We'll do this in separate PR.

Write to your heart's content, include:

- [ ] A link to the Figma design in your story (list regularly updated [here](https://docs.google.com/spreadsheets/d/19-5gNbYuKjOb-kk7VTQZ0i_VWVd_8lubx-uhrlPUu1E/edit#gid=0))
- [ ] A link to the Story(ies) in the description
- [ ] Is there a Jira ticket associated?
- [ ] If useful, add screenshots or videos
- [ ] If useful (and unclear), add a little explanation of why a certain path was taken
- [ ] Instructions on how to test
- [ ] Found any restrictions/limitations? Let us know!

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/PortalWrapper/PortalWrapper.jsx
+++ b/src/components/PortalWrapper/PortalWrapper.jsx
@@ -54,6 +54,7 @@ const PortalWrapper = (options = defaultOptions) => ComposedComponent => {
     static defaultProps = {
       closeOnEscape: true,
       isOpen: false,
+      delayedContentRender: false,
       preventEscActionElements: extendedOptions.preventEscActionElements || [],
     }
     static childContextTypes = {
@@ -108,7 +109,7 @@ const PortalWrapper = (options = defaultOptions) => ComposedComponent => {
     }
 
     componentDidMount() {
-      const { path } = this.props
+      const { path, delayedContentRender } = this.props
       this._isMounted = true
       this.setTriggerNode()
 
@@ -116,7 +117,7 @@ const PortalWrapper = (options = defaultOptions) => ComposedComponent => {
         this.openPortal()
       }
 
-      if (this.state.isOpen) {
+      if (this.state.isOpen && delayedContentRender) {
         this.safeSetState({ renderContent: true })
       }
     }
@@ -142,7 +143,11 @@ const PortalWrapper = (options = defaultOptions) => ComposedComponent => {
         this.refocusTriggerNode()
       }
 
-      if (isOpen !== prevState.isOpen && isOpen) {
+      if (
+        isOpen !== prevState.isOpen &&
+        isOpen &&
+        this.props.delayedContentRender
+      ) {
         this.safeSetState({ renderContent: true })
       }
     }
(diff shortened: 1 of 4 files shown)


jakubjanczyk:
@plbabin @tinkertrain 
This is a bit hacky, but I have not found another way to fix the issue for Beacon. I know Modal is a bit in deprecation state (to be replaced with SimpleModal), but it's really hard to get rid of it from Beacon code and we don't have capacity for it now :smile: I suspect the issue we have would not be present in SimpleModal, but maybe one day we'll replace it (I'd love to because it much lighter).

Since this repo is open, I don't want to put links to Beacon preview to show you the issue and fixed version, but please let me know in case you'd want it, I'll share on slack :)

src/components/PortalWrapper/PortalWrapper.jsx
```diff
@@ -299,20 +314,22 @@ const PortalWrapper = (options = defaultOptions) => ComposedComponent => {
...
-                portalIsOpen={portalIsOpen}
-                portalIsMounted={portalIsOpen}
-                forceClosePortal={this.forceClosePortal}
-                trigger={trigger}
-                zIndex={zIndex}
-                {...rest}
-              />
-            </Content>
+            {(!this.props.delayedContentRender || this.state.renderContent) && (
+              <Content manager={this._MrManager} id={this._portalWrapperId}>
```
Charca, line 318: `MrManager` 😆

Charca (approved):
Looks good to me! I like that you added the `delayedContentRender` prop so that we don't potentially break any existing behavior!

src/components/PortalWrapper/PortalWrapper.jsx
jakubjanczyk: Every time I look into PortalWrapper, there's this piece of code:
```
// Welcome aboard, Mr. Manager!
_MrManager = setupManager(managerNamespace)
// Wow, I'm Mr. Manager!
// Well, manager… we we just say manager.
```
:stuck_out_tongue:

tinkertrain (approved):
:shipit:

PR #1051, opened 2022-04-19
ActionSelect: tooltip on SelectTag toggler

tinkertrain:
## Feature

Adds Tooltip support to DropList's `SelectTag` toggler, much the same as the icon button togglers:

```jsx
<SelectTag withTooltip tooltipProps={{ title: 'hello' }} />
```

This in turn has been added to the ActionSelect component:

```jsx
<ActionSelect withTooltip tooltipProps={{ title: 'hello' }} />
```

`withTooltip` and `tooltipProps` will just be forwarded to SelectTag (with some sensible defaults for ActionSelect, but can be overridden)

Screen Shot 2022-04-19 at 2 02 41 PM

https://510986f6.hsds-react.pages.dev/?path=/story/phaseout-actionselect--default-story

### Tooltip
1. Adjusted the custom width and padding so that Tippy positions the tooltip accurately (the padding directly on the Tooltip was messing the calculations Tippy does, so I moved it to the title container)

CleanShot 2022-04-19 at 14 15 24@2x

2. Positioning of the arrow was off if using the "start/end" variations of "top/bottom" (left/right are fine)

CleanShot 2022-04-19 at 13 53 37@2x

### DropList "Meat Button" toggler
1. There were some tippy props that were messing with the trigger area (making it the parent of the button instead of the button), I just removed those (can't even remember why they were there in the first place 🤷 )

CleanShot 2022-04-19 at 14 11 18

## Other
- Updates Tippy to `4.2.6`


justinwolfe (approved):
🥳 Looks awesome, @tinkertrain!! Thanks for this

plbabin (approved):
🎉 🎉 🎉

PR #1052, opened 2022-04-19
Switch: updates styles for accessibility

tinkertrain:
## Problem
The Switch focus state was basically invisible, we update the styles to match the most up-to-date design:

Screen Shot 2022-04-19 at 3 27 37 PM

## Solution
Use the `focusRing` we use on buttons, update colours:

shows

Updates the story to showcase better: https://735dad87.hsds-react.pages.dev/?path=/story/components-forms-switch--default-story


jakubjanczyk (approved):
Perfect! :100: :fire:

PR #1055, opened 2022-04-27
HSDS-259: Embedding video in Message and Attachment custom content

jakubjanczyk:
# Problem/Feature
### Embedded video in Message Card
Added possibility to embed Video in a Message, by setting provided iframe HTML code on the MessageCard.
There is a handler to not show a Message until the video iframe is fully loaded.

This can be tested by opening newly added stories: `MessageCard` -> `With video` and `With iframely video url`.

The second story is added for the purpose of showing difference between embedding video directly from service URL vs. using Iframely hosted ones. In the second case, waiting for video to load before showing a Message, and you can observe a moment where empty element is visible on the Message (which should not happen on the first story).
This is something that has be dealt with outside of HSDS, but I'm leaving those two here so we have easy comparison.

Screenshot from 2022-04-27 16-37-20

### New icon
Additionally, new icon `play-video-large` has been added to the set of icons - this would be used in Messages builder.

Screenshot from 2022-04-27 16-37-43

### Attachment custom content
Again, for the purpose of Message usage I modified `Attachment` component so that it can take custom content to be render within the container.
Here can be find example of what we need to render in the end: https://www.figma.com/file/WiSBxBAFZgk1iX4RFdF2hl/WIP%2F-Video-Support-for-Messages?node-id=28%3A12711

I've figured out it doesn't make much sense to provide some `prefix` or whatever properties, but just let user set the content they need in given case.

There is a story for this change: `Attachment` -> `Preview theme with custom content`

Screenshot from 2022-04-28 11-14-11

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/src/components/Attachment/Attachment.jsx
+++ b/src/components/Attachment/Attachment.jsx
@@ -78,6 +78,7 @@ export class Attachment extends React.PureComponent {
       truncateLimit,
       type,
       url,
+      content,
       ...rest
     } = this.props
     const { theme } = this.context
@@ -99,25 +100,31 @@ export class Attachment extends React.PureComponent {
       </Text>
     ) : null
 
-    const contentMarkup = imageUrl ? (
-      <span className="c-Attachment__content">
-        <Image
-          block
-          className="c-Attachment__image"
-          src={imageUrl}
-          alt={name}
-        />
-      </span>
-    ) : (
-      <span className="c-Attachment__content">
-        <Text className="c-Attachment__name" lineHeightReset>
-          <Truncate limit={truncateLimit} text={name} type="middle">
-            {name}
-          </Truncate>
-        </Text>
-        {sizeMarkup}
-      </span>
-    )
+    function contentMarkup() {
+      if (content) {
+        return <span className="c-Attachment__content">{content}</span>
+      }
+
+      return imageUrl ? (
+        <span className="c-Attachment__content">
+          <Image
+            block
+            className="c-Attachment__image"
+            src={imageUrl}
+            alt={name}
+          />
+        </span>
+      ) : (
+        <span className="c-Attachment__content">
+          <Text className="c-Attachment__name" lineHeightReset>
+            <Truncate limit={truncateLimit} text={name} type="middle">
+              {name}
+            </Truncate>
+          </Text>
+          {sizeMarkup}
+        </span>
+      )
+    }
 
     const closeMarkup = isThemePreview ? (
       <CloseButton
@@ -143,7 +150,7 @@ export class Attachment extends React.PureComponent {
         title={name}
         {...downloadProps}
       >
-        {contentMarkup}
+        {contentMarkup()}
         {closeMarkup}
         {this.renderErrorBorder()}
       </AttachmentUI>
--- a/src/components/Attachment/Attachment.jsx
+++ b/src/components/Attachment/Attachment.jsx
@@ -102,27 +102,28 @@ export class Attachment extends React.PureComponent {
 
     function contentMarkup() {
       if (content) {
-        return <span className="c-Attachment__content">{content}</span>
+        return content
       }
-
-      return imageUrl ? (
-        <span className="c-Attachment__content">
+      if (imageUrl) {
+        return (
           <Image
             block
             className="c-Attachment__image"
             src={imageUrl}
             alt={name}
           />
-        </span>
-      ) : (
-        <span className="c-Attachment__content">
+        )
+      }
+
+      return (
+        <>
           <Text className="c-Attachment__name" lineHeightReset>
             <Truncate limit={truncateLimit} text={name} type="middle">
               {name}
             </Truncate>
           </Text>
           {sizeMarkup}
-        </span>
+        </>
       )
     }
 
@@ -150,7 +151,7 @@ export class Attachment extends React.PureComponent {
         title={name}
         {...downloadProps}
       >
-        {contentMarkup()}
+        <span className="c-Attachment__content">{contentMarkup()}</span>
         {closeMarkup}
         {this.renderErrorBorder()}
       </AttachmentUI>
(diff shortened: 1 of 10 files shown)


jakubjanczyk:
@Charca @tinkertrain just remembered that there is one other change I need to make to `Attachment` component in HSDS for Messages, so I'm marking this one as WIP for now :)

src/components/MessageCard/MessageCard.test.js
```diff
@@ -344,6 +344,67 @@ describe('image', () => {
...
+
+    act(() => {
+      jest.runAllTimers()
+    })
+
+    expect(cardWrapper(container)).toBeVisible()
+    expect(onShowSpy).toHaveBeenCalled()
+  })
+
+  test('Should not be visible by default if there is a video, but become visible when image loads', () => {
```
Charca, line 378: I love how clear this test is 😍

Charca (commented):
I know you're not done with this PR yet, Jakub, but the code is looking good to me so far!

jakubjanczyk:
ok, @tinkertrain @Charca this is good for review :smile:

src/components/Attachment/Attachment.jsx
```diff
@@ -99,25 +100,31 @@ export class Attachment extends React.PureComponent {
...
-      <span className="c-Attachment__content">
-        <Text className="c-Attachment__name" lineHeightReset>
-          <Truncate limit={truncateLimit} text={name} type="middle">
-            {name}
-          </Truncate>
-        </Text>
-        {sizeMarkup}
-      </span>
-    )
+    function contentMarkup() {
```
tinkertrain, line 103: I think you have an opportunity to clean this up a bit, something like:

```jsx
function contentMarkup() {
  if (content) {
    return content
  } else if (imageUrl) {
    return (
      <Image block className="c-Attachment__image" src={imageUrl} alt={name} />
    )
  }
  return (
    <Text className="c-Attachment__name" lineHeightReset>
      <Truncate limit={truncateLimit} text={name} type="middle">
        {name}
      </Truncate>
    </Text>
  )
}

// down below
<span className="c-Attachment__content">{contentMarkup()}</span>
```

src/components/MessageCard/MessageCard.jsx
```diff
@@ -13,8 +13,8 @@ import { MessageCardContent } from './components/MessageCard.Content'
 import { MessageCardSurvey } from './components/MessageCard.Survey'
 import MessageCardUrlAttachmentImage from './components/MessageCard.UrlAttachmentImage'
 import MessageCardArticleCard from './components/MessageCard.ArticleCard'
-
-function noop() {}
+import { MessageCardVideo } from './components/MessageCard.Video'
+import { noop } from '../../utilities/other'
```
tinkertrain, line 17: `noop` usage in hsds-react coming from 'utils/other' has been removed. The only reason it still lives in the repo is because it's used in apps and we ned to update them to not use it.

src/components/MessageCard/MessageCard.jsx
```diff
@@ -53,11 +54,17 @@ export const MessageCard = React.memo(
...
+        }, 0)
+      }, [])
+
       useEffect(() => {
-        if (inProp && !hasImage() && !isShown.current) {
+        if (inProp && !hasImage() && !isShown.current && !video) {
           makeMessageVisible()
         }
-      }, [inProp, hasImage])
+      }, [inProp, hasImage, video, makeMessageVisible])
```
tinkertrain, line 67: Since video is on object (and I assume there is a reason for that) and `useEffect` compares values referencially and not deeply, it might trigger rerendering when you don't want to.

You can either use the `use-deep-compare-effect` package (already installed, works the same as useEffect) or if you know that `video.html` is always a string, extract that and pass it to useEffect:

```js
const { html } = video

useEffect(() => {
  if (inProp && !hasImage() && !isShown.current && !video) {
    makeMessageVisible()
  }
}, [inProp, hasImage, html, makeMessageVisible])
```

src/components/MessageCard/MessageCard.jsx
```diff
@@ -97,14 +98,15 @@ export const MessageCard = React.memo(
...
-              render={!!(body || image || children)}
+              render={!!(body || image || children || video)}
             >
               <MessageCardBody
                 body={body}
                 onClick={onBodyClick}
                 variables={variables}
               />
               <MessageCardImage image={image} onLoad={makeMessageVisible} />
+              <MessageCardVideo video={video} onLoad={makeMessageVisible} />
```
tinkertrain, line 109: Suggestion, I think is clearer that MessageCardVideo will only render if `video` is present if you put it here:
```jsx
{ video ? <MessageCardVideo video={video} onLoad={makeMessageVisible} /> : null }
```

remove this check in `MessageCardVideo`: 
```js
if (!video) {
  return null
}
```
And make the `video` prop in there required.

src/components/MessageCard/components/MessageCard.Video.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+    return () => {
+      if (videoIframeRef.current) {
+        videoIframeRef.current.onload = undefined
+        videoIframeRef.current.onerror = undefined
+      }
+    }
+  }, [onLoad, video])
+
+  if (!video) {
```
tinkertrain, line 29: As stated in MessageCard, I would move this off here

tinkertrain (changes requested):

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: ah, thanks, good to know!

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: Great catch! I'll use comparing the `html` :)

src/components/MessageCard/MessageCard.jsx
jakubjanczyk: Will do! (if there isn't any reason why I did it like this, but I'll take a look)

jakubjanczyk:
@tinkertrain ready for re-review!

tinkertrain (approved):
Nice work! 💯

tinkertrain (approved):
🚢

PR #1059, opened 2022-05-04
useClickOutside hook fix

tinkertrain:
# Problem

There are cases when an element inside a modal is no longer there byt the time `useClickOutside` decides whether to execute the callback, making the check always true and so, executing the callback when it shouldn't.

## Solution

Add an extra check: make sure the element clicked is visible to then execute the callback.

Added a test case in the story: https://1713afae.hsds-react.pages.dev/?path=/story/components-overlay-simplemodal--with-trigger-and-nested

1. Open the modal
2. Click the green button, it will get removed and the modal stays open (whereas previously it was closing it)
3. Clicking outside the modal, closes the modal

### DIFF
--- a/src/components/SidePanel/SidePanel.test.js
+++ b/src/components/SidePanel/SidePanel.test.js
@@ -96,6 +96,40 @@ describe('Overlay', () => {
 })
 
 describe('Closing', () => {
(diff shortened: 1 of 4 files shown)


knicklabs (approved):
Thanks for this!

PR #1061, opened 2022-05-10
HSD- 261 Update CheckmarkCard disabled UI

plbabin:
# Problem/Feature
CleanShot 2022-05-10 at 10 13 54
 
To support the work in an upcoming project, we had to implement some changes to the disabled state of the CheckmarkCard:

## Update the UI
We removed all shadows and set a grey background (`grey.200`). Avatar is faded out and the heading/subtitle color are slightly lighter
CleanShot 2022-05-10 at 10 35 26@2x

## Add tooltip support to the whole card
By passing a cardTooltipText prop to the CheckmarkCard, a tooltip will appear on focus/mouseover. The tooltip will still work when the card is disabled
CleanShot 2022-05-10 at 10 37 05@2x

## Focus ring
we migrate the component to use the `focusRing` mixin for both the card and mark component. It simplifies the code but nothing should have change visually

### DIFF
--- a/src/components/CheckMarkCard/CheckMarkCard.css.js
+++ b/src/components/CheckMarkCard/CheckMarkCard.css.js
@@ -4,9 +4,11 @@ import { d400, d400Effect } from '../../styles/mixins/depth.css'
 import ChoiceGroup from '../ChoiceGroup'
 
 import { getColor } from '../../styles/utilities/color'
-import { rgba } from '../../utilities/color'
+import { focusRing } from '../../styles/mixins/focusRing.css'
 
 export const MarkUI = styled('div')`
+  ${focusRing};
+  --focusRingRadius: 3px;
   position: absolute;
   top: 0;
   left: 0;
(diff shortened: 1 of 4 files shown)


tinkertrain (approved):
Looks great! One thing, perhaps for @phamburglar, normally a disabled element (a button, input, etc) is not focusable as it's not actionable. I think it makes sense to apply this pattern to this element.

I realize this isn't the case currently (and RadioCard probably "suffers" the same issue, I haven't checked). Just thought of it and if we want it, it might make sense to include the change here (the change shouldn't be big I don't think).

plbabin:
@phamburglar @tinkertrain It's possible to hide the focus state on when the CheckmarkCard is disabled. I did include it here because I wanted to show the tooltip by both mouse and keyboard.

The question is, what is more inline with a11y here:  not focusable when disabled or not being able to show the tooltip when keyboard navigating to it 🤷. There is no perfect answer here I guess. The call we'll made will be the right one.

Long story short, anything's possible here 😄

phamburglar:
@plbabin @tinkertrain Went down a mini rabbit hole and came across [this link](https://css-tricks.com/making-disabled-buttons-more-inclusive/#aa-aria-to-the-rescue) that explains how to make our scenario more accessible.

tldr; We should allow it to be focusable, even when disabled so that the user can see the tool tip. Essentially the "greyed" out opacity state will visually let them know it's disabled, but the focusable state will allow them to display the tooltip in line with what they're focused on. 👍

plbabin:
@phamburglar so are we good with the changes in this PR? I know you approved it but just want to make sure that everything's ok following the a11y discussion

PR #1062, opened 2022-05-10
Capture all keyboad originating in Simple Modal

TerrenceLJones:
## Problem

[Jira ticket](https://helpscout.atlassian.net/browse/HSDS-263)

Keyboard events originating when the Simple Modal was displayed were leaking out onto the page, causing page keyboard shortcuts to be triggered. 

## Solutions

1. Stop propagation of key and click events from the modal
2. Add a `blocksGlobalHotkeys` prop to SimpleModal (and SidePanel) that will add a data attribute that external hooks or events can look for to perform or not a certain action when a modal is present

### Other
1. When having multiple modals on screen outside clicks were being firing 'onClose' for every modal, we added scenarios were we don't want this to happen, for example if having 2 modals open, and we click outside the modal on top, we only want to fire the onClose for that one, not the underneath one.

#### Unrelated
1. Tooltip: removed some styles that were added to fix arrow positioning but now we've seen they cause issues when the tooltip is appended to anything but the default.
2. DropList: little tweak to make the input on comboboxes always show the blue outline to avoid flashing when selecting items

### DIFF
--- a/src/components/SimpleModal/SimpleModal.jsx
+++ b/src/components/SimpleModal/SimpleModal.jsx
@@ -54,8 +54,10 @@ function SimpleModal({
   }
 
   function handleOverlayKeyDown(e) {
+    // Prevent keyboard events originating from the modal from leaking out onto the page.
+    e.stopPropagation()
+
     if (shouldRender && e.key === 'Escape') {
-      e.stopPropagation()
       onClose(e)
     } else if (e.key === 'Tab' && trapFocus && modalRef.current) {
       manageTrappedFocus(modalRef.current, e)
--- a/src/components/SimpleModal/SimpleModal.jsx
+++ b/src/components/SimpleModal/SimpleModal.jsx
@@ -14,6 +14,7 @@ function noop() {}
 
 function SimpleModal({
   ariaLabelledBy = '',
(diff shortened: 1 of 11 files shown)


src/components/SimpleModal/SimpleModal.jsx
```diff
@@ -54,8 +55,10 @@ function SimpleModal({
   }
 
   function handleOverlayKeyDown(e) {
+    // Prevent keyboard events originating from the modal from leaking out onto the page.
```
TerrenceLJones, line 58: @tinkertrain With your solution...would this change still be needed or can we revert it?

TerrenceLJones (commented):
Love this solution! Excellent work!

src/components/SimpleModal/SimpleModal.jsx
tinkertrain: Hmm, good question. There's a clear case for halting propagation with `esc`. If we think about the modal as a "scoped context" in a page, perhaps your change should stay, I can't came up with a scenario where this would be trouble so my hunch is to keep it.

Let's bring the big guns... @plbabin any thougths?

TerrenceLJones (commented):
LGTM!

PR #1065, opened 2022-05-18
SimpleModal: Adds confirmation layout

tinkertrain:
# Feature

Adds confirmation layout to simple-modal

Follows: https://www.figma.com/file/Wp9mDxTvWicSTWtezj2ImkMP/HSDS-Product?node-id=2582%3A10019

Usage:

```jsx
<SimpleModal show>
    <Confirmation
      heading="Alert!"
      body={
        <div>
			You are about to merge the selected conversation with this one — this action cannot be <em>undone</em>.
		</div>
		}
      confirmButtonText="Okay"
      cancelButtonText="Cancel"
      danger
      onCancel={noop}
      onConfirm={noop}
    />
</SimpleModal>
```

Screen Shot 2022-05-19 at 13 09 54-fullpage

Story: https://55497051.hsds-react.pages.dev/?path=/story/components-overlay-simplemodal--confirmation-layout

## unsetDimensions

I also added a new prop `<SimpleModal unsetDimensions />`, it will clear the width and height that is normally set on the modal. This is useful if you need to let the content take care of the size without having to override with styled components or something like that.


kyliedunkley:
Hiii Juan Pablo, 

Can we please make the top padding 50px and change the title to "Alert Title"? (we are trying to go away from the "Are you sure?" language. Below the title should be 6px, unless no sub copy and then it is 25px.

This is amazing work! Thank you!

src/components/SimpleModal/SimpleModal.layouts.jsx
```diff
@@ -47,3 +58,45 @@ HeaderAndFooter.propTypes = {
...
+      </ConfirmationFooterUI>
+    </ConfirmationWrapperUI>
+  )
+}
+
+Confirmation.propTypes = {
+  /** Text of the modal heading */
+  heading: PropTypes.string.isRequired,
+  /** Text of the modal body */
+  body: PropTypes.string,
```
knicklabs, line 91: Can we make the body accept a React Node instead of string? I think we may need to apply text formatting, i.e. to create emphasis.

knicklabs (changes requested):
This is looking good to me @tinkertrain! The API should be compatible with our use. Looks like a really clean solution. I may try to do a beta release tomorrow AM so I can try it out in app and double-check compatibility, but in the meantime, I'm just wondering if we can change `body` to accept a React node so we can pass in a component instead of text in cases where we want to apply formatting, for example, to create emphasis on just part of the body.

src/components/SimpleModal/SimpleModal.layouts.jsx
tinkertrain: Of course! No problem, I added that to the heading as well 👍

tinkertrain:
@kyliedunkley Thanks for looking at it!

About the padding... the reason I put 45px is because when I measured with 50px I got:

Screen Shot 2022-05-19 at 11 58 08 AM

And with 45px:

Screen Shot 2022-05-19 at 11 55 50 AM

It has to do with the line height of the heading, I can set it to a smaller value to match the actual height of the text but there is a case where that will look bad: if the heading for some reason spans 2 lines. Granted, that is an edge case for this type of modal.

But is up to you, I can change it, no problem!

tinkertrain:
@knicklabs @kyliedunkley Quick question, I just had a thought, do we ever use this type of modal with just one button? Sort of as "notice" of sorts? Wanted to know if it's worth embedding that case here or not (at the moment it always renders 2 buttons)

tinkertrain:
@kyliedunkley Thanks for looking at it!

kyliedunkley:
> @knicklabs @kyliedunkley Quick question, I just had a thought, do we ever use this type of modal with just one button? Sort of as "notice" of sorts? Wanted to know if it's worth embedding that case here or not (at the moment it always renders 2 buttons)

I have only seen use case of one button. :)

knicklabs:
> @knicklabs @kyliedunkley Quick question, I just had a thought, do we ever use this type of modal with just one button? Sort of as "notice" of sorts? Wanted to know if it's worth embedding that case here or not (at the moment it always renders 2 buttons)

@tinkertrain so far I've only come across the case with both a primary button and cancel button.

knicklabs (approved):

PR #1066, opened 2022-05-24
Fix survey issues

jakubjanczyk:
# Problem/Feature

This PR fixes some issues found for Messages survey components:
- Color of submit button was always blue, it should be using the current theme instead
- Text for submit button was not truncated
- The animations/transitions when selecting thumb/face was a bit clunky - I had to delay (optionally) showing confirmation form after selecting an option, so that the animation for option finishes before. I also added some height animation to the form itself for nicer entry

## Guidelines

Make sure the pull request:

- [ ] Adds unit tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@helpscout/hsds-react",
-  "version": "3.54.0",
+  "version": "3.54.1-0",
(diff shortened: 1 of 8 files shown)


tinkertrain (approved):
Looks good to me!

PR #1069, opened 2022-05-25
SimpleModal (confirmation layout): Swap which buttons handlers are applied to

knicklabs:
# Problem/Feature

The button handlers were connected in reverse:

- The `onCancel` handler was connected to the "Confirm" button.
- The `onConfirm` handler was connected to the "Cancel" button.

# Solution

This pull request is a small change to flip the usage of the button handlers:

- The `onCancel` handler is connected to the "Cancel" button.
- The `onConfirm` handler is connected to the "Confirm" button.

That's it!

### One more thing

Adds 2 props `cancelButtonProps` and `confirmButtonProps` to pass any valid Button prop to each button.

### DIFF
--- a/src/components/SimpleModal/SimpleModal.layouts.jsx
+++ b/src/components/SimpleModal/SimpleModal.layouts.jsx
@@ -73,10 +73,10 @@ export function Confirmation({
       <ConfirmationHeadingUI>{heading}</ConfirmationHeadingUI>
       {body && <ConfirmationBodyUI>{body}</ConfirmationBodyUI>}
       <ConfirmationFooterUI>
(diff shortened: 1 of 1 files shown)


tinkertrain (approved):
🤦 Sorry!

I took the liberty to add 2 props `cancelButtonProps` and `confirmButtonProps` to pass any valid Button prop to each button, the reason for this is that I reviewed a PR by Hope in which the only thing missing for he to use this was the ability to pass a `loading` prop to the confirm button... so we might as well add it here.

cc @hope-uwa

knicklabs:
@tinkertrain the new props look good!

PR #1073, opened 2022-06-08
HSDS-266 Attachment & AttachmentList update

plbabin:
# Feature
To support some work happening on the mailbox, we had to update the UI of the attachment & attachment list plugin. This component has not been touch for over 3 years, we took the time to rewrite it as a functional component. All that while staying retro compatible with the API and with the way Provider was use in some places.

CleanShot 2022-06-07 at 16 22 42

CleanShot 2022-06-07 at 16 19 59@2x

## html cleanup
We remove the usage of `Inline` for the grid layout of the AttachmentList. We now rely on flexbox directly on Attachment children.

## Broken image
The old UI was just displaying a broken image from the browser. Now we detect that there was an error while loading the image and flip the UI to a grey background component. That ensure that the file name will not overflow out of the container. We also wrap the component with a tooltip.

## isRemovable
It's not possible to hide the remove button with this props. By default it will be visible.

## Download All
We add the possibility of displaying a download all button in preview mode.

## Focus state
We add a focus state for the attachment when there is an url attach to it. User will be able to <kbd>tab</kbd> through both attachment and the remove button.

## React testing library
Re rewrote all test with RTL


kyliedunkley:
when I tab through it goes to the x before the image/block in all of them except the last. I am pretty sure that may be a constraint of storybook but wanted to check :)

plbabin:
> when I tab through it goes to the x before the image/block in all of them except the last. I am pretty sure that may be a constraint of storybook but wanted to check :)

this is when there is no url on the attachment (so they are not "clickable"). This is a scenario that will 99.5% never exists in prod

kyliedunkley:
ooo gotcha then I will go and approve!

kyliedunkley (approved):

src/components/AttachmentList/AttachmentList.jsx
```diff
@@ -1,148 +1,124 @@
...
-      theme && `is-theme-${theme}`,
-      withOverflow && 'withOverflow',
-      className
-    )
-
-    const attachmentChildren = React.Children.toArray(children).filter(
-      child => child.type && child.type === Attachment
+    return (
+      <Attachment
+        elementClassName="AttachmentList__DownloadAll"
```
luketlancaster, line 74: should this be `c-AttachmentList__DownloadAll`? What're y'all's rules around prepending the classnames with `c-`?

src/components/AttachmentList/AttachmentList.test.js
```diff
@@ -1,210 +1,164 @@
 import React from 'react'
-import { mount, shallow } from 'enzyme'
+import { render, screen } from '@testing-library/react'
+import { mount } from 'enzyme'
 import { AttachmentList } from './AttachmentList'
 import { Attachment, Icon } from '../index'
```
luketlancaster, line 5: looks like you don't need to import `screen`, `mount` or `Icon` here

src/components/AttachmentList/AttachmentList.test.js
```diff
@@ -1,210 +1,164 @@
...
+    const { getByTestId } = render(<AttachmentList className={customClass} />)
+    expect(getByTestId('AttachmentList')).toHaveClass(customClass)
   })
 })
 
 describe('Custom attributes', () => {
   test('Can render custom HTML attributes', () => {
-    const wrapper = shallow(<AttachmentList data-tie="piano-key" />)
-    const html = wrapper.html()
+    const { getByTestId, container } = render(
```
luketlancaster, line 29: `container` is unused

luketlancaster:
@plbabin pulled down the beta and was testing it out in chat. Only thing that looks off is that we're defaulting the Download All button to show, so if we ever end up updating hsds in hs-app that'll start showing up (but since it's not wired up nothing happens when you click). I think we should default to **not** showing that button, and having to "enable" it in new uses of this component: thoughts?

CleanShot 2022-06-15 at 09 52 15

plbabin:
> oh hm...actually thinking about this, I think we show the downloadall by default in the other uses of this component so I don't think we should switch that behavior 🤔
> 
> I guess we'll just add in the prop to hs-app if we ever get around to updating hsds over there? Hmmm

yeah that make sense, 7bce1786 contain the update. I keep the download all visible by default on the list theme (preview is turned off by default)

src/components/AttachmentList/AttachmentList.jsx
```diff
@@ -1,148 +1,124 @@
...
+    showDownloadAll,
+    withOverflow,
+    onScrollEnd,
+    theme: themeProp,
+    ...rest
+  } = props
+
+  const attachmentChildren = React.Children.toArray(children).filter(
+    child => child.type && child.type === Attachment
+  )
```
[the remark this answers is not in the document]
plbabin: not really needed, just in case someone do insert something not supported within a list
```javascript
<AttachmentList>
  <Attachment />
  <Button />
  <Attachment />
</AttachmentList>
```

src/components/AttachmentList/AttachmentList.jsx
plbabin: it should

src/components/Attachment/Attachment.jsx
```diff
@@ -1,170 +1,193 @@
...
 
-  renderErrorBorder() {
-    return (
-      this.isError() && (
-        <ErrorBorderUI
-          className="c-Attachment__errorBorder"
-          isCard={this.isThemePreview()}
+  const downloadProps = {
+    download: download !== undefined ? download : url ? true : null,
+    target: target !== undefined ? target : url ? '_blank' : '',
```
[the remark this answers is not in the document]
plbabin: yep, I cleanup the code to be easier to understand (while it's a bit more verbose but that's fine)

```javascript
  const downloadProps = { download, target }
  if (downloadProps.download === undefined) {
    downloadProps.download = Boolean(url)
  }
  if (downloadProps.target === undefined) {
    downloadProps.target = url ? '_blank' : ''
  }
```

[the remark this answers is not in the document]
luketlancaster: oh nice, thanks for the explainer!

luketlancaster:
This looks good to me now, gonna run another beta release to verify things still are looking good. If so we can get this thing shipped!

PR #1076, opened 2022-06-30
Table: adds the ability to change the table role

tinkertrain:
# Feature
We add a new table prop `tableRole` which changes the `role` attribute of the table element for special cases that require it.

When we change this attribute to something other than "table", we also make other attribute changes (in sortable tables)

- `th` no longer get the `aria-sort` attribute set
- Instead we add a `aria-label` attribute on the button inside the `th`, that reads "Sorted {columnName} name {order}" for example: "Sorted by name ascending"

We also adjust the row border focus from 2px to 3px
And we update all table tests to use testing-library

### DIFF
--- a/src/components/Table/Table.Head.jsx
+++ b/src/components/Table/Table.Head.jsx
@@ -17,6 +17,7 @@ function TableHead({
   selected,
   selectionKey,
   sortedInfo,
+  tableRole,
   withSelectableRows,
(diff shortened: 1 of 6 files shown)


knicklabs (approved):
Looks good @tinkertrain! I'll go ahead and ship this as we discussed in Slack.

